PR #4, opened 2019-05-04
feat: Initial Module Extraction Implementation

YashishDua:
Two broad cases:
1. If the go.mod file is present, packages are extracted from the file.
2. Else If a vendor folder is present:
    A. Check for modules.txt file. If present, extract packages from this file.
    B. If not, check for vendor.json to extract packages.

Currently, the integration is loosely placed in a separate file. @kamilogorek would be shifting it to the right place.


integrations.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"bufio"
+	"io/ioutil"
+  "os"
+  "fmt"
+	"encoding/json"
+	"strings"
+)
+
+func GetModules() (map[string]string, error) {
```
kamilogorek: Can we split this into 3 private functions, one for each possible scenario? It'd make `GetModules` much easier to read.

```go
func GetModules() (map[string]string, error) {
  if something {
    return _getModulesOneWay()
  }

  if somethingElse {
    return _getModulesOtherWay()
  }

  return _defaultWay()
}
```

integrations.go
```diff
@@ -0,0 +1,105 @@
...
+          lastPath = path
+        }
+      }
+
+      return modules, nil
+    }
+	}
+
+  return nil, fmt.Errorf("Module integration failed")
+}
```
kamilogorek: Can you run the code through `gofmt`?

util.go
```diff
@@ -15,3 +16,10 @@ func uuid() string {
 	id[8] |= 0x80 // set to IETF variant
 	return hex.EncodeToString(id)
 }
+
+func checkFileExist(fileName string) bool {
```
kamilogorek: Can we name it `fileExists` instead? It reads better in conditionals

```go
if fileExists('some/path.go') {
  // do thing
}
```

integrations.go
```diff
@@ -0,0 +1,105 @@
+package sentry
```
kamilogorek: Modules will never change throughout the lifecycle of the app and we want to minimize number of I/O operations. Can we cache the result once it was called?

integrations.go
YashishDua: For sure.

integrations.go
YashishDua: Alright.

util.go
YashishDua: Makes sense.

integrations.go
YashishDua: How can it not change? Every time any dependency is added to the project. Modules will change of course and this can happen at any frequency.

kamilogorek: And every time you restart your app, it'll cache a new state :)

YashishDua: Oh, in this way. Yes, makes sense! I will do the requested changes by tomorrow.

kamilogorek:
Thanks! 🎉

YashishDua:
Yaya! Super happy. 🎉

PR #5, opened 2019-05-08
feat: Transport implementation

kamilogorek:
What default transport should be able to do:
- DONE: queue events delivery
- DONE: drop events after queue is filled (30 requests)
- DONE: allow for knowing when everything was sent (`flush`)
- DONE: blocks the queue when server returns `429` using `Retry-After` header or 60sec default
- DONE: allow for configuring proxy and custom ssl certificated
- DONE: log to a debugger when server responds with error

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@ go:
   - 1.11.x
   - 1.12.x
 
+env:
+  - GO111MODULE=on
+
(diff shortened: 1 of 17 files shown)


kamilogorek:
I'll merge it so we can iterate on this and discuss it further

PR #20, opened 2019-06-25
fix: Check for initialized Client in AddBreadcrumbs

tonyo:
Otherwise, we get a segfault at https://github.com/getsentry/sentry-go/blob/master/hub.go#L184 when Client is not initialized (e.g., empty DSN)


kamilogorek:
We can most likely just store it as is, without checcking for `BeforeBreadcrumb` and `MaxBreadcrumbs` as they live on the scope anyway. Will think about it next week.

PR #22, opened 2019-07-03
Allow empty DSN instead of returning an error

chuyeow:
Based on https://github.com/getsentry/sentry-go/blob/e3426d1877ca5368ddc0bfc112368a0c293fad0e/client.go#L130, it looks like the intent is to allow an empty DSN (e.g. when the `SENTRY_DSN` env var is not set).

The current behavior when the DSN is empty is to return an error from `NewClient()`. This PR allows the DSN to be empy.

### DIFF
--- a/client.go
+++ b/client.go
@@ -120,13 +120,14 @@ func NewClient(options ClientOptions) (*Client, error) {
 		options.Environment = os.Getenv("SENTRY_ENVIRONMENT")
 	}
 
-	dsn, err := NewDsn(options.Dsn)
-
-	if err != nil {
-		return nil, err
-	}
-
-	if dsn == nil {
+	var dsn *Dsn
+	if options.Dsn != "" {
+		var err error
+		dsn, err = NewDsn(options.Dsn)
+		if err != nil {
+			return nil, err
+		}
+	} else {
 		Logger.Println("Sentry client initialized with an empty DSN")
 	}
 
--- a/client_test.go
+++ b/client_test.go
@@ -5,6 +5,19 @@ import (
 	"testing"
 )
 
+func TestNewClientAllowsEmptyDSN(t *testing.T) {
+	transport := &TransportMock{}
+	client, err := NewClient(ClientOptions{
+		Transport: transport,
+	})
+	if err != nil {
+		t.Fatalf("expected no error when creating client without a DNS but got %v", err)
+	}
+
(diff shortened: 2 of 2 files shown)


kamilogorek:
Thanks for filling a PR!

To make it fully functional, we need to also create `NoopTransport` which implements `Transport` interface and just logs information about the error being dropped. With the current code, transport will still be installed and it'll try to create Dsn which will fail - https://github.com/getsentry/sentry-go/blob/e3426d1877ca5368ddc0bfc112368a0c293fad0e/transport.go#L100-L104

We need to use `NoopTransport` as the default transport for empty Dsn scenario instead.

chuyeow:
>  With the current code, transport will still be installed and it'll try to create Dsn which will fail

Whoops, I hadn't expected that. I'd assumed that because the global `Hub` has no `Client` bound (https://github.com/getsentry/sentry-go/blob/e3426d1877ca5368ddc0bfc112368a0c293fad0e/hub.go#L24), it will "do the right thing" and do nothing. Can you confirm that my thinking is wrong? I just looked at https://github.com/getsentry/sentry-go/blob/e3426d1877ca5368ddc0bfc112368a0c293fad0e/sentry.go#L39-L40 and https://github.com/getsentry/sentry-go/blob/e3426d1877ca5368ddc0bfc112368a0c293fad0e/hub.go#L172-L174 and it looks to me that since `hub.Client()` is `nil`, nothing is done.

kamilogorek:
It binds a client during the first `init` call, and your change makes empty DSN not return an error, thus it will be bound correctly.

https://github.com/getsentry/sentry-go/blob/master/sentry.go#L12-L20

Current output when calling `init` will be:
```
[Sentry] 2019/07/09 15:20:01 Sentry client initialized with an empty DSN
[Sentry] 2019/07/09 15:20:01 [Sentry] DsnParseError: invalid scheme
```

We need to disable transport (or use noop) and log information about it.

kamilogorek:
Closed in favor of https://github.com/getsentry/sentry-go/pull/27
@chuyeow your commit is preserved :)

chuyeow:
@kamilogorek woah thanks for doing that! I couldn’t find the time to work on this before, sorry for the late response.

kamilogorek:
No worries! Thanks :)

Issue #21, opened 2019-06-27
Upgrading to latest causes Lambdas to fail


mattcollum:
We've been using 0.0.1-beta.2 for a while successfully.  There was a message in the Sentry console today that we should update to 0.1.0 so we did.  Local testing seemed fine but when running in an actual environment (Lambda based) for some reason it short circuited all processing of the function.  Rolling back to the older version resolved it.  Any idea what may be causing this?  Are there code changes we should be making to get on to the current version?

kamilogorek:
@mattcollum all breaking changes are marked here in the changelog - https://github.com/getsentry/sentry-go/blob/master/CHANGELOG.md

Do you have any more insights for this issue? I'll take a look at it first thing on Monday.

kamilogorek:
I just tested everything on AWS and it seems to work just fine. When working with serverless solutions, you may want to take a look at https://docs.sentry.io/platforms/go/transports and use `HTTPSyncTransport` now, as it doesn't require `Flush` calls.

mattcollum:
@kamilogorek sorry for the delayed response.  holiday long weekend here :)

so the Lambdas just seem to exit after we init sentry and call `defer sentry.Recover()`.  

The init is pretty straightforward.  Calls to sentry include the following the first time:

sentry.Init with Dsn, AttachStacktrace and BeforeSend

And on subsequent calls if the Lambda was already inititalized we're calling:

`sentry.ConfigureScope(func(scope *sentry.Scope) {
			scope.Clear()
		})`

Haven't had a chance to review the change log or look into HTTPSyncTransport as we're pretty busy right now but can look down the road for sure.  For now I've locked the version back to beta.2 and it's working again.  If you happen to see anything in this init or Recover calls that could be the cause let me know

kamilogorek:
@mattcollum if you don't await event delivery, your lambda will instantly exit the process. You have to either flush the queue manually or use SyncTransport. Although I wonder how it's possible that it worked in `beta.2` without it?

Just deployed code below to Lambda and can confirm that it correctly reports to Sentry:

```go
package main

import (
	"encoding/json"
	"errors"
	"fmt"
	"github.com/aws/aws-lambda-go/lambda"
	"github.com/getsentry/sentry-go"
)

func foo() error {
	return bar()
}

func bar() error {
	return baz()
}

func baz() error {
	panic(errors.New("boom"))
}

func HandleRequest() (string, error) {
	defer sentry.Recover()
	foo()
[...]

PR #24, opened 2019-07-08
fix: Use NewScope instead of literal struct inside a scope.Clear call

kamilogorek:
Fixes https://github.com/getsentry/sentry-go/issues/21
cc: @mattcollum


mattcollum:
thx @kamilogorek 

lmk when the next release is coming out and I'll give it a shot

kamilogorek:
@mattcollum released as `0.1.2`

PR #26, opened 2019-07-10
feat: Better source code location resolution and more useful inapp frames

kamilogorek:
This allows us to read files content for multi-package apps deployed to AWS Lambda with
```
$ GOOS=linux go build main.go && zip -r handler.zip *
```

### DIFF
--- a/stacktrace.go
+++ b/stacktrace.go
@@ -1,6 +1,7 @@
 package sentry
 
 import (
+	"go/build"
 	"path/filepath"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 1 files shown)


stacktrace.go
```diff
@@ -237,8 +239,8 @@ func contextifyFrames(frames []Frame) []Frame {
 		switch {
 		case fileExists(frame.AbsPath):
 			path = frame.AbsPath
-		case fileExists(frame.Filename):
-			path = frame.Filename
+		case fileExists(strings.TrimPrefix(frame.AbsPath, sourceCodeLocationPrefix)):
```
bruno-garcia: What about assigning the result to be reused in the `case` block instead of running it twice?

PR #27, opened 2019-07-10
 feat: Use noopTransport when no DSN provided

kamilogorek:


client.go
```diff
@@ -120,14 +120,13 @@ func NewClient(options ClientOptions) (*Client, error) {
...
-	dsn, err := NewDsn(options.Dsn)
-
-	if err != nil {
-		return nil, err
-	}
-
-	if dsn == nil {
-		Logger.Println("Sentry client initialized with an empty DSN")
+	var dsn *Dsn
+	if options.Dsn != "" {
```
bruno-garcia: Do we look for `SENTRY_DSN` env var before hitting this?

PR #28, opened 2019-07-11
feat: Move frames context reading into contextifyFramesIntegration

kamilogorek:


integrations.go
```diff
@@ -260,3 +260,104 @@ func getIgnoreErrorsSuspects(event *Event) []string {
...
+		components = components[1:]
+		possibleLocation := strings.Join(components, "/")
+
+		if fileExists(possibleLocation) {
+			cfi.cachedLocations[originalPath] = possibleLocation
+			return possibleLocation
+		}
+	}
+
+	return ""
```
bruno-garcia: Will lookup forever in this case. Needs caching result

integrations.go
```diff
@@ -260,3 +260,104 @@ func getIgnoreErrorsSuspects(event *Event) []string {
...
+
+	for _, frame := range frames {
+		if !frame.InApp {
+			contextifiedFrames = append(contextifiedFrames, frame)
+			continue
+		}
+
+		var path string
+		// Optimize for happy path here
+		if fileExists(frame.AbsPath) {
```
bruno-garcia: Lookup the cache first to avoid I/O

integrations.go
```diff
@@ -260,3 +260,104 @@ func getIgnoreErrorsSuspects(event *Event) []string {
...
+		} else {
+			path = cfi.findNearbySourceCodeLocation(frame.AbsPath)
+		}
+
+		if path == "" {
+			contextifiedFrames = append(contextifiedFrames, frame)
+			continue
+		}
+
+		lines, initial := cfi.sr.readContextLines(path, frame.Lineno, cfi.contextLines)
```
bruno-garcia: Blocking I/O?

integrations.go
```diff
@@ -260,3 +260,104 @@ func getIgnoreErrorsSuspects(event *Event) []string {
...
+		if fileExists(possibleLocation) {
+			cfi.cachedLocations[originalPath] = possibleLocation
+			return possibleLocation
+		}
+	}
+
+	return ""
+}
+
+func (cfi *contextifyFramesIntegration) addContextLinesToFrame(frame Frame, lines [][]byte, initial int) Frame {
```
bruno-garcia: `initial` should be called `context_line` or something more meaningful?

integrations.go
kamilogorek: Yes, but it's all cached in the source reader once done

integrations.go
kamilogorek: Fixed

integrations.go
bruno-garcia: When it comes to blocking I/O, there's no `yes, but` 👿 
I guess ASP.NET Core is a bad influence?! or is it?

PR #29, opened 2019-07-19
Add SetTransaction() to Scope

jayd3e:
We noticed that according to the [unified spec](https://docs.sentry.io/development/sdk-dev/unified-api/#scope), there should be an `AddTransaction()` method that exists on the `Scope` object, which is used to create an event.  This currently does not exist in `sentry-go`.  This was a feature that we needed on our own project, so we went ahead and implemented it in `sentry-go`.

Bitcoin donations welcome: 38tt2b58t7TE9FS1BgjuZGYNUhmhGpwZ2D 😆

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -28,6 +28,7 @@ type Scope struct {
 	extra           map[string]interface{}
 	fingerprint     []string
 	level           Level
+	transaction     string
 	request         Request
 	eventProcessors []EventProcessor
 }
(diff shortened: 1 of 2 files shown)


kamilogorek:
Looks good! Could you also add tests to `scope_test.go` please? Thanks! :)

jayd3e:
Can you run the tests to ensure there are no failures?  I was not able to get the tests passing locally.  I am using Go 1.9 and receive a nil pointer exception.  The same result happens if I run tests on `master`, so it must be something specific to my local environment.

kamilogorek:
Added some more tests and updated `Clear` method to take `Transaction` into account as well.
Your commits were rebased and author was preserved -https://github.com/getsentry/sentry-go/commits/master

Thanks! :)

(and yes, tests are all passing on 1.10-1.12)

Issue #30, opened 2019-07-22
sentry.Recover() does not wait until error is sent


DimitriosLisenko:
In the docs, it is recommended that you call
```
defer sentry.Recover()
```
before running code that may panic.

However, it doesn't look like the framework actually waits for the error to be sent upstream. Modifying the `sentry.Recover()` code to call `sentry.Flush(...)` before returning the event ID resolves this issue.

I have added a provisional PR that resolves this here: https://github.com/getsentry/sentry-go/pull/31.
If you have ideas for other approaches, or would like to make the Flush time configurable, let me know there.

kamilogorek:
Recovered panics mean that they won't kill the process, thus it should be enough time to send the event.

However, if you want to control it yourself, you can catch the panic yourself and pass it through while awaiting the response, eg.:

```go
func() {
	defer func() {
		err := recover()
		if err != nil {
			switch v := err.(type) {
			case string:
				sentry.CaptureMessage(v)
			case error:
				sentry.CaptureException(v)
			}
			sentry.Flush(time.Second * 3)
		}
	}()

	throwPanic()
}()
```

or if you dont want to handle type-switching yourself, you can call `Recover` method on `Hub` directly, as it accepts any interface

```go
func() {
	defer func() {
		err := recover()
		if err != nil {
			sentry.CurrentHub().Recover(err)
			sentry.Flush(time.Second * 3)
		}
	}()

	throwPanic()
}()
```

DimitriosLisenko:
Thanks @kamilogorek. I ended up writing a variation similar to your first example yesterday.

Just to be clear though, I opened this ticket because I was not able to make `defer sentry.Recover()` work during my tests at all. I'm assuming this is because the time between when the event is asynchronously sent and when the process shuts down after the recovery procedure is not actually enough time to send the event. It could be something specific to my machine, but just so you're aware. :)

kamilogorek:
That's correct, everything is async by default. If you want to make sure that all events are delivered in sync manner (instead of using Flush), you can use `HTTPSyncTransport` described here https://docs.sentry.io/platforms/go/transports :)
Cheers!

PR #31, opened 2019-07-22
When recovering, make sure that there is enough time for the error to…

DimitriosLisenko:
… be sent


kamilogorek:
Answered in #30

PR #36, opened 2019-07-25
fix: Add RWMutex locks to internal Hub and Scope changes

kamilogorek:


hub.go
```diff
@@ -67,60 +70,98 @@ func (hub *Hub) LastEventID() EventID {
...
 
 func (hub *Hub) stackTop() *layer {
+	hub.RLock()
 	stack := hub.stack
 	if stack == nil || len(*stack) == 0 {
+		hub.RUnlock()
 		return nil
 	}
-	return (*stack)[len(*stack)-1]
+	top := (*stack)[len(*stack)-1]
```
bruno-garcia: can it throw here and leave it unlocked?
no `try/finaly`?

hub.go
```diff
@@ -67,60 +70,98 @@ func (hub *Hub) LastEventID() EventID {
...
 	top := hub.stackTop()
+	hub.RUnlock()
+
 	if top == nil {
 		return nil
 	}
-	return top.scope
+
+	top.RLock()
+	scope := top.scope
```
bruno-garcia: You're synchronizing reading the field but then giving the reference away. Is this intended?

hub.go
```diff
@@ -67,60 +70,98 @@ func (hub *Hub) LastEventID() EventID {
...
+	top := (*stack)[len(*stack)-1]
+	hub.RUnlock()
+
+	return top
 }
 
 // Clone returns a copy of the current Hub with top-most scope and client copied over.
 func (hub *Hub) Clone() *Hub {
-	return NewHub(hub.Client(), hub.Scope().Clone())
+	clone := NewHub(hub.Client(), hub.Scope().Clone())
```
bruno-garcia: assuming this is semantically correct, wouldn't it make sense to have the Read lock here once and the clones done under the lock? instead of calling two methods that will individually lock?

In between getting the client and scope, things might have changed.

hub.go
```diff
@@ -67,60 +70,98 @@ func (hub *Hub) LastEventID() EventID {
...
+	top.RUnlock()
+
+	return client
 }
 
 // PushScope pushes a new scope for the current `Hub` and reuses previously bound `Client`.
 func (hub *Hub) PushScope() *Scope {
-	scope := hub.Scope().Clone()
+	hub.RLock()
+	client := hub.Client()
```
bruno-garcia: Is the Go reader write lock happy with acquiring a read lock on a thread that already has one?
You'll get another read lock for `.Client()` and `.Scope()`.

hub.go
```diff
@@ -67,60 +70,98 @@ func (hub *Hub) LastEventID() EventID {
 }
 
 func (hub *Hub) stackTop() *layer {
+	hub.RLock()
 	stack := hub.stack
 	if stack == nil || len(*stack) == 0 {
+		hub.RUnlock()
```
HazAT: I have no idea about GO, but can we maybe use `defer`?
So this is valid for every function, so we always do:
```
hub.RLock()
defer hub.RUnlock()
```

scope.go
```diff
@@ -50,136 +50,117 @@ func NewScope() *Scope {
 // AddBreadcrumb adds new breadcrumb to the current scope
 // and optionaly throws the old one if limit is reached.
 func (scope *Scope) AddBreadcrumb(breadcrumb *Breadcrumb, limit int) {
+	scope.Lock()
```
bruno-garcia: Why moving this one up?
The scope of the lock was actually smaller with the previous approach (no need to lock the timestamp work)

scope.go
```diff
@@ -50,136 +50,117 @@ func NewScope() *Scope {
...
 	for k, v := range contexts {
 		scope.contexts[k] = v
 	}
-	scope.Unlock()
 }
 
 // RemoveContext removes a context from the current scope.
 func (scope *Scope) RemoveContext(key string) {
-	scope.Lock()
 	delete(scope.contexts, key)
```
bruno-garcia: Don't know how `delete` works in go but my guess is that it needs a writer lock here too.

scope.go
```diff
@@ -50,136 +50,117 @@ func NewScope() *Scope {
...
-	scope.Unlock()
 }
 
 // SetExtra adds an extra to the current scope.
 func (scope *Scope) SetExtra(key string, value interface{}) {
-	scope.Lock()
 	scope.extra[key] = value
```
bruno-garcia: Modifying a map, my guess is that u need a write lock for this collection too. Unless it's a thread-safe collection

Issue #44, opened 2019-08-20
Handle marshalling error


IzeBerg:
Transports don't handle marshaling errors and send empty events without any alert. The server just ignores requests like that. This means that messages are simply lost without any alerts.

It would be better if in case of an error it displayed in debug log or something like that.

https://github.com/getsentry/sentry-go/blob/64e024f37b61555761d98a3e41f43f143127e417/transport.go#L134
https://github.com/getsentry/sentry-go/blob/64e024f37b61555761d98a3e41f43f143127e417/transport.go#L260

Simple way to reproduce:
```
hub := sentry.CurrentHub().Clone()
hub.ConfigureScope(func(scope *sentry.Scope) {
	type test struct {
		UnsupportedField func()
	}
	scope.SetExtra(`test`, test{})
})
hub.CaptureMessage("broken event")
hub.Flush(5*time.Second)
```

PR #47, opened 2019-08-28
fix: Retry event marshalling without contextual data if the first pass fails

kamilogorek:
Fixes https://github.com/getsentry/sentry-go/issues/44

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


kamilogorek:
@IzeBerg wdyt about that change?

IzeBerg:
Maybe add info into the event that it changed? Like message into the log, but in the event.

kamilogorek:
@IzeBerg added the info

PR #52, opened 2019-09-04
fix: Include url.Parse error in DsnParseError

rhcarvalho:
Go 1.11.13 and 1.12.8 received backported fixes for security issues with
url.Parse. One of the fixes includes stricter port validation.

With those minor versions of Go running in Travis,
TestInvalidDsnInvalidPort fails because url.Parse returns an error when
it previously didn't, thus NewDsn returns &DsnParseError{"invalid url"}
instead of &DsnParseError{"invalid port"}.

Including the error from url.Parse in the message both provides more
context for SDK users to understand why their DSN string was refused,
and fixes the test because "invalid port" is part of the now included
error message from url.Parse.

For reference, the backports are:

- https://github.com/golang/go/issues/33632
- https://github.com/golang/go/issues/33633

### DIFF
--- a/dsn.go
+++ b/dsn.go
@@ -52,7 +52,7 @@ func NewDsn(rawURL string) (*Dsn, error) {
 	// Parse
 	parsedURL, err := url.Parse(rawURL)
 	if err != nil {
-		return nil, &DsnParseError{"invalid url"}
+		return nil, &DsnParseError{fmt.Sprintf("invalid url: %v", err)}
 	}
 
 	// Scheme


rhcarvalho:
This fixes the Travis builds for Go 1.11 and 1.12.

For Go 1.10, the problem is that `github.com/kataras/iris` at current HEAD is incompatible with Go 1.10 (uses `strings.ReplaceAll` added in Go 1.12). We could make Travis use an older version of iris, but it seems to be time to drop Go 1.10 and add Go 1.13 to Travis, in which case we can specify minimum version requirements with go.mod and the problem will go away.

Issue #55, opened 2019-09-04
Sentry client crashes entire process with "concurrent map writes" due to unsynchronized contextifyFramesIntegration


james-johnston-thumbtack:
We are attempting to use the same Sentry client in a server with many goroutines, as we assumed this client was thread-safe.  However, we're getting "concurrent map writes" errors.

1. Is the Sentry client intended to be thread-safe?  Specifically, is it intended to be safe to concurrently send events to the client returned by `sentry.NewClient`?

1. If yes, then we have found a bug: the `contextifyFramesIntegration` in `integrations.go` is not thread-safe.  In particular, it has a `cachedLocations` variable which is a map, and access to this variable is not synchronized.  We're getting stack traces from the concurrent map writes errors that point to reads/writes of this map.

I noticed that `sourceReader` has a similar caching mechanism, but that one is protected by a mutex....

We're using the Sentry client from commit 510f01e3c1e95d434331b518fdc62b2adaf5cff8, but I don't see anything fixed in the latest master, either.

kamilogorek:
@james-johnston-thumbtack thanks for the report. I'll try to investigate it this week.

PR #59, opened 2019-09-09
fix: Make sure that cache is locked when updating for contextifyFramesIntegration

kamilogorek:
Fixes https://github.com/getsentry/sentry-go/issues/55

### DIFF
--- a/integrations.go
+++ b/integrations.go
@@ -9,6 +9,7 @@ import (
 	"regexp"
 	"runtime"
 	"strings"
+	"sync"
 )
 
 // ================================
@@ -268,7 +269,7 @@ func getIgnoreErrorsSuspects(event *Event) []string {
 type contextifyFramesIntegration struct {
 	sr              sourceReader
 	contextLines    int
-	cachedLocations map[string]string
+	cachedLocations sync.Map
 }
 
 func (cfi *contextifyFramesIntegration) Name() string {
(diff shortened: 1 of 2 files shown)


integrations.go
```diff
@@ -374,3 +376,9 @@ func (cfi *contextifyFramesIntegration) addContextLinesToFrame(frame Frame, line
 	}
 	return frame
 }
+
+func (cfi *contextifyFramesIntegration) updateCachedLocation(key, val string) {
+	cfi.mu.Lock()
```
bruno-garcia: This is a write-lock. I can't spot a use of a read-lock. Is the reader-write really needed here as opposed to a plain mutex?

integrations.go
```diff
@@ -352,12 +354,12 @@ func (cfi *contextifyFramesIntegration) findNearbySourceCodeLocation(originalPat
 		possibleLocation := strings.Join(components, "/")
 
 		if fileExists(possibleLocation) {
-			cfi.cachedLocations[originalPath] = possibleLocation
+			cfi.updateCachedLocation(originalPath, possibleLocation)
```
bruno-garcia: You need a `cfi.mu.RLock()` here.
If this collection isn't thread-safe, we must make sure no read happens while we're writing into it. If I understand correctly you are guarding the writes to it via `cfi.mu.Lock()` but not the reads with `cfi.mu.RLock()`.

integrations_test.go
```diff
@@ -224,3 +225,15 @@ func TestContextifyFramesNonexistingFilesShouldNotDropFrames(t *testing.T) {
 	contextifiedFrames := cfi.contextify(frames)
 	assertEqual(t, len(contextifiedFrames), len(frames))
 }
+
+func TestContextifyFramesConcurrentCacheUpdates(t *testing.T) {
+	cfi := contextifyFramesIntegration{
```
bruno-garcia: Does this test repro the issue before the fix is applied?

kamilogorek: Yes. I made sure it fails first. Although I wasnt able to do this for reads, thus I didnt change them anywhere.

PR #65, opened 2019-10-05
Fix module integration

faulkner:
NewEvent() doesn't return Modules as nil, which was preventing this integration from working.
The extra flag was added to prevent repeatedly trying to cache the modules if getModules() returned nil.


faulkner:
The failing test should be resolved by https://github.com/getsentry/sentry-go/pull/66, but I can add it to this PR if that helps.

kamilogorek:
Thanks!

PR #72, opened 2019-10-24
accept nil EventModifier

rerorero:
* to avoid nil reference
* it's a bit bothering to pass it when using client directly


kamilogorek:
Hey, thanks for the patience. I've scheduled some time for patches/features and releasing a new version of sentry-go this and next week. I'll make sure to fix all the remaining issues till the EOD next week. Cheers!

Issue #62, opened 2019-09-30
Give more control by overriding http.Client or http.RoundTripper


yansal:
This issue is a feature request.

It's possible to override the http.Transport via the HTTPTransport field in ClientOptions.

Instead, it would be great to allow to override the http.Client, or http.RoundTripper (which is an abstract http.Transport). This feature would allow users of this package to add hooks at the http level.

kamilogorek:
Hey, thanks for the patience. I've scheduled some time for patches/features and releasing a new version of sentry-go this and next week. I'll make sure to fix all the remaining issues till the EOD next week. Cheers!

kamilogorek:
@yansal done. Will release new tag soon

PR #86, opened 2019-11-13
feat: Add HTTPClient option

kamilogorek:
Closes https://github.com/getsentry/sentry-go/issues/62

TODO: Update docs.sentry.io

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -143,9 +143,13 @@ func (t *HTTPTransport) Configure(options ClientOptions) {
 		}
 	}
 
-	t.client = &http.Client{
-		Transport: t.transport,
-		Timeout:   t.Timeout,
+	if options.HTTPClient != nil {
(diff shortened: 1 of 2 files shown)


transport.go
```diff
@@ -276,9 +280,13 @@ func (t *HTTPSyncTransport) Configure(options ClientOptions) {
 		}
 	}
 
-	t.client = &http.Client{
-		Transport: t.transport,
-		Timeout:   t.Timeout,
+	if options.HTTPClient != nil {
```
bruno-garcia: I'd suggest to try to refactor the two `Configure` functions somehow.
If I understood the async one has a buffer creation and the `start` call but the rest is the same, so maybe the two public methods could call a private one that prepares things and the async sets the buffer after the return and starts it or else

kamilogorek: Discussed the issue in person

PR #88, opened 2019-11-17
docs: Update links to event-playloads

steveazz:
Each exposed struct has a link to the docs that lead to a 404.


kamilogorek:
Thanks!

kamilogorek:
Thanks!

Issue #84, opened 2019-11-13
Request body logging isn't working for server requests


flash286:
Hello everyone,
Trying to use request body logging in some cases, but it doesn't seem to be working

```go
request := sentry.Request{}
request = request.FromHTTPRequest(req)
```
request.Data is always empty :(

I guess a problem can be here

```go
if request.GetBody != nil {
		if bodyCopy, err := request.GetBody(); err == nil && bodyCopy != nil {
			body, err := ioutil.ReadAll(bodyCopy)
			if err == nil {
				r.Data = string(body)
			}
		}
	}
```

request.GetBody is always == nil and [Doc](https://golang.org/pkg/net/http/#Request) says 
> For server requests, it is unused.

request.Body looks like a right field.

flash286:
up

aschade92:
I'm having the same issue

kamilogorek:
@flash286 @aschade92 I'll take a look at it in the morning

flash286:
@kamilogorek let me know if it's a bug. I'm ready to make PR.

rhcarvalho:
Hi @flash286! I can confirm this is a bug, thank you for reporting it 😄 

Please go ahead and send in a PR, I'll be happy to help you get that merged. Thanks!

PR #94, opened 2019-11-27
fix: request body logging [#84]

flash286:
fix: request body logging [#84]

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -1,6 +1,7 @@
 package sentry
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io/ioutil"
@@ -101,15 +102,15 @@ func (r Request) FromHTTPRequest(request *http.Request) Request {
 	r.QueryString = request.URL.RawQuery
 
 	// Body
-	if request.GetBody != nil {
-		if bodyCopy, err := request.GetBody(); err == nil && bodyCopy != nil {
-			body, err := ioutil.ReadAll(bodyCopy)
-			if err == nil {
-				r.Data = string(body)
-			}
+	if request.Body != nil {
+		bodyBytes, err := ioutil.ReadAll(request.Body)
+		_ = request.Body.Close()
+		if err == nil {
+			// We have to restore original state of *request.Body
+			request.Body = ioutil.NopCloser(bytes.NewBuffer(bodyBytes))
+			r.Data = string(bodyBytes)
 		}
 	}
-
 	return r
 }
 
--- a/interfaces_test.go
+++ b/interfaces_test.go
@@ -1,0 +1,37 @@
+package sentry
+
+import (
+	"bytes"
+	"io/ioutil"
+	"net/http"
+	"testing"
+)
+
+const testPayload = `{"test_data": true}`
+
+func TestRequest_FromHTTPRequest(t *testing.T) {
+	t.Run("reading_body", func(t *testing.T) {
+		payload := bytes.NewBufferString(testPayload)
+		req, err := http.NewRequest("POST", "/test/", payload)
+		assertEqual(t, err, nil)
+		assertNotEqual(t, req, nil)
+		sentryRequest := Request{}
+		sentryRequest = sentryRequest.FromHTTPRequest(req)
+		assertEqual(t, sentryRequest.Data, testPayload)
+	})
+
+	t.Run("reading_body_twice", func(t *testing.T) {
+		payload := bytes.NewBufferString(testPayload)
+		req, err := http.NewRequest("POST", "/test/", payload)
+		assertEqual(t, err, nil)
+		assertNotEqual(t, req, nil)
+		sentryRequest := Request{}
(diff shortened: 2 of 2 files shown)


rhcarvalho:
@flash286 thank you for the pull request! The approach looks good.

I would like to investigate a bit more before merging. In particular, look at the integrations that use FromHTTPRequest -- it may be the case that the body was read and is not available anymore when we want to report an error.

Will give more review feedback tomorrow, wanted to keep you posted ;)

Once again thank you!

interfaces.go
```diff
@@ -101,15 +102,15 @@ func (r Request) FromHTTPRequest(request *http.Request) Request {
...
 	// Body
-	if request.GetBody != nil {
-		if bodyCopy, err := request.GetBody(); err == nil && bodyCopy != nil {
-			body, err := ioutil.ReadAll(bodyCopy)
-			if err == nil {
-				r.Data = string(body)
-			}
+	if request.Body != nil {
+		bodyBytes, err := ioutil.ReadAll(request.Body)
+		_ = request.Body.Close()
```
rhcarvalho: Just my curiosity, why did you add the `_ =` part?

---

According to https://godoc.org/net/http#Request:

> ```go 
> Request {
>     //...
>     // Body is the request's body.
>     //
>     // For client requests, a nil body means the request has no
>     // body, such as a GET request. The HTTP Client's Transport
>     // is responsible for calling the Close method.
>     //
>     // For server requests, the Request Body is always non-nil
>     // but will return EOF immediately when no body is present.
>     // The Server will close the request body. The ServeHTTP
>     // Handler does not need to.
>     Body io.ReadCloser
> ```

I might be missing something, but I think we don't need to close the body here, the Server will do.

kamilogorek:
Agree with all your comments, nice review!

interfaces_test.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"bytes"
+	"io/ioutil"
+	"net/http"
+	"testing"
+)
+
+const testPayload = `{"test_data": true}`
+
+func TestRequest_FromHTTPRequest(t *testing.T) {
```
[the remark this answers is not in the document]
flash286: fixed

interfaces_test.go
```diff
@@ -0,0 +1,37 @@
...
+	t.Run("reading_body", func(t *testing.T) {
+		payload := bytes.NewBufferString(testPayload)
+		req, err := http.NewRequest("POST", "/test/", payload)
+		assertEqual(t, err, nil)
+		assertNotEqual(t, req, nil)
+		sentryRequest := Request{}
+		sentryRequest = sentryRequest.FromHTTPRequest(req)
+		assertEqual(t, sentryRequest.Data, testPayload)
+	})
```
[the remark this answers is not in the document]
flash286: I Agree, second one should cover all we need

interfaces_test.go
```diff
@@ -0,0 +1,37 @@
+package sentry
+
+import (
+	"bytes"
+	"io/ioutil"
+	"net/http"
+	"testing"
+)
+
+const testPayload = `{"test_data": true}`
```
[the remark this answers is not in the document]
flash286: fixed

interfaces.go
flash286: > why did you add the _ = part?

Just to let linter know that i don't want to handle error here, if Body reader already closed
> I might be missing something, but I think we don't need to close the body here, the Server will do.

Hmm, i wasn't sure about it. But we're overwriting the original body reader and server will close the new one, it seems like resource leak

rhcarvalho: > we're overwriting the original body reader and server will close the new one, it seems like resource leak

You're right! Sorry, my mistake.

interfaces.go
```diff
@@ -101,15 +102,15 @@ func (r Request) FromHTTPRequest(request *http.Request) Request {
...
+		if err == nil {
```
rhcarvalho: If `err != nil`, it might still be possible that some data was read.

> [ReadAll](https://golang.org/pkg/io/ioutil/#ReadAll) reads from r until an error or EOF and returns the data it read.

Wouldn't it be more beneficial to always create the buffer with the data that was read, and only log the error?

E.g. the client could abruptly close the connection while the server is reading the body.

rhcarvalho: Tracking improvement in #117.

rhcarvalho:
@flash286 thank you very much for your contribution!

interfaces.go
faulkner: This introduces a new bug. If `ioutil.ReadAll` returns an error the body is left in a closed state, so reading the body later results in a `http.ErrBodyReadAfterClose`, so instead of the original error getting sent to Sentry you end up just getting `http: invalid Read on closed Body` with no way of knowing what the original error was.

rhcarvalho: @faulkner solid. I'll take this into account when I work on #117, which should be some time next week.
Contributions welcome as well, just send us a patch if you have one. Thanks!

PR #102, opened 2019-12-03
Add support for path prefixes in the DSN

zerok:
Port of a [feature from the JavaScript SDK](https://github.com/getsentry/sentry-javascript/blob/645ff532941d9a23bb6531c6305e99544b356502/packages/core/src/api.ts#L48) which supports DSNs with a
custom path prefix (e.g. http://user@host/prefixed/42).


kamilogorek:
Thanks!

zerok:
Thank *you* :)

PR #121, opened 2019-12-13
ci(travis): Bump golangci-lint version

rhcarvalho:
The current recommended installation steps use githubusercontent.com
instead of goreleaser.com.

Additionally, pin to a specific tagged version of the install script to
ensure reproducible installations of the tool.

Use latest available version 1.21.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_install:
   - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/v1.19.1/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.19.1
 
 script:
-  - golangci-lint run
+  - golangci-lint run --new-from-rev=$(git merge-base origin/master HEAD)
   - go build
   - go test
   - go test -race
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -10,10 +10,10 @@ env:
   - GO111MODULE=on GOFLAGS=-mod=readonly
 
 before_install:
-  - curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(go env GOPATH)/bin v1.15.0
+  - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/v1.19.1/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.19.1
```
rhcarvalho: FWIW

@kamilogorek this bump introduces new linters that would cause the build to fail 😞 
I wouldn't want to please linters right now, some are even questionable. So the solution would be to make future code more strict by using the `--new-from-rev` flag.

I also had to give up on v1.20.0 and v1.21.0 as detailed in the commit message.

kamilogorek: Makes sense, thanks for updating that

Issue #100, opened 2019-12-02
Failing tests when sentry-go is outside of GOPATH


rhcarvalho:
When the package is checked out outside of GOPATH, some tests fail:

```
$ go test
--- FAIL: TestContextifyFrames (0.00s)
    integrations_test.go:181:
        got: []string{"}", "", "// NewStacktrace creates a stacktrace using `runtime.Callers`.", "func NewStacktrace() *Stacktrace {", "\tpcs := make([]uintptr, 100)"}
        want: []string{")", "", "// NOTE: if you modify this file, you are also responsible for updating LoC position in Stacktrace tests", "", "func Trace() *Stacktrace {"}
    integrations_test.go:188:
        got: "\tn := runtime.Callers(1, pcs)"
        want: "\treturn NewStacktrace()"
    integrations_test.go:189:
        got: []string{"", "\tif n == 0 {", "\t\treturn nil", "\t}", ""}
        want: []string{"}", "", "func RedPkgErrorsRanger() error {", "\treturn BluePkgErrorsRanger()", "}"}
--- FAIL: TestNewStacktrace (0.00s)
    stacktrace_test.go:32:
        got: 3
        want: 2
--- FAIL: TestStacktraceFrame (0.00s)
    stacktrace_test.go:50:
        got: "stacktrace.go"
        want: "errors_test.go"
    stacktrace_test.go:51:
        got: "/Users/rodolfo/sentry/sentry-go/stacktrace.go"
        want: "/Users/rodolfo/sentry/sentry-go/errors_test.go"
    stacktrace_test.go:52:
        got: "NewStacktrace"
        want: "Trace"
    stacktrace_test.go:53:
        got: 36
        want: 12
FAIL
exit status 1
FAIL    github.com/getsentry/sentry-go  0.232s
```

PR #122, opened 2019-12-13
ci(travis): Test more env and Go combinations

rhcarvalho:
- Test against go@master to help us verify if the SDK will work with a
new Go release before the release is shipped. Failed builds are allowed,
i.e., they do not block PRs.

- Test with GO111MODULE=off and GO111MODULE=on, as our users could be
using Go Modules or not.

- Test code outside of GOPATH. It should be okay to work on sentry-go
either inside or outside of GOPATH (Module mode). *Temporarily* marked
to allow failures. This reproduces #100 in our CI environment.

Build matrix increases from 3 jobs to 2*4+1 = 9 jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,33 @@
 sudo: false
 
 language: go
+
 go:
   - 1.11.x
   - 1.12.x
   - 1.13.x
+  - master
 
 env:
   - GO111MODULE=on GOFLAGS=-mod=readonly
+  - GO111MODULE=off
+
+jobs:
+  include:
+    - name: "Module support outside of GOPATH"
+      go: 1.13.x
+      before_script: >-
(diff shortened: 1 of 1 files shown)


rhcarvalho:
This still needs some tweaking to get `sentry-go` outside of `GOPATH` in Travis.

rhcarvalho:
@kamilogorek I'm sorry for the mess here. I forgot to mark the PR as draft, in retrospect it would probably have been easier to test it in a branch.

The main intention here is to reproduce #100 in Travis, so that when we ship a fix we also have a way to prevent regression. Looking forward to one day removing these artifacts connected with module x not-a-module!

This is the failing build that reproduces #100: https://travis-ci.com/getsentry/sentry-go/jobs/267002354.
As a follow up, I'll remove the entry in `allow_failures` and hopefully make a viable fix for #100.

PR #123, opened 2019-12-13
fix(stacktrace): Do not rely on AbsPath for filtering

rhcarvalho:
Fixes #100

PR #125, opened 2019-12-17
fix(stacktrace): Filter out internal SDK methods

rhcarvalho:
The frame.Module field in its current form may contain entries like:
github.com/getsentry/sentry-go.(*Hub)
github.com/getsentry/sentry-go.(*Client)
github.com/getsentry/sentry-go.(*Client)

Taking those into account now.

Fixes up #123.

### DIFF
--- a/stacktrace.go
+++ b/stacktrace.go
@@ -206,7 +206,8 @@ func filterFrames(frames []Frame) []Frame {
 			continue
 		}
 		// sentry internal frames
-		if frame.Module == "github.com/getsentry/sentry-go" {
+		if frame.Module == "github.com/getsentry/sentry-go" ||
+			strings.HasPrefix(frame.Module, "github.com/getsentry/sentry-go.") {
 			continue
 		}
 		filteredFrames = append(filteredFrames, frame)
--- a/stacktrace_external_test.go
+++ b/stacktrace_external_test.go
@@ -104,6 +104,9 @@ func TestNewStacktrace(t *testing.T) {
 				},
 			},
 		}},
+		"StacktraceTestHelper": {sentry.StacktraceTestHelper{}.NewStacktrace, &sentry.Stacktrace{
+			Frames: []sentry.Frame{},
+		}},
 	}
 	for name, tt := range tests {
 		t.Run(name, func(t *testing.T) {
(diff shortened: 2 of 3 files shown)


stacktrace.go
```diff
@@ -206,7 +206,8 @@ func filterFrames(frames []Frame) []Frame {
 			continue
 		}
 		// sentry internal frames
-		if frame.Module == "github.com/getsentry/sentry-go" {
+		if frame.Module == "github.com/getsentry/sentry-go" ||
```
kamilogorek: We can probably just unify this to a single `strings.HasPrefix` and remove ending dot

kamilogorek: Or `strings.Contains`. The chance of false positive is very unlikely.

rhcarvalho: I considered using only strings.HasPrefix. Eventually, what we need to do is define what frame.Module is supposed to mean.
Quite likely it should be github.com/getsentry/sentry-go, and never github.com/getsentry/sentry-go.(*Hub). Then the direct string comparison with == would be the condition here. What do you think?

kamilogorek: Double-checked it and agree. It should be direct comparison.
But why is `deconstructFunctionName` putting `.(Hub)` in `frame.Module`? It should be a part of a function name. And I'm pretty sure it works correctly, at least for the input that's supposedly incorrect - https://play.golang.org/p/U8SfYCLeyun

rhcarvalho: The input is different, see https://play.golang.org/p/bA6FSn1qG8C

It returns the Method name without the type prefix, and the type goes into the package import path.

kamilogorek: That's actually good, as we can reproduce it now. So what should change is the way to determine which parts belong to the module and which to the function name. Buuut it'll probably require a regexp instead of simple `.` index search 😢

rhcarvalho: We didn't touch `deconstructFunctionName` recently. I can add some new test cases and fix it/change the behavior in a separate PR. Then we can drop the `|| HasPrefix` part :)

rhcarvalho: See #126

rhcarvalho: > Quite likely it should be github.com/getsentry/sentry-go, and never github.com/getsentry/sentry-go.(*Hub). Then the direct string comparison with == would be the condition here.

Except that we have "subpackages": `github.com/getsentry/sentry-go/http`, `github.com/getsentry/sentry-go/...`.

I was wrong, forgot about those for a moment :(

Issue #75, opened 2019-10-30
Big distribution size of sentry-go (~ 8MB)


lozenko:
Hi,

Sentry-go has almost 40 indirect dependencies
https://github.com/getsentry/sentry-go/blob/master/go.mod

which makes a binary package for primitive < 5MB applications quite big

is there a way to work around this, (some sort of lightweight minimal client ?) or do we have HTTP client for Sentry that we can use? thanks

kamilogorek:
Hey, these dependencies shouldn't be bundled in the binary unless their respective integrations (http packages) are imported. Can you provide a minimal example of a build where I can repro this?

lozenko:
Hi, kamilogorek !

lozenko:
Hi, kamilogorek

Thanks for watching this. Reproduce steps:
1) add sentry-go to hello world app with go.mod
2) go mod download && go build the app itself
3) observe the package size

It's super visible when you do it in docker when go mod download all direct & indirect packages  and buld the executable file.

kamilogorek:
I verified that and the size is not caused by the external package, but by the number of built-in packages we use.

![without](https://user-images.githubusercontent.com/1523305/68682146-f005c580-0564-11ea-8ed7-f3f32e34c7a5.png)

Unless you are using specific integration, it won't be included in the build.

If you'll create an empty project without _any_ external dependency, and build it with the same built-in packages, you'll get almost the same size:

```go
package main

import (
	_ "bufio"
	_ "bytes"
	_ "context"
	_ "crypto/rand"
	_ "crypto/tls"
	_ "crypto/x509"
	_ "encoding/hex"
	_ "encoding/json"
	_ "fmt"
	_ "go/build"
	_ "io"
	_ "io/ioutil"
	_ "log"
	_ "math/rand"
	_ "net"
	_ "net/http"
	_ "net/url"
	_ "os"
	_ "path/filepath"
	_ "reflect"
	_ "regexp"
	_ "runtime"
	_ "sort"
	_ "strconv"
	_ "strings"
	_ "sync"
	_ "time"
)

func main() {}
```

```sh
-rwxr-xr-x  1 kamilogorek  staff   8.9M Nov 12 16:05 hello-world
```

rur:
Overhead may or may not end up in the artifact but it will spam my project dependencies, which I care about. 

Please read https://research.swtch.com/deps

```
[...]

PR #127, opened 2019-12-18
fix(stacktrace): Correctly report package names and ignore subpackages

### LINKED ISSUE #126: stacktrace: deconstructFunctionName returns bad output for methods
`deconstructFunctionName` puts `.(Hub)` (receiver) in `frame.Module`, but it should be a part of the function (method) name.

_Originally posted by @kamilogorek in https://github.com/getsentry/sentry-go/pull/125/files#r359295441_

rhcarvalho:
Makes function receivers part of the function name and not part of the package name.

Add tests and fix `filteredFrames` to ensure it ignores all packages related to the SDK.

Fixes #126.

### DIFF
--- a/stacktrace.go
+++ b/stacktrace.go
@@ -149,7 +149,7 @@ func NewFrame(f runtime.Frame) Frame {
 	abspath := f.File
 	filename := f.File
 	function := f.Function
-	var module string
+	var pkg string
 
 	if filename != "" {
 		filename = filepath.Base(filename)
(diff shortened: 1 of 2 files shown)


stacktrace.go
```diff
@@ -206,8 +215,7 @@ func filterFrames(frames []Frame) []Frame {
 			continue
 		}
 		// sentry internal frames
-		if frame.Module == "github.com/getsentry/sentry-go" ||
-			strings.HasPrefix(frame.Module, "github.com/getsentry/sentry-go.") {
+		if frame.Module == "github.com/getsentry/sentry-go" {
```
untitaker: Will this be `github.com/getsentry/sentry-go/fasthttp` for integrations? I feel like it should be if we want to be close to Python.

stacktrace.go
```diff
@@ -226,21 +234,42 @@ func isInAppFrame(frame Frame) bool {
...
 	callerFrame, _ := callersFrames.Next()
-	_, function := deconstructFunctionName(callerFrame.Function)
-	return function
+	return baseName(callerFrame.Function)
+}
+
+// packageName returns the package part of the symbol name, or the empty string
+// if there is none.
+// It replicates https://golang.org/pkg/debug/gosym/#Sym.PackageName, avoiding a
+// dependency on debug/gosym.
```
rhcarvalho: This is intentional, because:

1. https://go-proverbs.github.io, ["A little copying is better than a little dependency."](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s)
2. Avoids increasing the size of binaries that include the SDK (see #75)

stacktrace.go
rhcarvalho: See https://github.com/getsentry/sentry-go/pull/125#discussion_r359278184

rhcarvalho: Yes it will be, and we should filter it out -- what we don't :( Yet.

I realize #125 was incomplete, we need to filter out frames of all packages belonging to the SDK (keeping the behavior of the latest released version).

Thanks for asking! I'll fix it.

rhcarvalho: Fixed in the latest patch.

PR #133, opened 2019-12-29
Fix TestConcurrentHubClone

### LINKED ISSUE #132: TestConcurrentHubClone failure
https://travis-ci.com/getsentry/sentry-go/jobs/270853268

```
$ go test -race
--- FAIL: TestConcurrentHubClone (0.00s)
    hub_test.go:404: 
        got: []sentry.TestEvent{sentry.TestEvent{Message:"Hello from goroutine! #2", Tags:map[string]string{"secretTag":"go#2"}}, sentry.TestEvent{Message:"Hello from goroutine! #3", Tags:map[string]string{"secretTag":"go#3"}}}
        want: []sentry.TestEvent{sentry.TestEvent{Message:"Hello from goroutine! #1", Tags:map[string]string{"secretTag":"go#1"}}, sentry.TestEvent{Message:"Hello from goroutine! #2", Tags:map[string]string{"secretTag":"go#2"}}, sentry.TestEvent{Message:"Hello from goroutine! #3", Tags:map[string]string{"secretTag":"go#3"}}}
FAIL
exit status 1
FAIL	github.com/getsentry/sentry-go	0.063s
```

rhcarvalho:
Temporarily binding the hub client to nil introduces the (unintended)
possibility of dropping events, as Hub.CaptureMessage and related
methods bail when their client is nil.

It was possible to reproduce the issue by running the tests many times
in a row with a command like:

go test -race -count 500 -failfast -run TestConcurrentHubClone

Fixes #132

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -12,3 +12,7 @@ issues:
   exclude:
     - "not declared by package utf8"
     - "unicode/utf8/utf8.go"
+  exclude-rules:
+    - path: _test\.go
+      linters:
+        - prealloc
--- a/hub_test.go
+++ b/hub_test.go
@@ -3,9 +3,11 @@ package sentry
 import (
 	"context"
 	"fmt"
-	"sort"
 	"sync"
 	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 )
 
 func setupHubTest() (*Hub, *Client, *Scope) {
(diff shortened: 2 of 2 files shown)


hub_test.go
```diff
@@ -390,16 +391,17 @@ func TestConcurrentHubClone(t *testing.T) {
 		},
 	}
 
-	got := make([]TestEvent, len(transport.Events()))
-	for i, event := range transport.Events() {
-		got[i] = TestEvent{
+	var got []TestEvent
+	for _, event := range transport.Events() {
+		got = append(got, TestEvent{
```
rhcarvalho: This is a rather stylistic change that requires the accompanying ignore-rule in the linter config.

The more typical idiom in Go is to declare a slice and `append` to it in the loop. Initializing with `make` is unnecessary, perhaps tending to a premature optimization (as is [documented in the very linter rule that likely made this line get this shape](https://github.com/golangci/golangci-lint/blob/d3e36a97cdc1c68566b0a274df367a44cd766cfb/.golangci.example.yml#L172-L174)).

.golangci.yml
```diff
@@ -12,3 +12,7 @@ issues:
   exclude:
     - "not declared by package utf8"
     - "unicode/utf8/utf8.go"
+  exclude-rules:
+    - path: _test\.go
+      linters:
+        - prealloc
```
rhcarvalho: See https://github.com/golangci/golangci-lint/blob/d3e36a97cdc1c68566b0a274df367a44cd766cfb/.golangci.example.yml#L172-L174

PR #140, opened 2020-01-21
fix: HTTPTransport.Flush panic and data race

### LINKED ISSUE #103: Intermittently seeing "sync: WaitGroup is reused before previous Wait has returned" in transport.go
I've seen the panic `"sync: WaitGroup is reused before the previous Wait has returned"` in a routine at `github.com/getsentry/sentry-go/transport.go:201`. Not sure how to debug this or how to explore the conditions. I can't find anything in the documentation that can help me explore the conditions post-panic. The stack trace leads me to...

```
vendor/github.com/getsentry/sentry-go/transport.go:201 +0x31 
github.com/getsentry/sentry-go.(*HTTPTransport).Flush 
vendor/github.com/getsentry/sentry-go/transport.go:200 +0x7c 
```

Any advice is appreciated on tools I can use to better break this open when it does occur!

rhcarvalho:
Rewrite HTTPTransport internals to remove a data race and occasional
panics.

Prior to this, HTTPTransport used a `sync.WaitGroup` to track how many
in-flight requests existed, and `Flush` waited until the observed number
of in-flight requests reached zero.

Unsynchronized access to the WaitGroup lead to panics (reuse before
wg.Wait returns) and data races (undefined order of wg.Add and wg.Wait
calls). See https://github.com/getsentry/sentry-go/issues/103#issuecomment-574376825, https://github.com/getsentry/sentry-go/issues/103#issuecomment-574883037 and  https://github.com/getsentry/sentry-go/issues/103#issuecomment-574885317.

The new implementation changes the `Flush` behavior to wait until the
current in-flight requests are processed, inline with other SDKs and the
Unified API.

Fixes #103.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -96,16 +96,25 @@ func getRequestBodyFromEvent(event *Event) []byte {
 // HTTPTransport
 // ================================
 
+// A batch groups items that are processed sequentially.
+type batch struct {
+	items   chan *http.Request
+	started chan struct{} // closed to signal items started to be worked on
+	done    chan struct{} // closed to signal completion of all items
+}
+
 // HTTPTransport is a default implementation of `Transport` interface used by `Client`.
 type HTTPTransport struct {
 	dsn       *Dsn
 	client    *http.Client
 	transport *http.Transport
 
-	buffer        chan *http.Request
+	// buffer is a channel of batches. Calling Flush terminates work on the
+	// current in-flight items and starts a new batch for subsequent events.
+	buffer chan batch
+
 	disabledUntil time.Time
 
-	wg    sync.WaitGroup
 	start sync.Once
 
 	// Size of the transport buffer. Defaults to 30.
@@ -130,9 +139,17 @@ func (t *HTTPTransport) Configure(options ClientOptions) {
 		Logger.Printf("%v\n", err)
 		return
 	}
-
 	t.dsn = dsn
-	t.buffer = make(chan *http.Request, t.BufferSize)
+
+	// A buffered channel with capacity 1 works like a mutex, ensuring only one
+	// goroutine can access the current batch at a given time. Access is
+	// synchronized by reading from and writing to the channel.
+	t.buffer = make(chan batch, 1)
+	t.buffer <- batch{
+		items:   make(chan *http.Request, t.BufferSize),
+		started: make(chan struct{}),
+		done:    make(chan struct{}),
+	}
 
 	if options.HTTPTransport != nil {
 		t.transport = options.HTTPTransport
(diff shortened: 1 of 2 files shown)


rhcarvalho:
There are no automated tests covering the `HTTPTransport.Flush` behavior. In comments of #103 I posted what test code I used to guide this PR, however that's not a very good test to have in the repo.

Would love some ideas in that regard.

Thinking of testing some properties of Flush, like:

- After sending a few events and calling `Flush` those events should show up in a test server. Note that if too many events are sent, some will be dropped because of overflow.
- Calling `Flush` multiple times in a row should work and "be quick".
- ...

rhcarvalho:
I've included a few tests to exercise `HTTPTransport` usage flows.

transport.go
```diff
@@ -178,10 +192,10 @@ func (t *HTTPTransport) SendEvent(event *Event) {
 		request.Header.Set(headerKey, headerValue)
 	}
 
-	t.wg.Add(1)
+	b := <-t.buffer
```
rhcarvalho: This works as acquiring a lock to access the current `batch`.
A few lines below, `t.buffer <- b` is like releasing the lock.

We must hold the lock during the `select` block below, to guarantee that `b.items` is not closed. Sending on a closed channel panics.

The `select` block "should be quick" (takes a bounded amount of CPU time), because of the `default` case that is executed in case sending on `b.items` would block.

transport.go
```diff
@@ -130,9 +139,14 @@ func (t *HTTPTransport) Configure(options ClientOptions) {
 		Logger.Printf("%v\n", err)
 		return
 	}
-
 	t.dsn = dsn
-	t.buffer = make(chan *http.Request, t.BufferSize)
+
+	t.buffer = make(chan batch, 1)
```
rhcarvalho: The buffered channel with capacity 1 works as a mutex, ensuring only one goroutine can access the current batch at a given time.

transport.go
```diff
@@ -190,51 +204,87 @@ func (t *HTTPTransport) SendEvent(event *Event) {
...
 
 // Flush notifies when all the buffered events have been sent by returning `true`
 // or `false` if timeout was reached.
 func (t *HTTPTransport) Flush(timeout time.Duration) bool {
-	c := make(chan struct{})
+	toolate := time.After(timeout)
+
+	var b batch
+	for {
+		// Wait until processing the current batch has started or the timeout.
```
rhcarvalho: We must wait until the `worker` has seen the current batch, because it is the only way `b.done` will be closed.
If we do not wait, there is a possible execution flow in which `b.done` is never closed, and the only way out of Flush would be waiting for the time out.

rhcarvalho:
Added more code comments describing how the mechanism works and what invariants are considered.

Renamed and reorganized test code based on code review in person with @bruno-garcia. Thanks @bruno-garcia!!!

PR #145, opened 2020-01-24
ci: Bump golangci-lint to v1.27.0

ferhatelmas:
While doing #144, realized it was behind a couple of versions so bumped.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ jobs:
   fast_finish: true
 
 before_install:
-  - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.27.0
+  - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/v1.27.0/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.27.0
   # Fetch origin/master. This is required for `git merge-base` when testing a
   # branch, since Travis clones only the target branch.
   - git fetch origin master:remotes/origin/master
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ jobs:
   fast_finish: true
 
 before_install:
-  - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.23.1
+  - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.27.0
   # Fetch origin/master. This is required for `git merge-base` when testing a
   # branch, since Travis clones only the target branch.
   - git fetch origin master:remotes/origin/master
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ jobs:
   fast_finish: true
 
 before_install:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,5 +1,3 @@
-sudo: false
```
rhcarvalho: 👍 

`sudo: false` is deprecated for over a year now. Thanks for the clean up!

Reference:
https://docs.travis-ci.com/user/reference/trusty/#container-based-infrastructure
https://blog.travis-ci.com/2018-10-04-combining-linux-infrastructures
https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration#timeline---its-happening-fast

.travis.yml
```diff
@@ -29,7 +27,7 @@ jobs:
   fast_finish: true
 
 before_install:
-  - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/v1.19.1/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.19.1
+  - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.23.1
```
rhcarvalho: - We pin the version of the install script aiming for reproducible installs.
- We are stuck with v1.19 because of (still) open issues in golangci-lint.

The last commit on this file has the full details, see https://github.com/getsentry/sentry-go/commit/df74221585995c6f41deac769f9f9bb0e4d02c0c.

I'm afraid it is not time to bump the version just yet. See the `golangci-lint` build problem happened in this very PR: https://travis-ci.com/getsentry/sentry-go/jobs/279652187.

ferhatelmas: I see it's related to `GO111MODULE=off` with `1.11` version. Let's keep this open for the right time.

.travis.yml
ferhatelmas: extracted to #146

rhcarvalho:
I don't see us bumping `golangci-lint` anytime soon, and there is also no great need to do so. The existing/configured linters do what they need to do.

That said, I'm considering closing this PR to clean up our backlog. Will come back here in a week or so.

Anyone please comment if I'm missing anything in my reasoning. Thanks!

ernado:
Most of flaky issues with `golangci-lint` should be fixed now (current version is `v1.27.0`), probably you can reconsider version bump.

ferhatelmas:
Yup, I will bump it later today.

ferhatelmas:
@rhcarvalho updated. It's using install script from master, do we want to keep it as it is ?

ernado:
IMO it is better to use `v1.27.0` version of install script to install `v1.27.0` :)

Btw now we have https://github.com/marketplace/actions/run-golangci-lint

ferhatelmas:
@ernado Updated version but action adaption is nice and good to do but I think, out of scope this PR.

scope_test.go
```diff
@@ -586,9 +586,10 @@ func TestEventProcessorsModifiesEvent(t *testing.T) {
 
 	if processedEvent == nil {
 		t.Error("event should not be dropped")
+	} else {
+		assertEqual(t, LevelFatal, processedEvent.Level)
+		assertEqual(t, []string{"wat"}, processedEvent.Fingerprint)
 	}
-	assertEqual(t, LevelFatal, processedEvent.Level)
-	assertEqual(t, []string{"wat"}, processedEvent.Fingerprint)
```
rhcarvalho, line 591: The fix here is to use `t.Fatal` above. I had that in a local branch, merged to master now.

rhcarvalho:
Thanks @ferhatelmas and @ernado!

scope_test.go
ferhatelmas: LGTM, thanks!

PR #149, opened 2020-01-27
ci: Update craft config

rhcarvalho:
The new statusProvider config option was added in craft@0.9.0 and
allows us to remove a dependency on zeus.ci for build status checks.

The includeNames option skips artifact fetching. Reference:
https://github.com/getsentry/craft/blob/59486ba128c3aa24e87e585a6a0b87667fc5b4c0/src/targets/base.ts#L67

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -1,10 +1,14 @@
+minVersion: '0.9.0'
 github:
   owner: getsentry
   repo: sentry-go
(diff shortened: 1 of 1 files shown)


rhcarvalho:
This is 100% based on https://github.com/getsentry/sentry-php/commit/7acdcdd1d507e29a38b4c282527a2ae3c9fca130

PR #150, opened 2020-01-27
Add badges to GoDoc and go.dev

rhcarvalho:
GoDoc has the latest docs (master), go.dev has docs for all released
versions.


rhcarvalho:
Outcome, two new badges:

image

Issue #106, opened 2019-12-03
Improve documentation of sentry.Flush


rhcarvalho:
The documentation in https://godoc.org/github.com/getsentry/sentry-go#Flush could be more specific to help users understand when `Flush` needs to be called, if at all.

Similarly, the example code in https://docs.sentry.io/platforms/go/ can be improved to follow the same line.

Typically `Flush` should only be called once from the main goroutine when a program is terminating.

---

Subtasks:

- [x] Update README.md -- #152
- [x] Update documentation in https://godoc.org/github.com/getsentry/sentry-go#Flush -- #153
- [x] Update example in https://docs.sentry.io/platforms/go/ -- https://github.com/getsentry/sentry-docs/pull/1482

rhcarvalho:
Tracking issue to help users understand better how to use the SDK and avoid cases like https://github.com/getsentry/sentry-go/issues/103#issuecomment-561379867
> I've been using flush on every error I wanted to send to Sentry

mmacheerpuppy:
Thanks @rhcarvalho I think this line in the example code is where I derived the confusion

```
	if err != nil {
		sentry.CaptureException(err)
		sentry.Flush(time.Second * 5)
	}
```

rhcarvalho:
I randomly run into this code looking at Reddit r/golang today: [gist-cleaner/backend/main.go#L312-L317](https://github.com/NdoleStudio/gist-cleaner/blob/74a98f01dd4afc447ace1c620d6f0a53df39f7dc/backend/main.go#L312-L317)

The usage of `time.Second * 5` (more commonly written as `5 * time.Second`) matches the docs in https://docs.sentry.io/platforms/go/#usage and reinforces the importance of clarifying the docs.

I plan to solve this in the coming weeks by implementing `sentry.Close` (#111) and improving the docs (this issue). Both improvements should land in [v0.5.0](https://github.com/getsentry/sentry-go/milestone/1).

rhcarvalho:
@mmacheerpuppy @BonnieMilian I acknowledge you might be busy.

Given that you both ran into #103, I'd like to share that I've rewritten the example code and usage steps documented in the README.md of the Sentry Go SDK.

I would appreciate feedback if you have any! The changes are live here: https://github.com/rhcarvalho/sentry-go/blob/readme/README.md#usage

Hope to clear out any confusion with regards to `sentry.Flush`.

Thanks a lot!

mmacheerpuppy:
[...]

Issue #111, opened 2019-12-06
Unified API Client.Close()


rhcarvalho:
The [Shutdown and Draining](https://docs.sentry.io/error-reporting/configuration/draining/?platform=go) docs page for Go suggests using `sentry.Flush` where other SDKs use `Close`.

Then it says:

> After shutdown the client cannot be used any more so make sure to only do that right before you shut down the application.

That is incorrect docs, since it doesn't match what `Flush` does.

We should implement `Close` and fix the docs to use it.

As seen in recent issues, `Flush` is often not the answer and is being misused.

kamilogorek:
Agree, I missed that. `Close` is basically `Flush` + `enabled = false`

rhcarvalho:
Close would also involve releasing resources. Like `os.File.Close`, `http.Response.Body.Close` or `context.CancelFunc` do.

Potential signatures:

1. Explicit timeout argument

    ```
    // Close shuts down the SDK and blocks until all outstanding events have been
    // sent or the timeout is reached. Unlike Flush, Close releases resources
    // associated with the SDK. Code should typically defer a call to Close after
    // Init.
    func Close(timeout time.Duration) {}
    ```

2. Timeout from `Client.Options`

    Some SDKs (e.g. sentry-dotnet) take the timeout from configuration.

    ```
    // Close shuts down the SDK and blocks until all outstanding events have been
    // sent or ShutdownTimeout is reached. Unlike Flush, Close releases resources
    // associated with the SDK. Code should typically defer a call to Close after
    // Init.
    func Close() {}
    ```

    Alternatively, we can implement `io.Closer`:

    ```
    func Close() error {}
    ```

rhcarvalho:
While this is part of the Unified SDK spec, this is not a critical feature.

Originally I wanted to bundle this together with improvements to Flush and updating docs.
Now I realize we do not need to add this *right now*, we might as well take the time to learn what would be the best signature if we end up adding it later.

Considering recent conversations with @bruno-garcia and having in mind his influence to remind me that ["when in doubt, leave it out"](https://www.youtube.com/watch?v=aAb7hSCtvGw), I decided to leave it out for now.

Correct programs can be written using `sentry.Flush`.

[...]

PR #153, opened 2020-01-28
docs: Update Flush documentation

rhcarvalho:
Related to #106.

### DIFF
--- a/client.go
+++ b/client.go
@@ -266,8 +266,17 @@ func (client *Client) RecoverWithContext(
 	return nil
 }
 
-// Flush notifies when all the buffered events have been sent by returning `true`
-// or `false` if timeout was reached. It calls `Flush` method of the configured `Transport`.
+// Flush waits until the underlying Transport sends any buffered events to the
+// Sentry server, blocking for at most the given timeout. It returns false if
+// the timeout was reached. In that case, some events may not have been sent.
+//
+// Flush should be called before terminating the program to avoid
+// unintentionally dropping events.
+//
+// Do not call Flush indiscriminately after every call to CaptureEvent,
+// CaptureException or CaptureMessage. Instead, to have the SDK send events over
+// the network synchronously, configure it to use the HTTPSyncTransport in the
+// call to Init.
 func (client *Client) Flush(timeout time.Duration) bool {
 	return client.Transport.Flush(timeout)
 }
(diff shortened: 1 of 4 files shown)


rhcarvalho:
There is one nuance here that is not addressed:

- In an ideal program, there needs to be one call to `Flush` per transport upon termination.
- All the multiple `Flush` methods eventually call `Transport.Flush`. `Transport` is the actual thing that needs flushing.
- If the program has many clients/transports, calling `sentry.Flush` only flushes the top-level client/transport pair in the internal stack of the current hub.

As a user, one might feel overwhelmed with so many `Flush` methods and understanding when what needs to be called.

@kamilogorek @HazAT @bruno-garcia do you have any opinions or advice on how to deal with that complexity?

BTW should `sentry.Flush` and `Hub.Flush` flush all clients on its stack?

bruno-garcia:
IMHO User only knows about `Sentry.flush` and worst case `SentryClient.flush` if they are working with the client directly (outside the SDK.init/Hub). No one is flushing a transport directly. To my knowledge transport is an implementation. Sure its flush leaks out all the way to the static API but we optimize for simplicity.

I agree it's no ideal, the design. We've discussed this but no one put the effort to propose a solution.

Also, I rather not focus on Flush. For the longest time I didn't want to add Flush at all. Only when serverless came along I ran out of arguments against it.

rhcarvalho:
> IMHO User only knows about `Sentry.flush` and worst case `SentryClient.flush` if they are working with the client directly (outside the SDK.init/Hub)

I wish this was simpler too.

If there is only ever one client, then `sentry.Flush` does the job.
If new clients are created, then more `Flush` calls need to be sprinkled around for correctness. We need a clear example for this usage.

rhcarvalho:
> Also, I rather not focus on Flush. For the longest time I didn't want to add Flush at all. Only when serverless came along I ran out of arguments against it.

Unlike C# that has `Disposable` and other languages that have hooks to call `Flush` under the hood, in Go there are no magical hooks. Programs must call `Flush` explicitly.

What we're trying to clarify here is that the typical use involves a single call when the program terminates, instead of multiple calls, one after every `Capture*` call.

With this and our recent face to face conversations, and having in mind your influence to remind me that ["when in doubt, leave it out"](https://www.youtube.com/watch?v=aAb7hSCtvGw), I decided to postpone and deprioritize adding `sentry.Close` (#111).

PR #159, opened 2020-02-03
Update release steps

rhcarvalho:
- Show how to prepare the changelog (at least how to get started)
- Remove bad flag to `craft`
- Update markdown block language to highlight contents as shell session
- Explicitly configure `craft` to check commit status using Travis

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,5 +49,5 @@ $ golangci-lint run
 
     ```console
     $ craft prepare X.X.X
-    $ craft publish X.X.X --skip-status-check
+    $ craft publish X.X.X
     ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,11 +35,19 @@ $ golangci-lint run
 
 ## Release
 
-1. Update changelog with new version in `vX.X.X` format title and list of changes
-2. Commit with `misc: vX.X.X changelog` commit message and push to `master`
-3. Let `craft` do the rest
+1. Update `CHANGELOG.md` with new version in `vX.X.X` format title and list of changes.
 
-```console
-$ craft prepare X.X.X
-$ craft publish X.X.X --skip-status-check
-```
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -4,42 +4,50 @@ The public-facing channels for support and development of Sentry SDKs can be fou
...
+1. Update `CHANGELOG.md` with new version in `vX.X.X` format title and list of changes.
 
-```bash
-$ craft prepare X.X.X
-$ craft publish X.X.X --skip-status-check
-```
+    The command below can be used to get a list of changes since the last tag, with the format used in `CHANGELOG.md`:
+
+    ```console
+    $ git log --no-merges --format=%s $(git describe --abbrev=0).. | sed 's/^/- /'
```
rhcarvalho: @kamilogorek did you have a better way to accomplish this step?

kamilogorek:
```console
git log --format=%s $(git describe --tags --abbrev=0)..HEAD | sed '$d' | sed 's/^/- /'
```

This will pull everything since last tag (which is created automatically by craft), minus last commit, which is "Merged release X.X.X" commit.

rhcarvalho: Thanks! If you don't mind I'll keep the current version because:

- `sed '$d'` deletes the last line, which is typically `Merge release/X.X.X into master`. Using `--no-merges` achieves the same result, plus would remove any other merge commits if there was any.
- `ref..` is equivalent to `ref..HEAD`
- `craft` uses annotated tags for releases, meaning we don't need the `--tags` argument (it brings in other tags, maybe local-only stuff).

What I didn't mention here and what we both do is to actually filter the output of that command to what is relevant to mention in the CHANGELOG, sometimes with a few edits. That is a subjective process.

BTW thanks for having this nicely documented in the repo, it allowed me to make releases without breaking a sweat! ❤️

kamilogorek: Sounds good! And glad it actually helped someone :)

PR #160, opened 2020-02-04
handling "err.Cause() == nil" case

MikeKonan:
"err.Cause()" is nillable and leads to panic while using "err.Cause().Error()"

This pr adds handling for this nil case.


sodabrew:
@rhcarvalho Would you tag this as sentry-go 0.5.1?

rhcarvalho:
@sodabrew done https://github.com/getsentry/sentry-go/releases/tag/v0.5.1

```
go get github.com/getsentry/sentry-go@v0.5.1
```

Issue #148, opened 2020-01-27
NewClient mutates global Logger


rhcarvalho:
All `Client`s (and the rest of the SDK) share a single global `Logger`:

https://github.com/getsentry/sentry-go/blob/763dc7f49323f467d4c62ec15804723ec5350b37/client.go#L20

That means that calling `NewClient` mutates global state in surprising ways:

https://github.com/getsentry/sentry-go/blob/763dc7f49323f467d4c62ec15804723ec5350b37/client.go#L105-L113

If a program has multiple clients, setting the Debug option affects **all other clients**:

```go
go func() {
	c1, err := NewClient(ClientOptions{
		Debug: true,
	})
	if err != nil {
		log.Fatal(err)
	}
	c1.CaptureMessage("message from c1")
}()
go func() {
	c2, err := NewClient(ClientOptions{
		Debug: false,
	})
	if err != nil {
		log.Fatal(err)
	}
	c2.CaptureMessage("message from c2")
}()
```

The state of Debug logging for `c2` is undefined, might be on, might be off. No way to turn it off unless messing with the global `Logger` value.

---

**Updates:**
- 2020-03-14:
  - Changed title to remove suggestion of a specific solution.
  - While `NewClient` mutates a global logger and that's a problem in itself, the whole SDK depends on the global `Logger`, including `Hub`, `Scope`, `Client`, `Transport` implementations and integrations. We need to rethink logging with all of those in mind ⇒ #182.

YashishDua:
Checking in if this needs more discussion or ready to get into the implementation state?

rhcarvalho:
@YashishDua I think the goal here is clear. Thanks for chiming in!

YashishDua:
Perfect! Picking this up then.

PR #163, opened 2020-02-17
Attached new logger instance to client

YashishDua:
New logger instance attached to client creation. Tests are tweaked to accommodate new changes but no new tests are written for error checks. I am not sure whether we want to do it or not.

FYI: I have still kept Global logger to be used directly, however, the mutation would still exist with this.

### DIFF
--- a/client.go
+++ b/client.go
@@ -15,7 +15,7 @@ import (
 	"time"
 )
 
-// Logger is in instance of log.Logger and can be used directly to log anything manually
+// Logger is an instance of log.Logger and can be used directly to log using Sentry.Logger
 var Logger = log.New(ioutil.Discard, "[Sentry] ", log.LstdFlags) //nolint: gochecknoglobals
 
 type EventProcessor func(event *Event, hint *EventHint) (*Event, error)
--- a/client.go
+++ b/client.go
@@ -15,14 +15,13 @@ import (
 	"time"
 )
 
-// Logger is an instance of log.Logger that is use to provide debug information about running Sentry Client
-// can be enabled by either using `Logger.SetOutput` directly or with `Debug` client option
+// Logger is in instance of log.Logger and can be used directly to log anything manually
 var Logger = log.New(ioutil.Discard, "[Sentry] ", log.LstdFlags) //nolint: gochecknoglobals
 
-type EventProcessor func(event *Event, hint *EventHint) *Event
+type EventProcessor func(event *Event, hint *EventHint) (*Event, error)
 
 type EventModifier interface {
-	ApplyToEvent(event *Event, hint *EventHint) *Event
+	ApplyToEvent(event *Event, hint *EventHint) (*Event, error)
 }
 
 var globalEventProcessors []EventProcessor //nolint: gochecknoglobals
@@ -100,15 +99,22 @@ type Client struct {
 	eventProcessors []EventProcessor
 	integrations    []Integration
 	Transport       Transport
+	// Logger is an instance of log.Logger that is use to provide debug information about running Sentry Client
+	// can be enabled with `Debug` client option
+	Logger *log.Logger
 }
 
 // NewClient creates and returns an instance of `Client` configured using `ClientOptions`.
 func NewClient(options ClientOptions) (*Client, error) {
+	var logger = log.New(ioutil.Discard, "[Sentry] ", log.LstdFlags)
 	if options.Debug {
 		debugWriter := options.DebugWriter
 		if debugWriter == nil {
 			debugWriter = os.Stdout
 		}
+		// Client Logger
+		logger.SetOutput(debugWriter)
+		// Global Logger (This will get mutated every time a new client is created)
 		Logger.SetOutput(debugWriter)
 	}
 
@@ -136,6 +142,7 @@ func NewClient(options ClientOptions) (*Client, error) {
 	client := Client{
 		options: options,
 		dsn:     dsn,
+		Logger:  logger,
 	}
 
 	client.setupTransport()
@@ -149,9 +156,9 @@ func (client *Client) setupTransport() {
 
 	if transport == nil {
 		if client.options.Dsn == "" {
-			transport = new(noopTransport)
+			transport = NewNoopTransport(client.Logger)
 		} else {
-			transport = NewHTTPTransport()
+			transport = NewHTTPTransport(client.Logger)
 		}
 	}
 
@@ -173,12 +180,12 @@ func (client *Client) setupIntegrations() {
 
 	for _, integration := range integrations {
 		if client.integrationAlreadyInstalled(integration.Name()) {
-			Logger.Printf("Integration %s is already installed\n", integration.Name())
+			client.Logger.Printf("Integration %s is already installed\n", integration.Name())
 			continue
 		}
 		client.integrations = append(client.integrations, integration)
 		integration.SetupOnce(client)
-		Logger.Printf("Integration installed: %s\n", integration.Name())
+		client.Logger.Printf("Integration installed: %s\n", integration.Name())
 	}
 }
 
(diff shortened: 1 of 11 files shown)


rhcarvalho:
@YashishDua thank you for the PR!

> FYI: I have still kept Global logger to be used directly, however, the mutation would still exist with this.

:+1: indeed it would be a smoother transition if we can keep that global logger around for a while, deprecate it and remove it as part of cleaning up the exported API.

rhcarvalho:
@YashishDua the changes in the current state break the example code: https://travis-ci.com/getsentry/sentry-go/jobs/288114402

There are API changes, we need to be careful and diligent with those.

transport.go
```diff
@@ -64,10 +66,10 @@ func retryAfter(now time.Time, r *http.Response) time.Duration {
 	return defaultRetryAfter
 }
 
-func getRequestBodyFromEvent(event *Event) []byte {
+func getRequestBodyFromEvent(event *Event) ([]byte, error) {
```
rhcarvalho: Here looks like logging was used where we could return errors.

However, most of the time when we return an error as the last argument it means the other return values are no good. Only few functions have "partial" success and that's harder to reason about.

Since this is not exported, we have more freedom to change. I'd suggest passing in the logger as an explicit argument.

```suggestion
func getRequestBodyFromEvent(event *Event, logger *log.Logger) []byte {
```

client.go
```diff
@@ -15,14 +15,13 @@ import (
...
-// can be enabled by either using `Logger.SetOutput` directly or with `Debug` client option
+// Logger is an instance of log.Logger and can be used directly to log using Sentry.Logger
 var Logger = log.New(ioutil.Discard, "[Sentry] ", log.LstdFlags) //nolint: gochecknoglobals
 
-type EventProcessor func(event *Event, hint *EventHint) *Event
+type EventProcessor func(event *Event, hint *EventHint) (*Event, error)
 
 type EventModifier interface {
-	ApplyToEvent(event *Event, hint *EventHint) *Event
+	ApplyToEvent(event *Event, hint *EventHint) (*Event, error)
```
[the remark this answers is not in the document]
YashishDua: Yes, initially I thought passing logger might not be the right choice, and at places, passing could be at nested. So, I was trying to avoid that.

YashishDua:
@rhcarvalho You can have a look now!

rhcarvalho:
@YashishDua sorry I didn't have capacity to review this in detail yet. Getting rid of the global logger is more involved than I anticipated.
I may need a few more days to get back to this.

YashishDua:
@rhcarvalho Any update on this?

rhcarvalho:
Hi @YashishDua! I've thought about this a few times over the last weeks and now had another look.

I ended up reading (again) https://dave.cheney.net/2017/01/23/the-package-level-logger-anti-pattern and logging-related discussions in other projects like Kubernetes and posts in golang-nuts.

Your work and this PR makes it clear that several parts of the public API would require change. I'm now reluctant to think `*log.Logger` is the right dependency to impose on users of the SDK.

People should be free to use whatever logging framework they want (and possibly already use in their `main` package), be it [`log`](https://golang.org/pkg/log/), [`go.uber.org/zap`](https://github.com/uber-go/zap), [`github.com/golang/glog`](https://github.com/golang/glog), [`k8s.io/klog`](https://github.com/kubernetes/klog), [`github.com/sirupsen/logrus`](https://www.github.com/sirupsen/logrus), ..., the list is really long ⇒ [awesome-go#logging](https://github.com/avelino/awesome-go#logging).

I do not want to disappoint you, but I would like to consider the API change more carefully. I hope we can agree that there is a cost for users when we make breaking changes, therefore, let's take a step back and trace a path forward.

Right now consumers of `sentry-go` can enable/disable (debug) logging as they please, and they can configure where the output is sent. It is not ideal, and has problems as I described in #148. As far as I know nobody has stepped up to complain about that yet, except for myself who filed the issue.

Sorry for the long message, just wanted to put down my thoughts and hear what your ideas are. Can we discuss the requirements and design first before we invest more time on the solution? I apologize for not having discussed it in more detail before in #148. #148 was a mental note, something I had planned to work on, but had not investigated the consequences in detail. Your work made things clear, thank you! I appreciate your help.

rhcarvalho:
@YashishDua I have an idea to move this forward.

Looking back at #148, I realized we can split the problem into two parts. The first one is #148 itself, which concerns only `Client` and the problem of calling `NewClient`.

For all the rest I discussed in my previous comment, the deeper changes required to deprecate/remove the global `Logger`, I've opened #182.

I believe we can fix #148 in this PR without changes to the public API.

client.go
```diff
@@ -136,6 +142,7 @@ func NewClient(options ClientOptions) (*Client, error) {
 	client := Client{
 		options: options,
 		dsn:     dsn,
+		Logger:  logger,
```
rhcarvalho:
```suggestion
		logger:  logger,
```

client.go
```diff
@@ -100,15 +99,22 @@ type Client struct {
...
+	var logger = log.New(ioutil.Discard, "[Sentry] ", log.LstdFlags)
 	if options.Debug {
 		debugWriter := options.DebugWriter
 		if debugWriter == nil {
 			debugWriter = os.Stdout
 		}
+		// Client Logger
+		logger.SetOutput(debugWriter)
+		// Global Logger (This will get mutated every time a new client is created)
 		Logger.SetOutput(debugWriter)
```
rhcarvalho: Suggestion: move this to `Init`, so that `Client` becomes independent of the global logger.

In `NewClient` / `Client`, we only use a local, unexpected, logger (`Client.logger`).

client.go
```diff
@@ -149,9 +156,9 @@ func (client *Client) setupTransport() {
 
 	if transport == nil {
 		if client.options.Dsn == "" {
-			transport = new(noopTransport)
+			transport = &noopTransport{Logger: client.Logger}
```
rhcarvalho: Let the rest of the SDK use the global `Logger` for now, we solve that later. Tracked by #182.

client.go
```diff
@@ -173,12 +180,12 @@ func (client *Client) setupIntegrations() {
 
 	for _, integration := range integrations {
 		if client.integrationAlreadyInstalled(integration.Name()) {
-			Logger.Printf("Integration %s is already installed\n", integration.Name())
+			client.Logger.Printf("Integration %s is already installed\n", integration.Name())
```
rhcarvalho:
```suggestion
			client.logger.Printf("Integration %s is already installed\n", integration.Name())
```

client.go
```diff
@@ -173,12 +180,12 @@ func (client *Client) setupIntegrations() {
...
 			continue
 		}
 		client.integrations = append(client.integrations, integration)
 		integration.SetupOnce(client)
-		Logger.Printf("Integration installed: %s\n", integration.Name())
+		client.Logger.Printf("Integration installed: %s\n", integration.Name())
```
rhcarvalho:
```suggestion
		client.logger.Printf("Integration installed: %s\n", integration.Name())
```

client.go
```diff
@@ -339,7 +346,7 @@ func (client *Client) processEvent(event *Event, hint *EventHint, scope EventMod
 	if options.SampleRate != 0.0 {
 		randomFloat := rand.New(rand.NewSource(time.Now().UnixNano())).Float64()
 		if randomFloat > options.SampleRate {
-			Logger.Println("Event dropped due to SampleRate hit.")
+			client.Logger.Println("Event dropped due to SampleRate hit.")
```
rhcarvalho:
```suggestion
			client.logger.Println("Event dropped due to SampleRate hit.")
```

client.go
```diff
@@ -354,7 +361,7 @@ func (client *Client) processEvent(event *Event, hint *EventHint, scope EventMod
 			h = hint
 		}
 		if event = options.BeforeSend(event, h); event == nil {
-			Logger.Println("Event dropped due to BeforeSend callback.")
+			client.Logger.Println("Event dropped due to BeforeSend callback.")
```
rhcarvalho:
```suggestion
			client.logger.Println("Event dropped due to BeforeSend callback.")
```

Issue #97, opened 2019-11-28
Add tests to integrations


rhcarvalho:
Currently the packages implementing integrations (http, gin, iris, martini, etc) lack tests.

rhcarvalho:
When working on #105 I noticed that modules integration is also missing tests.
The implementation is in `integrations.go`

PR #167, opened 2020-02-24
Add tests for net/http and fasthttp integrations

rhcarvalho:
Tests should allow us to make changes to `sentry-go` with some confidence that integrations still work.

Targeting `net/http` and `fasthttp` because other integrations (`echo`, `gin`, `iris`, `martini`, `negroni`) are similar to `net/http`, so we cover most of the surface area.

The tests cover 3 scenarios:
- Capturing panics
- Capturing message with request body
- Capturing message without request body

The tests in `sentryhttp` and `sentryfasthttp` are pretty similar, however need to account for differences in `net/http` and `fasthttp`.

Updates #97.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,8 @@ jobs:
         export GOPATH= &&
         go env GOPATH
       script: >-
-        go test &&
-        go test -race
+        go test ./... &&
+        go test ./... -race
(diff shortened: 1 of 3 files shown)


kamilogorek:
Awesome!

PR #170, opened 2020-02-26
ci: Test against Go 1.14, drop 1.11

rhcarvalho:
Go 1.14 has been released, so update the versions we test against.

Testing of Module mode outside of GOPATH can use the latest stable
release, thus not requiring a config update with each release of Go 1.x.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: go
 
 go:
-  - 1.11.x
   - 1.12.x
   - 1.13.x
+  - 1.14.x
   - master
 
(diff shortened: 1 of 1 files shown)


rhcarvalho:
Note that we're not bumping the language version in `go.mod`. We may defer that until some later time.

With my current understanding, we should keep that to the lowest language version we support (i.e, if we change it we should bump from 1.11 to 1.12, not 1.14).

Since we don't require anything from 1.12+ right now, it seems okay to leave that alone for now.

PR #183, opened 2020-03-14
Add Exception.ThreadID field

Pr0Ger:
Doc says that crashed thread should not have stacktrace, but instead, the `thread_id` attribute should be set on the exception https://docs.sentry.io/development/sdk-dev/event-payloads/exception/

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -112,6 +112,7 @@ type Exception struct {
 	Type          string      `json:"type,omitempty"`
 	Value         string      `json:"value,omitempty"`
 	Module        string      `json:"module,omitempty"`
+	ThreadID      string      `json:"thread_id,omitempty"`
 	Stacktrace    *Stacktrace `json:"stacktrace,omitempty"`
 	RawStacktrace *Stacktrace `json:"raw_stacktrace,omitempty"`
 }


rhcarvalho:
Hi @Pr0Ger! There is no "Goroutine ID" available to user code in Go. The docs describe a general protocol we have to adapt to the reality of different languages/platforms.

(In fact, we do have a field `Thread.ID` that is never set..., and never sent)

Pr0Ger:
Actually [there is](https://github.com/golang/net/blob/244492dfa37ae2ce87222fd06250a03160745faa/http2/gotrack.go#L51-L68), although there is no friendly API for it.

The reason why I thought to add this field is to have stacktraces sent in a consistent way when there is an exception and when it's not

(Yeah, I saw some fields which are not only not used but even is not documented like `Exception.RawStacktrace`)

rhcarvalho:
> Actually [there is](https://github.com/golang/net/blob/244492dfa37ae2ce87222fd06250a03160745faa/http2/gotrack.go#L51-L68), although there is no friendly API for it.

Note that this particular file is part of the HTTP2 implementation and starts with the comment:

https://github.com/golang/net/blob/244492dfa37ae2ce87222fd06250a03160745faa/http2/gotrack.go#L5-L6
```go
// Defensive debug-only utility to track that functions run on the
// goroutine that they're supposed to.
```

AFAIK Goroutine IDs are never exported, and that's by the design of the language as explained in the FAQ: https://golang.org/doc/faq#no_goroutine_id


> The reason why I thought to add this field is to have stacktraces sent in a consistent way when there is an exception and when it's not

Could you elaborate? Do you plan on adding more code to this PR?

Pr0Ger:
Sure. I'm writing a library to integrate zap logger with sentry. Debug messages are stored as breadcrumbs, messages with error level and higher are sent as events, and if a message has an error field it will be added as an exception to the event.

So, there are the following options:
error message: create an event with stacktrace in in `threads`
panic/fatal message: add crashed: true to thread
and if the message has an attached error field it will be added to `exception`

Here is a thing: if I send both thread (without stacktrace) and exception (with stacktrace) it will not be shown in UI (stacktrace section will contain info about thread crashed or not and "No or unknown stacktrace"/"Thread Crashed" instead of stacktrace). Sending stacktrace in both thread and exception works but seems unnecessary. But if I send exception with `thread_id` UI properly show thread with stacktrace, exception info and crashed flag. Almost the same can be achieved by sending only exception (without thread section) although there will no place to pass crashed flag

PR #185, opened 2020-03-23
Add support for chained errors

### LINKED ISSUE #137: Support for stack frames from xerrors.
It would be incredibly helpful if sentry would extract all included stack frames out of errors wrapped using xerrors.Errorf (or the go 1.13 errors package)

### LINKED ISSUE #176: Make eventFromException handle recursively wrapped errors
Currently, `client.eventFromException` only calls the `Cause` method on the exception once.

Using `github.com/pkg/errors`, if we use the `errors.Wrap` function to wrap our error (let's say of type `TestError`), this gives us the following chain: 
`errors.WithStack` -> `errors.WithMessage` -> `TestError`
and since the Cause method returns the direct cause, not the underlying one, we end up with type `errors.WithMessage` being reported to Sentry. (Instead of `TestError`.)

I believe the underlying cause should be reported to sentry, i.e. `eventFromException` should determine cause in the same way [github.com/pingcap/errors](https://github.com/pingcap/errors/blob/master/errors.go#L292-L298) and [github.com/pkg/errors](https://github.com/pkg/errors/blob/master/errors.go#L275-L288) do (both call the Cause method repeatedly, the first recursively, the second iteratively). 

The proposed change of [this part](https://github.com/getsentry/sentry-go/blob/v0.5.1/client.go#L314-L320):
```go
for exception != nil {
	if ex, ok := exception.(interface{ Cause() error }); ok {
		exception = ex.Cause()
	} else {
		break
	}
}
cause := exception
```

Let me know what you think and whether I can open a PR to fix this :)

rhcarvalho:
Instead of reporting only the top parent of an error chain, report all errors that are part of the chain.

Supports chains of errors using the `Cause` method (used by popular error packages predating Go 1.13).

Support for errors using the `Unwrap` method (introduced in Go 1.13) or a combination of `Unwrap` and `Cause` will come briefly in a follow up.

Fixes #176.

---

**Example report of chained errors**

image

### DIFF
--- a/client.go
+++ b/client.go
@@ -15,6 +15,15 @@ import (
 	"time"
 )
 
+// maxErrorDepth is the maximum number of errors reported in a chain of errors.
+// This protects the SDK from an arbitrarily long chain of wrapped errors.
+//
+// An additional consideration is that arguably reporting a long chain of errors
+// is of little use when debugging production errors with Sentry. The Sentry UI
+// is not optimized for long chains either. The top-level error together with a
+// stack trace is often the most useful information.
+const maxErrorDepth = 10
+
 // usageError is used to report to Sentry an SDK usage error.
 //
 // It is not exported because it is never returned by any function or method in
@@ -306,32 +315,46 @@ func (client *Client) eventFromMessage(message string, level Level) *Event {
 }
 
 func (client *Client) eventFromException(exception error, level Level) *Event {
-	if exception == nil {
-		exception = usageError{fmt.Errorf("%s called with nil error", callerFunctionName())}
+	err := exception
+	if err == nil {
+		err = usageError{fmt.Errorf("%s called with nil error", callerFunctionName())}
 	}
 
-	stacktrace := ExtractStacktrace(exception)
+	event := &Event{Level: level}
 
-	if stacktrace == nil {
-		stacktrace = NewStacktrace()
-	}
-
-	cause := exception
-	// Handle wrapped errors for github.com/pingcap/errors and github.com/pkg/errors
-	if ex, ok := exception.(interface{ Cause() error }); ok {
-		if c := ex.Cause(); c != nil {
-			cause = c
+	for i := 0; i < maxErrorDepth && err != nil; i++ {
+		event.Exception = append(event.Exception, Exception{
+			Value:      err.Error(),
+			Type:       reflect.TypeOf(err).String(),
+			Stacktrace: ExtractStacktrace(err),
+		})
+		if previous, ok := err.(interface{ Cause() error }); ok {
+			err = previous.Cause()
+		} else {
+			err = nil
 		}
 	}
 
-	event := NewEvent()
-	event.Level = level
-	event.Exception = []Exception{{
-		Value:      cause.Error(),
-		Type:       reflect.TypeOf(cause).String(),
-		Stacktrace: stacktrace,
-	}}
+	// Add a trace of the current stack to the most recent error in a chain if
+	// it doesn't have a stack trace yet.
+	// We only add to the most recent error to avoid duplication and because the
+	// current stack is most likely unrelated to errors deeper in the chain.
+	if event.Exception[0].Stacktrace == nil {
+		event.Exception[0].Stacktrace = NewStacktrace()
+	}
+
+	// event.Exception should be sorted such that the most recent error is last.
+	reverse(event.Exception)
+
 	return event
+}
+
+// reverse reverses the slice a in place.
+func reverse(a []Exception) {
+	for i := len(a)/2 - 1; i >= 0; i-- {
+		opp := len(a) - 1 - i
+		a[i], a[opp] = a[opp], a[i]
+	}
 }
 
 func (client *Client) processEvent(event *Event, hint *EventHint, scope EventModifier) *EventID {
--- a/client_test.go
+++ b/client_test.go
@@ -3,9 +3,9 @@ package sentry
 import (
 	"errors"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
 	pkgErrors "github.com/pkg/errors"
 )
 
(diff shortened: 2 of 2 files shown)


client_test.go
```diff
@@ -59,54 +59,12 @@ func TestCaptureMessageShouldSucceedWithoutNilScope(t *testing.T) {
 	assertEqual(t, transport.lastEvent.Message, "foo")
 }
 
-func TestCaptureExceptionShouldSendEventWithProvidedError(t *testing.T) {
-	client, scope, transport := setupClientTest()
-	client.CaptureException(errors.New("custom error"), nil, scope)
-	assertEqual(t, transport.lastEvent.Exception[0].Type, "*errors.errorString")
-	assertEqual(t, transport.lastEvent.Exception[0].Value, "custom error")
-}
```
rhcarvalho, line 67: This got replaced by the test `SimpleError` below.

client_test.go
```diff
@@ -59,54 +59,12 @@ func TestCaptureMessageShouldSucceedWithoutNilScope(t *testing.T) {
...
-		cmpopts.IgnoreTypes(&Stacktrace{}),
-		cmpopts.EquateEmpty(),
-	}
-	if diff := cmp.Diff(want, transport.lastEvent, opts); diff != "" {
-		t.Errorf("event mismatch (-want +got):\n%s", diff)
-	}
-	if transport.lastEvent.Exception[0].Stacktrace == nil {
-		t.Errorf("missing stacktrace")
-	}
-}
```
rhcarvalho, line 93: This got replaced by the test `NilError` below.

client_test.go
```diff
@@ -59,54 +59,12 @@ func TestCaptureMessageShouldSucceedWithoutNilScope(t *testing.T) {
...
 }
 
-func TestCaptureExceptionShouldExtractCorrectTypeAndValueForWrappedErrors(t *testing.T) {
-	client, scope, transport := setupClientTest()
-	cause := &customErr{}
-	err := pkgErrors.WithStack(cause)
-	client.CaptureException(err, nil, scope)
-	assertEqual(t, transport.lastEvent.Exception[0].Type, "*sentry.customErr")
-	assertEqual(t, transport.lastEvent.Exception[0].Value, "wat")
-}
```
rhcarvalho, line 108: This got replaced by the test `MostRecentErrorHasStack` below.

client_test.go
```diff
@@ -117,20 +75,161 @@ func (e *customErrWithCause) Cause() error {
 	return e.cause
 }
 
-func TestCaptureExceptionShouldNotUseCauseIfCauseIsNil(t *testing.T) {
-	client, scope, transport := setupClientTest()
-	err := &customErrWithCause{cause: nil}
-	client.CaptureException(err, nil, scope)
-	assertEqual(t, transport.lastEvent.Exception[0].Type, "*sentry.customErrWithCause")
-	assertEqual(t, transport.lastEvent.Exception[0].Value, "err")
```
rhcarvalho, line 125: This got replaced with the test `ChainWithNilCause` below.

client_test.go
```diff
@@ -117,20 +75,161 @@ func (e *customErrWithCause) Cause() error {
...
+	name  string
+	tests []captureExceptionTest
 }
 
-func TestCaptureExceptionShouldUseCauseIfCauseIsNotNil(t *testing.T) {
-	client, scope, transport := setupClientTest()
-	err := &customErrWithCause{cause: &customErr{}}
-	client.CaptureException(err, nil, scope)
-	assertEqual(t, transport.lastEvent.Exception[0].Type, "*sentry.customErr")
-	assertEqual(t, transport.lastEvent.Exception[0].Value, "wat")
```
rhcarvalho, line 133: This got replaced with the test `ChainWithoutStacktrace` below.

client_test.go
```diff
@@ -117,20 +75,161 @@ func (e *customErrWithCause) Cause() error {
...
+				},
+				// TODO: perhaps the list of packages is incomplete or there
+				// should not be any package at all.
+			},
+		},
+	}
+	got := transport.lastEvent
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Errorf("Event mismatch (-want +got):\n%s", diff)
+	}
```
rhcarvalho, line 232: This is notably not a very good test, but a starting point to add tests around code that fills in event fields.

It is so that `TestCaptureException` can focus on the list of "exceptions" (errors) instead of dealing with stuff like SDK info, Platform, etc.

client_test.go
```diff
@@ -117,20 +75,161 @@ func (e *customErrWithCause) Cause() error {
...
+			Packages: []SdkPackage{
+				{
+					// FIXME: name format doesn't follow spec in
+					// https://docs.sentry.io/development/sdk-dev/event-payloads/sdk/
+					Name:    "sentry-go",
+					Version: Version,
+				},
+				// TODO: perhaps the list of packages is incomplete or there
+				// should not be any package at all.
+			},
```
rhcarvalho, line 226: This is something we need to review in the future.

client.go
```diff
@@ -306,34 +315,48 @@ func (client *Client) eventFromMessage(message string, level Level) *Event {
...
-		Type:       reflect.TypeOf(cause).String(),
-		Stacktrace: stacktrace,
-	}}
+	// event.Exception should be sorted such that the most recent error is last.
+	reverse(event.Exception)
+
 	return event
 }
 
+// reverse reverses the slice a in place.
```
bruno-garcia, line 352:
```suggestion
// reverse reverses the slice in place.
```

rhcarvalho: Not a typo, `a` is the name of the argument :)

bruno-garcia: I have 0 to contribute.
Maybe useless questions: Is there really no built-in generic array reversal function in Go? Oh, no generic?
Also, what about renaming `a` -> exceptions` or, `e`?
And finally:

```suggestion
// reverse reverses the slice 'a' in place.
```

:trollface:

client.go
```diff
@@ -306,34 +315,48 @@ func (client *Client) eventFromMessage(message string, level Level) *Event {
...
+func reverse(a []Exception) {
```
kamilogorek, line 353: Ref: https://github.com/golang/go/wiki/SliceTricks#reversing :D

client_test.go
```diff
@@ -117,20 +75,161 @@ func (e *customErrWithCause) Cause() error {
...
+		},
+		{
+			name:  "ErrorChain",
+			tests: errorChainTests,
+		},
+	}
+
+	for _, grp := range tests {
+		for _, tt := range grp.tests {
+			tt := tt
```
kamilogorek, line 175: Is it for closure scoping purpose?

rhcarvalho: Yes. Because `tt` is in the scope of the subtest, having a new variable for each test prevents running into https://github.com/golang/go/wiki/CommonMistakes and allows us to run tests in parallel if we wanted to.

The observed behavior in the current code base today would be the same without this line, but I like to have it to help whomever will maintain this later.

It is easy to imagine some contribution to make tests run in parallel (not a priority today because they are pretty fast anyway), the tests would be green and we would not realize that we're only running the last test case over and over...

PR #189, opened 2020-04-03
fix: Report usage error including stack trace

rhcarvalho:
Prior to this, calling `CaptureException`, `Recover` or `RecoverWithContext` with a `nil` error would send a message event to Sentry without a stack trace.

Looking at the Sentry issues list, and issue detail, all you see is "Called CaptureException with nil value" without a clue of what caused it.

This change turns that message into a `sentry.usageError` (new type) with a stack trace, making it easier to understand what is wrong and fix user code.

## BEFORE

No stacktrace nor any clue where the incorrect call to `CaptureException` is.

image
image

## AFTER

Usage error with description and file name and line number information, plus source code context when available.

image
image

Without source context:

image

### DIFF
--- a/client.go
+++ b/client.go
@@ -15,6 +15,14 @@ import (
 	"time"
 )
 
+// usageError is used to report to Sentry an SDK usage error.
+//
+// It is not exported because it is never returned by any function or method in
+// the exported API.
+type usageError struct {
+	error
+}
+
 // Logger is an instance of log.Logger that is use to provide debug information about running Sentry Client
 // can be enabled by either using `Logger.SetOutput` directly or with `Debug` client option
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -299,10 +307,7 @@ func (client *Client) eventFromMessage(message string, level Level) *Event {
 
 func (client *Client) eventFromException(exception error, level Level) *Event {
 	if exception == nil {
-		event := NewEvent()
-		event.Level = level
-		event.Message = fmt.Sprintf("Called %s with nil value", callerFunctionName())
-		return event
+		exception = usageError{fmt.Errorf("%s called with nil error", callerFunctionName())}
```
rhcarvalho, line 310: Moved the function name to the very beginning because it is the most relevant information.

PR #191, opened 2020-04-06
fix: Use time.Time for timestamps

### LINKED ISSUE #190: Sending timestamps including milliseconds
> I didn't read through this thread in detail, but I came here searching for an issue where my millisecond and nanosecond breadcrumb timestamps (formated as a Unix timestamp integer with millis or nanos added) are being rejected by the Sentry API (regular Unix timestamps in seconds work as expected).
> 
> After verifying that my timestamps are correct, my mind drifted towards clock skew, which was how I found this issue.
> 
> Here's what the API injected into the raw JSON:
> 
> ```json
>   "_meta": {
>     "breadcrumbs": {
>       "values": {
>         "0": {
>           "timestamp": {
>             "": {
>               "val": 1585983484747,
>               "err": [
>                 [
>                   "invalid_data",
>                   {
>                     "reason": "timestamp out of range"
>                   }
>                 ]
>               ]
>             }
>           }
>         }
>       }
>     }
>   }
> ```
> 
> If this isn't related, I'll file a separate issue (assuming I'm not doing anything incorrectly).
> 
> This is using the Go SDK.

_Originally posted by @JeanMertz in https://github.com/getsentry/relay/issues/534#issuecomment-608990645_

rhcarvalho:
This gives us more flexibility to represent and manipulate time values,
as well as allow us to report values with sub-second resolution.

Fixes #190.

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -2,10 +2,12 @@ package sentry
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net"
 	"net/http"
 	"strings"
+	"time"
 )
 
 // Protocol Docs (kinda)
@@ -46,8 +48,30 @@ type Breadcrumb struct {
 	Data      map[string]interface{} `json:"data,omitempty"`
 	Level     Level                  `json:"level,omitempty"`
 	Message   string                 `json:"message,omitempty"`
-	Timestamp int64                  `json:"timestamp,omitempty"`
+	Timestamp time.Time              `json:"timestamp"`
 	Type      string                 `json:"type,omitempty"`
+}
+
+func (b *Breadcrumb) MarshalJSON() ([]byte, error) {
+	type alias Breadcrumb
+	// encoding/json doesn't support the "omitempty" option for struct types.
+	// See https://golang.org/issues/11939.
+	// This implementation of MarshalJSON shadows the original Timestamp field
+	// forcing it to be omitted when the Timestamp is the zero value of
+	// time.Time.
+	if b.Timestamp.IsZero() {
+		return json.Marshal(&struct {
+			*alias
+			Timestamp json.RawMessage `json:"timestamp,omitempty"`
+		}{
+			alias: (*alias)(b),
+		})
+	}
+	return json.Marshal(&struct {
+		*alias
+	}{
+		alias: (*alias)(b),
+	})
 }
 
 // https://docs.sentry.io/development/sdk-dev/event-payloads/user/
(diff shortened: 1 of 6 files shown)


rhcarvalho:
While this is a breaking change to the exported API, it is one that will affect only code that manually sets and reads the timestamp. An example of that would be custom code in the `BeforeSend` callback.

The "common case" is for the timestamp to be filled in by the SDK, in which case user code should continue to compile and work normally without disruption.

interfaces.go
```diff
@@ -46,7 +47,7 @@ type Breadcrumb struct {
 	Data      map[string]interface{} `json:"data,omitempty"`
 	Level     Level                  `json:"level,omitempty"`
 	Message   string                 `json:"message,omitempty"`
-	Timestamp int64                  `json:"timestamp,omitempty"`
+	Timestamp time.Time              `json:"timestamp,omitempty"`
```
rhcarvalho, line 50: We could do this without breaking the API by adding a new field and keeping the old. Since we're still `v0.x` I'd avoid leaving cruft in the exported API.

rhcarvalho:
This is not ready yet. Needs test updates and omitting the zero time from serialized events and breadcrumbs.

marshal_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestMarshalJSON(t *testing.T) {
+	tests := []struct {
+		in  interface{}
+		out string
+	}{
+		// TODO: eliminate empty struct fields from serialization of empty event.
```
rhcarvalho, line 15: Today we send `{}` for empty structs that appear anywhere within an `Event`, and we can extend the `MarshalJSON` implementations to handle those cases later.

scope_concurrency_test.go
```diff
@@ -54,7 +54,7 @@ func touchScope(scope *sentry.Scope, x int) {
 	scope.SetLevel(sentry.LevelDebug)
 	scope.SetTransaction("foo")
 	scope.SetFingerprint([]string{"foo"})
-	scope.AddBreadcrumb(&sentry.Breadcrumb{Timestamp: 1337, Message: "foo"}, 100)
+	scope.AddBreadcrumb(&sentry.Breadcrumb{Message: "foo"}, 100)
```
rhcarvalho, line 57: Here the `Timestamp` was irrelevant, so it was removed.

interfaces.go
```diff
@@ -145,6 +169,28 @@ type Event struct {
...
+	type alias Event
+	// encoding/json doesn't support the "omitempty" option for struct types.
+	// See https://golang.org/issues/11939.
+	// This implementation of MarshalJSON shadows the original Timestamp field
+	// forcing it to be omitted when the Timestamp is the zero value of
+	// time.Time.
+	if e.Timestamp.IsZero() {
+		return json.Marshal(&struct {
+			*alias
+			Timestamp json.RawMessage `json:"timestamp,omitempty"`
```
rhcarvalho, line 182: Here we could have used any type for `Timestamp` as long as it works with `omitempty`.

I chose `json.RawMessage` (over `bool`, `[]byte`, `int`, etc) because of the semantic value it carries.
Since this is not exported we can change it later at will.

IIRC, my intention was to use `json.RawMessage` in other fields to add pre-encoded chunks of JSON.

rhcarvalho:
> Ok this is ready for a review.

No it is not... linter is unhelpful... 😠

rhcarvalho:
Ok, CI green.

scope_test.go
```diff
@@ -7,8 +7,11 @@ import (
 	"time"
 )
 
+//nolint: gochecknoglobals
```
rhcarvalho, line 10: This lint rule is not useful. While minimizing global state is relevant, sprinkling code with lint comments is questionable.

`//nolint: gochecknoglobals` is 50% of the `//nolint:` comments in the code base.

PR #192, opened 2020-04-06
fix: check for dropped event

gawen:
`scope.ApplyToEvent` may return nil, but is not checked in `prepareEvent`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -415,6 +415,9 @@ func (client *Client) prepareEvent(event *Event, hint *EventHint, scope EventMod
 
 	if scope != nil {
 		event = scope.ApplyToEvent(event, hint)
+		if event == nil {
+			return nil
+		}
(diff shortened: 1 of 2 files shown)


rhcarvalho:
Thanks @gawen, good catch! ❤️ 

Would you like to add a test case?

gawen:
@rhcarvalho sure, although I don't see a relevant test case.

rhcarvalho:
The case is when an event processor in the scope drops an event.

The default scope starts without any event processor, that's probably why the bug was hiding for a long time.

rhcarvalho:
I've rebased against current master and added a simple change to a test that minimally covers the change (the test breaks without the change from @gawen).

gawen:
@rhcarvalho thanks for merging!

Issue #156, opened 2020-01-30
Split sentry-go into multiple submodules


rhcarvalho:
The topic has popped up in a few occasions. Some don't like to see a long list of indirect dependencies in their `go.mod` files.

https://github.com/getsentry/sentry-go/issues/75#issuecomment-553648455

https://github.com/getsentry/sentry-go/issues/75#issuecomment-557699949

https://github.com/getsentry/sentry-go/issues/79

rhcarvalho:
@rur wrote in https://github.com/getsentry/sentry-go/issues/75#issuecomment-553648455:

> Overhead may or may not end up in the artifact but it will spam my project dependencies, which I care about.
> 
> Please read https://research.swtch.com/deps
> 
> ```
> +github.com/Joker/hpp v0.0.0-20180418125244-6893e659854a/go.mod h1:MzD2WMdSxvbHw5fM/OXOFily/lipJWRc9C1px0Mt0ZE=
> +github.com/Joker/jade v1.0.0/go.mod h1:efZIdO0py/LtcJRSa/j2WEklMSAw84WV0zZVMxNToB8=
> +github.com/Shopify/goreferrer v0.0.0-20181106222321-ec9c9a553398/go.mod h1:a1uqRtAwp2Xwc6WNPJEufxJ7fx3npB4UV/JOLmbu5I0=
> +github.com/ajg/form v1.5.1/go.mod h1:uL1WgH+h2mgNtvBq0339dVnzXdBETtL2LeUXaIv25UY=
> +github.com/aymerick/raymond v2.0.2+incompatible/go.mod h1:osfaiScAUVup+UC9Nfq76eWqDhXlp+4UYaA8uhTBO6g=
> +github.com/codegangsta/inject v0.0.0-20150114235600-33e0aa1cb7c0/go.mod h1:4Zcjuz89kmFXt9morQgcfYZAYZ5n8WHjt81YYWIwtTM=
> +github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
> +github.com/dgrijalva/jwt-go v3.2.0+incompatible/go.mod h1:E3ru+11k8xSBh+hMPgOLZmtrrCbhqsmaPHjLKYnJCaQ=
[...]

PR #196, opened 2020-04-08
Add integration with buffalo

### LINKED ISSUE #194: Add an Integration for Buffalo Web Framework
It would be helpful to have an integration for Buffalo similar to how there are integrations for other web frameworks.

stnguyen90:
fixes #194


rhcarvalho:
Adding this will pull in a good bunch of new deps into `go.mod`, which is unfortunate.

Would be good to advance on #156 before, so that we don't bloat users' `go.mod` when they don't need the integration.

Alternatively, I'm also tracking https://github.com/golang/go/issues/36460 which might alleviate the problem and allow us to have a single module  == easier maintenance and distribution/consumption -- single coherent version. That's planned for Go 1.15.

PR #199, opened 2020-04-14
Read module dependencies from debug.BuildInfo instead of go.mod

### LINKED ISSUE #184: List of packages include "replace" package
Among packages from `go.mod`, package list is also showing a `replace` entry, looks wrong:

image

Here's the code that needs improvement:

https://github.com/getsentry/sentry-go/blob/a0f55ee0c7108aef71a7cb94e7c82edafdb0ab83/integrations.go#L70-L104

wingyplus:
This changes getting modules from `runtime/debug` instead of getting it from `go.mod` or `vendor/modules.txt`.

Screen Shot 2563-04-15 at 00 22 05

Fixes #184

### DIFF
--- a/integrations.go
+++ b/integrations.go
@@ -1,11 +1,7 @@
 package sentry
 
 import (
-	"bufio"
-	"encoding/json"
 	"fmt"
-	"io/ioutil"
-	"os"
 	"regexp"
 	"runtime"
 	"runtime/debug"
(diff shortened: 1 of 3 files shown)


wingyplus:
From the image above, the sentry module that's mark as version `0.0.0` is replace module. Do we needs to set the replacement module?

rhcarvalho:
@wingyplus thanks for your contribution!

Could you please update `go.mod` to `go 1.12` -- `debug.ReadBuildInfo` requires `1.12` and that's the minimum version we support today anyway.

integrations.go
```diff
@@ -67,39 +63,15 @@ func getModules() (map[string]string, error) {
...
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		return nil, fmt.Errorf("cannot dependencies from debug.BuildInfo")
 	}
-
-	if scannerErr := scanner.Err(); scannerErr != nil {
-		return nil, scannerErr
+	modules := make(map[string]string)
+	for _, dep := range info.Deps {
+		modules[dep.Path] = dep.Version
```
rhcarvalho, line 73: Needs to include `info.Main` too 😉 

---

As you well noted in a comment, we also need to handle replacements. The Sentry server only takes a `map[string]string` for package/module name => version, no concept of replacements. While we could look into adding replacements server-side and in the web interface, I think we can well start with just reporting replacements within the current server constraints. 

There are replacements pointing to a module+version, and replacements pointing to absolute local path and relative local path, see https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive.

We could use a syntax similar to how `go list -m all` prints replacements:

```
module/name v0.0.0 => replacement
```

```go
map[string]string{"module/name": "v0.0.0 => replacement"}
```

What do you think?

wingyplus: I think it sound good to me for doing this.

wingyplus:
@rhcarvalho Thanks for your review. Main issue was adressed. 

> The basic test cases would be around the kinds of output debug.ReadBuildInfo() can return: no replacements, replacements with mod+version, absolute path, relative path.

Integration tests would a bit hard because debug.ReadBuildInfo inject modules at compile time. Mocking `debug.ReadBuildInfo` might be a bit easier. The illustrate that I think is:

```go
var readBuildInfo = debug.ReadBuildInfo // this way we can inject a function that returns a modules list. 

func extractModules(){
  info, err := readBuildInfo()
  // ....
}
```

Not sure what do you think?

integrations.go
wingyplus: Fixed.

integrations.go
```diff
@@ -40,125 +37,22 @@ func (mi *modulesIntegration) processor(event *Event, hint *EventHint) *Event {
...
-			if len(splits) > 2 {
-				modules[strings.TrimSpace(splits[1])] = splits[2]
-				return modules, nil
-			}
-		} else if areModulesPresent && splits[0] != ")" && splits[0] != "" {
-			modules[strings.TrimSpace(splits[0])] = splits[1]
+	for _, dep := range info.Deps {
+		ver := dep.Version
+		if dep.Replace != nil {
+			ver += fmt.Sprintf(" => %s %s", dep.Replace.Path, dep.Version)
```
rhcarvalho, line 51: When the replacement is coming from a local path, there is no version (empty string).

wingyplus: As I try with my local modules. It's mark as version `v0.0.0`

Screen Shot 2563-04-15 at 21 02 14

Screen Shot 2563-04-15 at 20 59 46

wingyplus:
> I do have a suggestion for an alternative design. Here's a bit of a skeleton:
> 
> func extractModules(*debug.BuildInfo) map[string]string {
>     // ...
> }
> 
> // ...
> info, ok := debug.ReadBuildInfo()
> // ...
> _ = extractModules(info)
> 
> func Test...(t *testing.T) {
>     tests := []...
>     // ...
> 
>     for _, tt := range tests {
>         tt := tt
>         // ...
>         t.Run( //...
>            got := extractModules(tt.info)
>            if diff := cmp.Diff(tt.want, got); diff != "" {
>                 // ...
>            }
>         })
>     }
> }
> 
> With this we can have several test cases that each define the input as a *debug.BuildInfo and the corresponding expected map[string]string. This way we are back to simple input-output comparisons and no need for any mocks or other supporting code.
> 
> What do you think?

@rhcarvalho That's great idea! Let's try.

integrations.go
wingyplus: Or we could make same as `go list -m all` presents:

```
github.com/getsentry/sentry-go v0.0.0 => ./pkg/sentry-go  
```

wingyplus: This is my mistake. It should be `dep.Replace.Version` not `dep.Version`.

wingyplus: @rhcarvalho Fixed

integrations_test.go
```diff
@@ -238,3 +241,104 @@ func TestContextifyFramesNonexistingFilesShouldNotDropFrames(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			got := extractModules(tt.info)
+			if diff := cmp.Diff(tt.want, got); diff != "" {
+				t.Errorf("modules info mismatch (-want +got):\n%s", diff)
+			}
```
rhcarvalho, line 341: I took the liberty to change this to use `cmp.Diff` to make it easier to spot problems when tests fail.

wingyplus:
@rhcarvalho thanks!! 😃

integrations_test.go
wingyplus: @rhcarvalho I think we can turn `assertEquals` to uses `cmp.Diff` to display the diff when it's not equal. What do you think?

rhcarvalho: I don't see the point, as it just limits the usability. Personally I prefer calling `cmp.Diff` directly, it is more clear and flexible.

`assertEquals` always calls `t.Fatal*`, even when `t.Error*` would be more appropriate. Moving `cmp.Diff` into `assertEquals` would also limit control of comparison options.

wingyplus: Ah, I see. Thanks for your point.

PR #205, opened 2020-04-16
Use http.RoundTripper in place of *http.Transport

### LINKED ISSUE #204: ClientOptions.Transport should be of type http.RoundTripper
At present, `ClientOptions` contains the following field:

```golang
    // An optional pointer to `http.Transport` that will be used with a default HTTPTransport.
    // Using your own transport will make HTTPProxy, HTTPSProxy and CaCerts options ignored.
    HTTPTransport *http.Transport
 ```

`*http.Transport` is meant to be used as an implementation of the [`http.RoundTripper` interface](https://golang.org/pkg/net/http/#RoundTripper).

Updating all instances of `*http.Transport` to `http.RoundTripper` seems to work perfectly, so this should not even be a breaking change.

flimzy:
Fixes #204

### DIFF
--- a/client.go
+++ b/client.go
@@ -87,7 +87,7 @@ type ClientOptions struct {
 	HTTPClient *http.Client
 	// An optional pointer to `http.Transport` that will be used with a default HTTPTransport.
 	// Using your own transport will make HTTPProxy, HTTPSProxy and CaCerts options ignored.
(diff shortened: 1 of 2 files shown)


flimzy:
Wow, thanks for the speedy merge! Now my unit tests can be greatly simplified! :grinning:

PR #208, opened 2020-04-16
Minor test improvements

rhcarvalho:
Rename test with wrong name.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -288,7 +288,7 @@ func TestCaptureEvent(t *testing.T) {
 	}
 }
 
-func TestCaptureEventShouldSendEventWithProvidedError(t *testing.T) {
+func TestCaptureEventShouldSendEventWithMessage(t *testing.T) {
 	client, scope, transport := setupClientTest()
 	event := NewEvent()
 	event.Message = "event message"


client_test.go
```diff
@@ -312,6 +315,9 @@ func TestBeforeSendGetAccessToEventHint(t *testing.T) {
 	client, scope, transport := setupClientTest()
 	client.options.BeforeSend = func(event *Event, hint *EventHint) *Event {
 		if ex, ok := hint.OriginalException.(customComplexError); ok {
+			if len(event.Exception) != 1 {
```
kamilogorek, line 318: `!= 1` and `missing` are logically not correct, as `x == 2` is not missing, yet it'd not pass :)

rhcarvalho: You're right.

In the newer tests we have we check for the whole "exception" slice -- so it is clear the number of items we expect in the slice.

`if len(...) > 0` would be sufficient to work around the panic in the next line. But the intention with `!= 1` is to guard also against wrong number of events.

Best is to rewrite these tests.

Will iterate on this later.

rhcarvalho:
Travis failure is interesting:

https://travis-ci.com/github/getsentry/sentry-go/jobs/320096411

```console
$ curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/v1.19.1/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.19.1
golangci/golangci-lint info checking GitHub for tag 'v1.19.1'
golangci/golangci-lint crit unable to find 'v1.19.1' - use 'latest' or see https://github.com/golangci/golangci-lint/releases for details
```

Hopefully it is just a flake and doesn't meant the version was retracted ;)

The download error was only in one test job.

rhcarvalho:
Dropped commit 68fb1a8cb00ef2be236ab6b7dbfc56f4451944f1, kept ec278019758da3506317c1a04491ee082c1f5eec rebased on top of current `master`.

AbhiPrasad (approved):

PR #209, opened 2020-04-17
Expose NewEventFromException

dennwc:
Currently there appears to be no way to reuse logic in `Client.eventFromException` and then modify the event to add addition information (e.g. user info, etc).

This change exposes this logic as a new `NewEventFromException` function.

### DIFF
--- a/client.go
+++ b/client.go
@@ -314,7 +314,9 @@ func (client *Client) eventFromMessage(message string, level Level) *Event {
 	return event
 }
 
-func (client *Client) eventFromException(exception error, level Level) *Event {
+// NewEventFromException creates a new Sentry event from a given error.
+// It will automatically populate the stack trace and unwrap supported error types.
+func NewEventFromException(exception error, level Level) *Event {
 	err := exception
 	if err == nil {
 		err = usageError{fmt.Errorf("%s called with nil error", callerFunctionName())}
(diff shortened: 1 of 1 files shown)


rhcarvalho:
At the moment the proposed patch breaks the existing tests.

> modify the event to add addition information (e.g. user info, etc).

Would you mind explaining the problem at hand? What are you trying to achieve and why?

Have you considered using [`BeforeSend`](https://docs.sentry.io/error-reporting/configuration/filtering/?platform=go)?

dennwc:
> Have you considered using `BeforeSend`?

Yes, it would have worked if it wasn't global to the Sentry client.

> Would you mind explaining the problem at hand? What are you trying to achieve and why?

The problem is that I want to attach a user ID, but this ID depends on the request data, and it's not global as `BeforeSend` assumes.

> At the moment the proposed patch breaks the existing tests.

Indeed, but I haven't figured out why that internal method name is not filtered from the stack trace.
And I cannot change the stack trace depth, because that exposed function will be used by both client and SDK code. I can refactor the function to accept a stack depth parameter, but maybe there is a better way?

rhcarvalho:
> The problem is that I want to attach a user ID

This can be done by using Scope, see https://docs.sentry.io/enriching-error-data/scopes/?platform=go.

I've sent a PR to https://github.com/athenianco/metadata/pull/92/ showing how to do it without copying SDK internals and without exposing new APIs.

rhcarvalho:
Since the use case (attach user info to events) is already supported by the SDK via a more general method (Scopes), we'll not export a new symbol.

@dennwc feel free to open a new issue if there's something the SDK cannot do. In general, it is desirable to discuss a problem in an issue before sending a code change, as it might turn out that some design discussion needs to take place, or we may simply find out that something is already possible with the current API.

Thank you for using Sentry and helping making it better!

Issue #212, opened 2020-04-24
Unmarshalable data behaves poorly


flimzy:
I noticed the following in a few of our event logs:

> Original event couldn't be marshalled. Succeeded by stripping the data that uses interface{} type. Please verify that the data you attach to the scope is serializable.

A couple of observations/suggestions (which I'm happy to implement in a PR if desired):

1. It would be very helpful to include the original error message.  I.e.

```golang
	partialMarshallMessage := fmt.Sprintf("Original event couldn't be marshaled: %s. Succeeded by stripping the data " +
		"that uses interface{} type. Please verify that the data you attach to the scope is serializable.", err)
```

2. The message seems to indicate that only some data will be stripped, but it strips all `Extra` data. It would not be difficult (although efficiency would suffer) to loop through each element of `Extra`, and strip only those that don't marshal.

2 would be far less important if 1 were implemented, to make debugging simpler. But both might be ideal.

Thoughts?

flimzy:
Ugh, sorry for the extraneous closing of this. Applying the patch in #213 to our private repo unintentionally closed this.

rhcarvalho:
Let's do "option 1" first. Thanks for sending PRs for the two variations. I took a chance to try to make the info message more concise and still mention specifically which fields got dropped in #258, replacing #214.

We can ship "option 2" later.

PR #214, opened 2020-04-24
Include marshal error in partial martial message

flimzy:
Possible fix for #212 that is slightly less invasive than #213


macnibblet:
I just spent an hour trying to resolve what was going on, and I think this a great addition!

PR #215, opened 2020-04-29
Allow custom stacktrace extraction

flimzy:
This PR makes it possible to use a custom stack-trace extractor function.  There was a TODO comment suggesting this as a possible feature, and I've found we need this feature, so I added it.

I look forward to any feedback.

### DIFF
--- a/client.go
+++ b/client.go
@@ -63,6 +63,10 @@ type ClientOptions struct {
 	Debug bool
 	// Configures whether SDK should generate and attach stacktraces to pure capture message calls.
 	AttachStacktrace bool
+	// StacktraceExtractor is a function to allow a custom implementation
+	// to extract a stack trace from a function. If not provided,
+	// ExtractStacktrace will be used.
+	StacktraceExtractor func(error) *Stacktrace
 	// The sample rate for event submission (0.0 - 1.0, defaults to 1.0).
 	SampleRate float64
 	// List of regexp strings that will be used to match against event's message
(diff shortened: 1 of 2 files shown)


rhcarvalho:
> I've found we need this feature

Are you using an open source error type that is not yet handled in `ExtractStacktrace`?

flimzy:
> Are you using an open source error type that is not yet handled in `ExtractStacktrace`?

No. The biggest thing I need is to change _which_ stack trace is extracted.

Consider a case where you have multiple wrapped errors. The current implementation extracts the outter-most stack trace. We typically need the inner-most.

Example:

```
func foo() error {
    return errors.New("failure!")
}

func bar() error {
    if err := foo(); err != nil {
        return errors.WithStack(err)
    }
}

func baz() eror {
    if err := bar(); err != nil {
        return errors.WithStack(err)
    }
}
```

If we then log the error returned by `baz`, the default behavior will show us the stack from `baz`. We need the stack from `bar`. Providing our own `ExtractStacktrace` implementation makes this possible.

rhcarvalho:
@flimzy what version of the SDK are you using? In 0.6.0, we added support for reporting a complete chain of errors, and that should include the stack for each error in the chain if the stacktrace is available.

flimzy:
> @flimzy what version of the SDK are you using? In 0.6.0, we added support for reporting a complete chain of errors, and that should include the stack for each error in the chain if the stacktrace is available.

We are using 0.6.0, but not yet everywhere.  I was not aware of that new feature. That would probably be sufficient for my needs. Thanks for letting me know.

PR #220, opened 2020-05-08
Fix event init in eventFromException

gukz:
In #185 the `event := NewEvent()` was changed to `event := &Event{Level: level}`. and caused this problem:`panic: assignment to entry in nil map` when you try to set value to `event.Extra`.
Extra useage example: https://github.com/getsentry/sentry-go/blob/master/example/with_extra/main.go#L56

### DIFF
--- a/client.go
+++ b/client.go
@@ -320,7 +320,8 @@ func (client *Client) eventFromException(exception error, level Level) *Event {
 		err = usageError{fmt.Errorf("%s called with nil error", callerFunctionName())}
 	}
 
-	event := &Event{Level: level}
+	event := NewEvent()
+	event.Level = level
 
 	for i := 0; i < maxErrorDepth && err != nil; i++ {
 		event.Exception = append(event.Exception, Exception{


rhcarvalho:
@gukz thanks for the PR!

I am sorry and I take all the blame for having changed the behavior so subtly. I've opened #221 and #222 with ideas for improvements from lessons learned here.

The change was intentional 😓 

It was to avoid allocating maps in all events, even when they are never used. I wanted to eventually remove `NewEvent` from the API in favor of using the `Event` type directly.

Unfortunately, I reckon it does require defensive `nil` checks, and makes it less convenient when mutating events in an `EventProcessor` or `BeforeSend`.

Some not-too-old threads about map initialization in golang-nuts can be illustrative:

- https://groups.google.com/d/msg/golang-nuts/5_8E9OblIho/nvX-T8xKCAAJ
- https://groups.google.com/d/msg/golang-nuts/Q4BS8Xrbg7M/sde7Z7ZwBAAJ

In a nutshell, there is no magical way to initialize maps on-demand.

## Moving forward

I'm considering what we could do best.

To strike a balance between forcing downstream users to `nil`-check/ initialize maps (`nil` check is a good defensive code practice anyway) and always wastefully allocate maps when they may never get used, I'm considering hiding the maps behind "getters" that would take care of lazily initializing the maps.

For example,

```go
event.Extra[key] = val
```

would become

```go
event.Extra()[key] = val
```

This would be implemented similar to https://play.golang.org/p/zb-JJemNMr7.

It is not without its drawbacks; for instance, if you happen to have an existing `map[string]interface{}` you can't simply assign it to `event.Extra`, but you have to iterate over the keys and set values one-by-one.

I imagine that having a pre-existing `map[string]interface{}` is not the common case, though. Even if it was common, it is not as common as not using `Extra`. The example code linked to in the PR description starts off from a `map[string]string` and in that case the for loop is necessary anyway.

It would be a breaking change, though easy to adjust downstream code. The compiler would complain, and one can mass string-replace `Extra` => `Extra()` (and this could be easily automated with tools like [golang.org/x/tools/cmd/eg](https://golang.org/x/tools/cmd/eg)).

@gukz what do you think?

sljeff:
@rhcarvalho 

If it is necessary to remove `NewEvenet`, i think add a `AddExtra` function would be better.

```go
event.AddExtra(key, value)
```

I think `event.Extra()[key] = val` is not very intuitive.

And is it really necessary to delete `NewEvent`? `NewEvent` is already been used in code/tests/docs, and it is an incompatible change for users. Is there more evidence to prove it has performance issues?

Or you will update example. Now my project locked in sentry-go@v0.5.1.

Issue #174, opened 2020-03-03
API change: Capture methods should not return *EventID (*string)


rhcarvalho:
In the last months I've seen in user code a few misuses of the return value of `CaptureException` and related. Seen again today, formalizing the problem here.

This is an example: https://github.com/ruslanBik4/httpgo/blob/@%7B2020.03.02%7D/logs/writer.go#L124-L126

```go
message = string(*(sentry.CaptureException(err)))
```

If `sentry.CaptureException` returns `nil`, and it is documented to return `nil` "in case Sentry is disabled or event was dropped", user code will panic.

I think this is a case where the zero-value (empty string) would make it easier to use the API (or rather harder to misuse).

Perhaps always returning a non-empty string is also reasonable.

And we may as well drop the EventID type altogether.

---

- [ ] Investigate the behavior of other SDKs: how do they represent EventID; is it ever null?.

rhcarvalho:
This is another example https://github.com/getsentry/sentry-go/issues/98#issue-530312849

rhcarvalho:
```go
gitlab.com/gitlab-org/gitaly/internal/dontpanic/retry.go
32-
33-     func() {
34-             defer func() {
35-                     recovered = recover()
36-                     if err, ok := recovered.(error); ok {
37:                             id = sentry.CaptureException(err)
38-                     }
39-             }()
40-             fn()
41-     }()
42-
43-     if id == nil || *id == "" {
44-             return
45-     }
46-
47-     logger.WithField("sentry_id", id).Errorf(
48-             "dontpanic: recovered value sent to Sentry: %+v", recovered,
49-     )
```

The code above does nothing when `id` is `nil`, even though it may also have recovered from a panic.

rhcarvalho:
```go
github.com/anshap1719/go-authentication/utils/log/log.go
18-                     res, err := e(ctx, req)
19-
20-                     if err != nil {
21-                             reqID := ctx.Value(middleware.RequestIDKey).(string)
22-
23:                             event := sentry.CaptureException(errors.WithMessagef(err, "[%s]: [%s] ERROR: %s", prefix, reqID, err.Error()))
24-                             eventID := string(*event)
25-
26-                             sentry.Flush(time.Second * 5)
```

The code above will panic when `event` is `nil`.

It also shows a misuse of `sentry.Flush`.

rhcarvalho:
```go
github.com/cockroachdb/errors/report/report.go
[...]

PR #224, opened 2020-05-18
Make Hub and Scope nil safe

willsilo:
Hi!

This PR makes the `*Hub` and `*Scope` methods safe to call with a nil receiver.

This is useful since it makes instrumentation easier:  The consumer doesn't need to worry about doing nil checks when using `GetHubFromContext`:
```go
func myFunc(ctx context.Context) {
	hub.CaptureMessage("zip zap")
}
```

This is particularly useful in unit test situations where the context (or something else that contains the hub) is not fully populated.  Consumers can still check `hub != nil` if they want to know if the hub is there and the error will be recored.

Thank you for your consideration!

### DIFF
--- a/hub.go
+++ b/hub.go
@@ -82,6 +82,9 @@ func CurrentHub() *Hub {
 
 // LastEventID returns an ID of last captured event for the current `Hub`.
 func (hub *Hub) LastEventID() EventID {
+	if hub == nil {
+		return ""
+	}
 	return hub.lastEventID
 }
 
@@ -105,6 +108,9 @@ func (hub *Hub) stackTop() *layer {
 
 // Clone returns a copy of the current Hub with top-most scope and client copied over.
 func (hub *Hub) Clone() *Hub {
+	if hub == nil {
+		return nil
+	}
 	top := hub.stackTop()
 	if top == nil {
 		return nil
@@ -118,6 +124,9 @@ func (hub *Hub) Clone() *Hub {
 
 // Scope returns top-level `Scope` of the current `Hub` or `nil` if no `Scope` is bound.
 func (hub *Hub) Scope() *Scope {
+	if hub == nil {
+		return nil
+	}
 	top := hub.stackTop()
 	if top == nil {
 		return nil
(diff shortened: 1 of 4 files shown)


willsilo:
Hi @rhcarvalho

Thanks for the fast response!

> As per the current implementation of the PR, hub.CaptureMessage (and everything else) becomes a silent no-op on nil hubs.

Understood.  This is an undeniable drawback.  Personally, though, I believe the risk of my error monitoring causing a panic (in my panic/error handling code, no less) is greater than the risk of missing error capture.  First principle:  Do no harm.  Finding out that the hub is nil may not be readily apparent since the hub is generally used in nonstandard error code paths.

> arrange your code such that context always have a hub (e.g. the http integration/middleware always sets the hub for request handlers)

I have done so with my middleware.  Nonetheless, this does not help with testing situations where the handler may be called outside of the usual routing mechanism.

> write a helper like: GetHubFromContextOrGlobal

This is the approach I intend to pursue.  The downside is needing to import a `sentryhelper` package throughout the codebase, and ensuring that `sentryhelper. GetHubFromContextOrGlobal` is used in place of `sentry.GetHubFromContext`.

I acknowledge there are pros and cons, and my inclination towards this behavior is the result of personal preferences.  Perhaps the majority of others agree with me, or do not! 😄   I value ease of use and safety from panics above all else.

willsilo:
Hey @rhcarvalho

Maybe it's a false choice:  Perhaps there's another solution, like different types, or a global config which determines whether or not a hub is nil safe?

rhcarvalho:
@willsilo I'm 100% in favor of making the SDK easier to use and harder to misuse.
For instance, I've suggested not using pointers for EventID in the API, for I see people dereferencing the pointer without checking (https://github.com/getsentry/sentry-go/issues/174).

Probably if we go to the extent of changing all methods in `Hub` and `Scope` to make them noop on `nil`, we should do the same for all other types, including e.g. `Client`, otherwise a similar example would panic:

```go
func myFunc(ctx context.Context) {
	hub.Client().CaptureMessage("zip zap")
}
```

And we'd need to document that behavior and maintain it going forward (i.e. do the same for every new method and new public types).

Before we make such change I'd like to understand the problem better.

Would you be so kind to explain how you're using hubs in your program? So far it seems you're building web servers and testing handler functions with partially populated context?

Do you have similar expectations to other dependencies you store into `context.Context`? E.g. do you store a handle to a database or something else that becomes noop in tests?

rhcarvalho:
This penalizes every usage of Hub and Scope methods with nil checks and complicates maintenance (adding a new method and forgetting to add the nil check will not be caught by proposed tests).

The use case presented is not convincing enough to justify the change.

If there is a more compelling use case for this, please open an issue and we can revisit.

Issue #230, opened 2020-05-22
Optional request body compression


rhcarvalho:
## Summary

Add option to enable data compression on the transport(s).

## Motivation

Sometimes it might be beneficial to compress request bodies before sending to Sentry.
This would trade some CPU for less bytes over the network.

https://github.com/getsentry/sentry-go/blob/af3076c5259eb407873c99da142af3ee29051105/transport.go#L195-L199

Sentry/Relay accepts compressed payloads, see docs:
- https://develop.sentry.dev/sdk/overview/#request-compression
- https://develop.sentry.dev/sdk/store/#request-compression

## Additional Context

This was an ask/observation from @thinkocapo back in April 27.

thinkocapo:
I found that sentry-javascript 15.13.2 (old) does not use gzip either, at least for how I'm using it in a React application. I checked https://develop.sentry.dev/sdk/store/#request-compression but it does not indicate it's required. I wonder how many sdk's are using it or not. @AbhiPrasad 

Per your comment @rhcarvalho , you mean CPU by the app running Go and the sdk? I didn't think about that trade-off.

rhcarvalho:
@thinkocapo, compression is optional.

- For `@sentry/browser` the need for external dependencies makes compression not attractive;
- For `@sentry/node` I've cut https://github.com/getsentry/sentry-javascript/issues/2613.

> Per your comment @rhcarvalho , you mean CPU by the app running Go and the sdk? I didn't think about that trade-off.

Yes, correct.

An addition consideration / trade-off is additional dependencies.

rhcarvalho:
Updated title/description moving away from `gzip` (specific algorithm) to general compression, also adding link to https://develop.sentry.dev/sdk/overview/#request-compression, which lists the supported compression formats.

PR #231, opened 2020-05-22
feat: Add option to compress requests with gzip

AbhiPrasad:
Closes https://github.com/getsentry/sentry-go/issues/230

## Motivation

The `store` endpoint has [support](https://develop.sentry.dev/sdk/store/#request-compression) for gzipped compressed request bodies. We should implement this functionality for the SDK.

## Implementation

I create a `Encoding` config option, that currently has two options, `json` (our current implementation) and `gzip` (the new addition from this PR). This config option is read by the transport to decide how to encode the request body. This option defaults to `json` (and if incorrectly set, will just use `json` anyway).

## Testing
I wrote a unit test for the function that encodes the request body.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - "I am running away from my responsibilities. And it feels good." – Michael Scott, Season 4, "Money"
+- feat: Add option to compress requests with gzip (#231)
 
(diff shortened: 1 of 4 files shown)


AbhiPrasad:
Closing this for now as there are other priorities, but will come back to this if I have time.

PR #233, opened 2020-05-27
fix: Set either Frame.Filename or Frame.AbsPath

### LINKED ISSUE #216: Relative path in stack traces
I see in #101 that it used to be the case that the path along with the filename was included in the UI for stack traces (similar to how ruby is). Now it's only showing the filename unless you mouse-over, which isn't easy to see a full stack trace.

How do I get the path along with the filename to be included? We use `go build -trimpath` when we build right now so that the full path names don't have meaningless prefixes.

rhcarvalho:
Despite the name "filename", the field is supposed to contain "The path to the source file relative to the project root directory." [1]

Setting `Frame.Filename` to `filepath.Base(FrameAbsPath)` is a problem because it creates ambiguous paths and may affect the "Suspect Commits" feature.

Depending on how the binary is built, the Go runtime knows either the absolute path or the relative path of source files.

In the general case, it can be difficult to derive a correct "project-relative" path. Code built using Go Modules can be anywhere in the file system (not only under GOPATH). GOPATH can be multiple paths.

There was an attempt to trim paths in `raven-go` [2], but the approach there relies on GOROOT and GOPATH as configured for "future builds", and does not reflect how the running binary has been built, thus not being a good general case solution.

For now we set either `Filename` or `AbsPath` depending on what is known, and we leave figuring out a relative path from an absolute path for a later improvement.

[1] https://develop.sentry.dev/sdk/event-payloads/stacktrace#frame-attributes
[2] https://github.com/getsentry/raven-go/blob/5c24d5110e0e198d9ae16f1f3465366085001d92/stacktrace.go#L141

Fixes #216.

### DIFF
--- a/stacktrace.go
+++ b/stacktrace.go
@@ -149,20 +149,36 @@ type Frame struct {
 
 // NewFrame assembles a stacktrace frame out of runtime.Frame.
 func NewFrame(f runtime.Frame) Frame {
-	abspath := f.File
-	filename := f.File
+	var abspath, relpath string
+	// NOTE: f.File paths historically use forward slash as path separator even
+	// on Windows, though this is not yet documented, see
+	// https://golang.org/issues/3335. In any case, filepath.IsAbs can work with
+	// paths with either slash or backslash on Windows.
+	switch {
+	case f.File == "":
+		relpath = unknown
+		// Leave abspath as the empty string to be omitted when serializing
+		// event as JSON.
+		abspath = ""
+	case filepath.IsAbs(f.File):
+		abspath = f.File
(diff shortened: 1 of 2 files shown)


rhcarvalho:
Travis failures are related to downloading github.com/CloudyKit/jet with GO111MODULE=off.

rhcarvalho:
The failure is due to fetching github.com/kataras/iris from master with modules turned off. Since [Iris only supports GO111MODULE=on](https://github.com/kataras/iris/wiki/Installation), we need to disable those tests accordingly.

PR #235, opened 2020-06-01
feat: Add Transactions to sentry-go

AbhiPrasad:
## Motivation

This PR aims to add the concept of a transaction to `sentry-go`. This is important for us so that we can leverage the `sentry-go` transport in the `otel-collector`. This also setups the interfaces so that we can possibly add AM support to `sentry-go` in the future.

## Implementation

We add a struct `type Span` to represent a Sentry span. We also edit the `Event` struct to add three new fields, `StartTimestamp`, `Spans`, and `Type`. By setting `Type == "transaction"`, users will send transaction events. 

The transport code was also changed to account for this. If a `type == "transaction"` is detected, it will use the envelopes endpoint instead.

## Testing

A few unit tests have been added to test the new functions added. This functionality was also demoed with the collector, and was found to be fully functional with no regressions found.


AbhiPrasad:
Seems CI is broken due to https://github.com/go-errors/errors/issues/27

dsn.go
```diff
@@ -152,11 +151,21 @@ func (dsn Dsn) StoreAPIURL() *url.URL {
 	if dsn.path != "" {
 		rawURL += dsn.path
 	}
-	rawURL += fmt.Sprintf("/api/%d/store/", dsn.projectID)
+	rawURL += fmt.Sprintf("/api/%d/%v/", dsn.projectID, s)
```
rhcarvalho, line 154:
```suggestion
	rawURL += fmt.Sprintf("/api/%d/%s/", dsn.projectID, s)
```

`%s` makes the intention explicit that we want a string there, and `go vet` would catch potential type errors.

interfaces.go
```diff
@@ -16,6 +16,9 @@ import (
 // Level marks the severity of the event
 type Level string
 
+// transactionType refers to an transaction event
```
rhcarvalho, line 19:
```suggestion
// transactionType is the type of a transaction event.
```

interfaces.go
```diff
@@ -143,8 +146,42 @@ type Exception struct {
...
+	type alias Span
+
+	return json.Marshal(&struct {
+		*alias
+	}{
+		alias: (*alias)(s),
+	})
+}
+
+// Event is the fundamental data structure that is sent to Sentry
```
rhcarvalho, line 182:
```suggestion
// Event is the fundamental data structure that is sent to Sentry.
```

interfaces.go
```diff
@@ -143,8 +146,42 @@ type Exception struct {
 
 type EventID string
 
-// https://docs.sentry.io/development/sdk-dev/event-payloads/
+// TraceContext describes the context of the trace.
```
rhcarvalho, line 149:
```suggestion
// TraceContext describes the context of the trace.
//
// Experimental: This is part of a beta feature.
```

The extra line creates a new paragraph, may look neater in rendered form.

interfaces.go
```diff
@@ -143,8 +146,42 @@ type Exception struct {
...
+type TraceContext struct {
+	TraceID     string `json:"trace_id"`
+	SpanID      string `json:"span_id"`
+	Op          string `json:"op,omitempty"`
+	Description string `json:"description,omitempty"`
+}
+
+// Span describes a AM Span following the Sentry format.
```
rhcarvalho, line 157: Might be hard to follow in API docs what is "AM".
We could start with something simpler. Sentry is redundant here (there is no confusion with OTel).

```suggestion
// Span describes a timed part of a trace.
```

or 

```suggestion
// Span describes a span.
```

or open for other possibilities ;)

interfaces.go
```diff
@@ -143,8 +146,42 @@ type Exception struct {
...
+// MarshalJSON converts the Span struct to JSON.
+func (s *Span) MarshalJSON() ([]byte, error) {
+	type alias Span
+
+	return json.Marshal(&struct {
+		*alias
+	}{
+		alias: (*alias)(s),
+	})
+}
```
rhcarvalho, line 180:
```suggestion
```

Looks like the custom marshaling is doing the same as the default behavior?

interfaces.go
```diff
@@ -167,15 +204,30 @@ type Event struct {
...
 
+// MarshalJSON converts the Event struct to JSON.
 func (e *Event) MarshalJSON() ([]byte, error) {
 	type alias Event
 	// encoding/json doesn't support the "omitempty" option for struct types.
 	// See https://golang.org/issues/11939.
 	// This implementation of MarshalJSON shadows the original Timestamp field
 	// forcing it to be omitted when the Timestamp is the zero value of
 	// time.Time.
+	if e.Timestamp.IsZero() && e.StartTimestamp.IsZero() {
```
rhcarvalho, line 221: `StartTimestamp` should never be the zero value for valid transactions, and it should always be the zero value (and omitted) for errors.

Maybe we can make this more explicit and defensive here?

pseudo code:

```
if error event {
    always omit StartTimestamp
    always omit Spans
    always omit Type
    omit Timestamp if zero
}
if transaction {
   always include Timestamp and StartTimestamp (they should always be non-zero, server will refuse)
   maybe always include spans? Might be better to see `{spans:[]}` than the omission
   always include type: transaction
}
```

interfaces_test.go
```diff
@@ -29,3 +38,57 @@ func TestNewRequest(t *testing.T) {
...
+				t.Error(err)
+			}
+
+			golden := filepath.Join(".", "testdata", fmt.Sprintf("%s.golden", test.testName))
+			if *update {
+				err := ioutil.WriteFile(golden, got, 0600)
+				if err != nil {
+					t.Error(err)
+				}
+			}
```
rhcarvalho, line 82: Neat touch to add the `-update` flag 👍 

```suggestion
			if *update {
				err := ioutil.WriteFile(golden, got, 0600)
				if err != nil {
					t.Fatal(err)
				}
			}
```

If we can't write the golden file, the test should abort completely -- otherwise it would proceed to read a possibly outdated file from disk next.

interfaces_test.go
```diff
@@ -29,3 +38,57 @@ func TestNewRequest(t *testing.T) {
...
+
+			want, err := ioutil.ReadFile(golden)
+			if err != nil {
+				t.Error(err)
```
rhcarvalho, line 86:
```suggestion
				t.Fatal(err)
```

If we cannot read the file, `want` has no bytes / nothing useful to compare with `got` later.

interfaces_test.go
```diff
@@ -29,3 +38,57 @@ func TestNewRequest(t *testing.T) {
...
+			}
+
+			if !bytes.Equal(got, want) {
+				t.Errorf("struct %s\n\tgot:\n%s\n\twant:\n%s", test.testName, got, want)
```
rhcarvalho, line 90:
```suggestion
				t.Errorf("got:\n%s\nwant:\n%s", got, want)
```

The test name is already part of the test output (because you're in a subtest started with `t.Run`), no need to repeat it here.

I suspect here would be a good use for `cmp.Diff`, otherwise we have to manually diff the two strings...
Did you look at the output when the test fails?

transport.go
```diff
@@ -92,6 +94,41 @@ func getRequestBodyFromEvent(event *Event) []byte {
...
+	fmt.Fprint(&b, "\n", `{"type":"transaction"}`, "\n")
+	b.Write(body)
+	b.WriteString("\n")
+	return &b
+}
+
+func getRequestFromEvent(event *Event, dsn *Dsn) (*http.Request, error) {
+	body := getRequestBodyFromEvent(event)
+	if body == nil {
+		return nil, errors.New("event could not be marshalled")
```
rhcarvalho, line 109: This error is always swallowed?

rhcarvalho: Well, just like `body == nil` was silently ignored before...

transport_test.go
```diff
@@ -125,6 +126,87 @@ func TestGetRequestBodyFromEventCompletelyInvalid(t *testing.T) {
...
+	}
+	sentAtStr := header["sent_at"].(string)
+	if sentAtStr == "" {
+		t.Error("sent_at was not set in header")
+	}
+
+	payload := envParts[2]
+	if payload == "" {
+		t.Error("Item payload is empty")
+	}
```
rhcarvalho, line 158: The intention here is to test that the serialized envelope is correct, right?
In that case, we don't need / should not be trying to do a roundtrip and deserialize the bytes.

Simply compare the bytes to expected output.

transport_test.go
```diff
@@ -125,6 +126,87 @@ func TestGetRequestBodyFromEventCompletelyInvalid(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range testCases {
+		test := test
+		t.Run(test.testName, func(t *testing.T) {
+			dsn, err := NewDsn("https://key@host/path/42")
+			if err != nil {
+				t.Fatal(err)
+			}
```
rhcarvalho, line 192: Please move this out of the loop. The test cases rely on the DSN in their `apiURL` values, and we don't need to re-parse the DSN in every test run.

interfaces.go
AbhiPrasad: Replaced with `Span describes a timed unit of work in a trace`

interfaces_test.go
AbhiPrasad: Good point, the diff will probably be much more useful here, switched to using `cmp.Diff`

transport_test.go
AbhiPrasad: The problem here is that `sentry_at` is generated using `time.Now()`, so I can't just compare strings. Can def compare bytes for the other parts of the envelope though.

transport.go
AbhiPrasad: Yes it's always swallowed currently.

transport_test.go
AbhiPrasad: I changed this so that we split it up to just check if `sent_at` is set, and then we use `cmp.Diff` to compare the rest of the payload.

interfaces.go
AbhiPrasad: Good catch, I completely missed this. I adjusted the `MarshalJSON` for the `Event` to check `e.Type == transactionType`, and then accordingly replicated logic similar to what you described above.

AbhiPrasad:
@rhcarvalho This is ready for another pass.

Summary of changes since last review:
- I adjusted the event marshalJSON to account for transaction specific fields (type, start_timestamp, spans). I also added new tests to account for this new functionality in `interfaces_test.go`
- I made the change from using `%v` to `%s` in the DSN `fmt`
- I adjusted the transport test to not re-parse DSN every test case loop
- I changed the envelope payload test to compare bytes, but still kept some functionality from before to account for the fact that the `sent_at` field is set using `time.Now()`
- I fixed up comments and formatting

transport_test.go
rhcarvalho: This sounds like the typical case to fake time in tests :)

rhcarvalho: Since there are many ways to go about it, I'll give an explicit suggestion on site (the function to create the request body / envelope).

transport.go
```diff
@@ -92,6 +94,41 @@ func getRequestBodyFromEvent(event *Event) []byte {
 	return nil
 }
 
+func getEnvelopeFromBody(body []byte) *bytes.Buffer {
+	var b bytes.Buffer
+	fmt.Fprintf(&b, `{"sent_at":"%s"}`, time.Now().UTC().Format(time.RFC3339Nano))
```
rhcarvalho, line 99:
```suggestion
func getEnvelopeFromBody(body []byte, now time.Time) *bytes.Buffer {
	var b bytes.Buffer
	fmt.Fprintf(&b, `{"sent_at":"%s"}`, now.UTC().Format(time.RFC3339Nano))
```

🌟 tah-dah, now we have a pure function that is easy to test. Given an input, the desired output is deterministic.

transport_test.go
```diff
@@ -125,6 +131,91 @@ func TestGetRequestBodyFromEventCompletelyInvalid(t *testing.T) {
...
 }
 
+func TestGetEnvelopeFromBody(t *testing.T) {
+	body := getRequestBodyFromEvent(&Event{
+		Type:           transactionType,
+		Spans:          []*Span{},
+		StartTimestamp: time.Unix(3, 0).UTC(),
+		Timestamp:      time.Unix(5, 0).UTC(),
+	})
+	env := getEnvelopeFromBody(body)
```
rhcarvalho, line 141: Pick your favorite date to pass to this function with the new argument as in https://github.com/getsentry/sentry-go/pull/235/files#r434621423, then some 20 lines of code can be deleted. No need to parse envelopes :)

transport.go
```diff
@@ -92,6 +94,41 @@ func getRequestBodyFromEvent(event *Event) []byte {
...
+}
+
+func getRequestFromEvent(event *Event, dsn *Dsn) (*http.Request, error) {
+	body := getRequestBodyFromEvent(event)
+	if body == nil {
+		return nil, errors.New("event could not be marshalled")
+	}
+
+	if event.Type == transactionType {
+		env := getEnvelopeFromBody(body)
```
rhcarvalho, line 113:
```suggestion
		env := getEnvelopeFromBody(body, time.Now())
```

transport.go
AbhiPrasad: Wow yes, I should have seen that way earlier 🤦. Fixed.

rhcarvalho: No worries, team work 🦾

transport.go
```diff
@@ -187,17 +223,11 @@ func (t *HTTPTransport) SendEvent(event *Event) {
...
+	request, err := getRequestFromEvent(event, t.dsn)
+	if err != nil {
 		return
 	}
 
-	request, _ := http.NewRequest(
-		http.MethodPost,
-		t.dsn.StoreAPIURL().String(),
-		bytes.NewBuffer(body),
-	)
```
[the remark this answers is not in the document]
AbhiPrasad: Will address this in next PR

PR #247, opened 2020-06-09
fix: wg.Done() must be outside RecoverRepanic

rhcarvalho:
This is a small change to the example for correctness.
The concurrency issue is subtle, but here is an attempt to explain it.

Keep in mind that:
- deferred calls happen in reverse order.
- to recover from panics in f, RecoverRepanic must defer a func call
before calling f.

If defer wg.Done() is part of f, the argument passed to RecoverRepanic,
it means it will be called before any deferred functions setup by
RecoverRepanic itself. If all goroutines started in the loop would call
wg.Done() before an event is reported to Sentry, the main goroutine
blocked on wg.Wait() would unblock and reach the end of func main,
terminating the execution of the program, consequently dropping any
other pending work in other goroutines, reporting to Sentry included.

---

I had this change locally and forgot to include in #241, my bad.

### DIFF
--- a/example/recover-repanic/main.go
+++ b/example/recover-repanic/main.go
@@ -50,14 +50,19 @@ func main() {
 	// Run some worker goroutines to simulate work.
 	for i := 0; i < nWorkers; i++ {
 		wg.Add(1)
-		go RecoverRepanic(func() {
+		go func() {
+			// Note that wg.Done() must be outside of RecoverRepanic, otherwise
+			// it would unblock wg.Wait() before RecoverRepanic is done doing
+			// its job (reporting panic to Sentry).
 			defer wg.Done()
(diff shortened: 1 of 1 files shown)


rhcarvalho:
@lmb, I'm sorry I forgot to include this in my original example.

While this PR doesn't change the implementation of `RecoverRepanic` and thus may not affect your own code, I thought of pinging you to make sure you don't copy my original mistake if you're using `sync.WaitGroup` or other patterns that may need similar deliberate care to make correct.

Also note that, as with `panic()` itself, an unhandled panic in one goroutine brings down the whole program, such that repanicking means that likely only one concurrent panic gets successfully reported to Sentry.

kamilogorek:
Makes sense. Nice explanation 👌

PR #245, opened 2020-06-04
doc: Add comments to all exported types and methods

AbhiPrasad:
Closes https://github.com/getsentry/sentry-go/issues/239

This PR adds comments to all Exported Types and Methods. It also links to the relevant SDK documentation where possible.

There are no code changes, only comments.

Issue #239, opened 2020-06-03
Document all Exported Types


AbhiPrasad:
Currently according to `go-lint` there are 25 instances of: `exported type X should have comment or be unexported`.

We should address this as it is important the sentry-go public API is well documented.

There are 25 total errors,
- [ ] `client.go` (4)
- [ ] `dsn.go` (3)
- [ ] `hub.go` (2)
- [ ] `interfaces.go` (13)
- [ ] `scope.go` (1)
- [ ] `sentry.go` (1)
- [ ] `stacktrace.go` (1)

PR #250, opened 2020-06-16
Improve godoc output

rhcarvalho:
One step towards better godoc output / better docs at godoc.org; not perfect.

## Before

No package overview:

image

Sometimes messy output on the console:

image

## After

Basic package overview:

image

Better output on the console:

image


rhcarvalho:
Inspired by #239 / #245.

PR #254, opened 2020-06-23
feat(am): Add data field to Span interface

AbhiPrasad:
`Data` is a valid field as part of the span interface, we should be including it.

https://develop.sentry.dev/sdk/event-payloads/span/

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -281,4 +281,5 @@ type Span struct {
 	Tags           map[string]string `json:"tags,omitempty"`
 	StartTimestamp time.Time         `json:"start_timestamp"`
 	EndTimestamp   time.Time         `json:"timestamp"`
+	Data           map[string]string `json:"data,omitempty"`
 }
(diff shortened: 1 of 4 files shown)


interfaces.go
```diff
@@ -281,4 +281,5 @@ type Span struct {
 	Tags           map[string]string `json:"tags,omitempty"`
 	StartTimestamp time.Time         `json:"start_timestamp"`
 	EndTimestamp   time.Time         `json:"timestamp"`
+	Data           map[string]string `json:"data,omitempty"`
```
rhcarvalho, line 284: The values are not limited to strings.

```suggestion
	Data           map[string]interface{} `json:"data,omitempty"`
```

(gofmt will likely change the surrounding lines too for alignment, so do not apply the suggestion on GH)

AbhiPrasad: Good catch, fixed.

PR #256, opened 2020-06-25
Avoid expensive operations in CaptureEvent hot path

rhcarvalho:
The original motivation was noticing that calling `rand.New` in `Client.processEvent` was odd.

Writing a small benchmark reviewed that calling `os.Hostname()` for every event was also wasteful.

This PR addresses only those two points and is not meant as a comprehensive optimization effort.

The two changes combined result in this [`benchstat`](//golang.org/x/perf/cmd/benchstat) output:

## go version go1.14.3 darwin/amd64

```
name             old time/op    new time/op    delta
ProcessEvent-16    10.3µs ±15%     0.8µs ± 3%  -92.11%  (p=0.000 n=8+7)

name             old alloc/op   new alloc/op   delta
ProcessEvent-16    6.45kB ± 1%    1.02kB ± 0%  -84.14%  (p=0.000 n=8+8)

name             old allocs/op  new allocs/op  delta
ProcessEvent-16      8.25 ± 9%      6.00 ± 0%  -27.27%  (p=0.000 n=8+8)
```

## go version go1.14.4 linux/amd64

```
name            old time/op    new time/op    delta
ProcessEvent-8    10.9µs ± 0%     1.1µs ± 2%  -89.49%  (p=0.001 n=7+7)

name            old alloc/op   new alloc/op   delta
ProcessEvent-8    6.41kB ± 0%    1.02kB ± 0%  -84.02%  (p=0.000 n=8+8)

name            old allocs/op  new allocs/op  delta
ProcessEvent-8      7.50 ± 7%      6.00 ± 0%  -20.00%  (p=0.000 n=8+8)
```

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -2,6 +2,9 @@ package sentry
 
 import (
 	"errors"
+	"fmt"
+	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


client_test.go
```diff
@@ -322,3 +325,71 @@ func TestBeforeSendGetAccessToEventHint(t *testing.T) {
...
 	assertEqual(t, transport.lastEvent.Message, "customComplexError: Foo 42")
 }
+
+func TestSampleRate(t *testing.T) {
+	tests := []struct {
+		SampleRate float64
+		// tolerated range is [SampleRate-MaxDelta, SampleRate+MaxDelta]
+		MaxDelta float64
+	}{
+		// {0.00, 0.0}, // oddly, sample rate = 0.0 means 1.0, skip test for now.
```
kamilogorek, line 335: It's because of zero-value for floats. I had no idea how to reasonably go around it ¯\_(ツ)_/¯

kamilogorek:
Nice patch!

client_test.go
rhcarvalho: There are a few possibilities like [Functional Options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis); use pointers to detect zero-value vs not-set; etc. All with pros and cons.

Since this is part of `sentry.Init`, any change would break every example code out there :/

PR #257, opened 2020-06-28
example for mux added

vashish1:


rhcarvalho:
Hi @vashish1! Thanks for sending a PR. This example is essentially the same as the `net/http` example.

Since using `github.com/gorilla/mux` or any other `net/http`-compatible middleware doesn't change how to use the SDK, I'm going to keep it out for now.

PR #262, opened 2020-07-17
feat: Extract stack frames from golang.org/x/xerrors

### LINKED ISSUE #246: Support for stack frames from xerrors
## Summary

This is related to the xerrors package: https://godoc.org/golang.org/x/xerrors

For high level context, we are still using this in favor of Go 1.13 stdlib because of the stack trace collection that when logged helps developers understand where the errors are coming from.

An issue was raised before in https://github.com/getsentry/sentry-go/issues/137. While I see that the unwrapping an xerror to its original error is working, I do not see stack traces from xerrors working in Sentry.

## Steps To Reproduce

```go
func foo() {
	sentry.CaptureException(x1())
}

func x1() error {
	return xerrors.Errorf("%w", x2())
}

func x2() error {
	return xerrors.New("x2msg")
}
```

## Expected Behavior

If you print the extended format of the xerror you see a stack trace like:

```
x2msg:
  grax/x/log.x1
    foo/x/log/error_test.go:84
- x2msg:
  grax/x/log.x2
    foo/x/log/error_test.go:88
```

But if I look at Sentry UI I do not see anything about x2 or x1 in the call stack. See the screenshot below.

## Screenshots

Screen Shot 2020-06-05 at 2 23 03 PM

### SDK
 - `sentry-go` version: `github.com/getsentry/sentry-go v0.6.1`
 - Go version: `go version go1.14.3 darwin/amd64`
 - Using Go Modules? yes

### Sentry

## Additional context

Digging into the `sentry-go` code I see support for getting stack frames from interfaces packages like https://github.com/pkg/errors use. But I do not see anything related to the xerrors interface.

https://github.com/getsentry/sentry-go/blob/master/stacktrace.go#L71

I will attempt to patch this, but wanted to open an issue first to discuss.

rhcarvalho:
Closes #246.

### DIFF
--- a/stacktrace.go
+++ b/stacktrace.go
@@ -50,11 +50,13 @@ func NewStacktrace() *Stacktrace {
 func ExtractStacktrace(err error) *Stacktrace {
 	method := extractReflectedStacktraceMethod(err)
 
-	if !method.IsValid() {
-		return nil
-	}
+	var pcs []uintptr
 
-	pcs := extractPcs(method)
+	if method.IsValid() {
+		pcs = extractPcs(method)
+	} else {
+		pcs = extractXErrorsPC(err)
+	}
 
 	if len(pcs) == 0 {
 		return nil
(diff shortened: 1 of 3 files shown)


sk0g:
Is this meant to work with `*xerrors.errorString`, created by `github.com/jackc/pgx/v4 in init`? Doesn't have a stack trace for me, and sentry SDK version is `0.7.0`.

rhcarvalho:
@sk0g yes, it should work. Could you please open a new issue with a minimal example to reproduce what you are observing?

PR #268, opened 2020-07-20
Recover and RecoverWithContext with arbitrary types

### LINKED ISSUE #219: Recover and RecoverWithContext with arbitrary types
`Recover` and `RecoverWithContext` send events to Sentry when they recover from panics and the recovered value has type `error` or `string`. Any other types are ignored.

https://github.com/getsentry/sentry-go/blob/550a179aaa1655092edfba5d2ea4bfc1690736e5/client.go#L233-L284

We should send events no matter the type we're recovering (and need to handle the case in which the value is not serializable as JSON).

rhcarvalho:
Fixes #219.

### DIFF
--- a/client.go
+++ b/client.go
@@ -310,23 +310,22 @@ func (client *Client) RecoverWithContext(
 	if err == nil {
 		err = recover()
 	}
-
-	if err != nil {
-		if hint.Context == nil && ctx != nil {
-			hint.Context = ctx
-		}
-
-		if err, ok := err.(error); ok {
-			event := client.eventFromException(err, LevelFatal)
-			return client.CaptureEvent(event, hint, scope)
-		}
-
-		if err, ok := err.(string); ok {
-			event := client.eventFromMessage(err, LevelFatal)
-			return client.CaptureEvent(event, hint, scope)
-		}
+	if err == nil {
+		return nil
 	}
 
+	if hint.Context == nil && ctx != nil {
+		hint.Context = ctx
+	}
+
+	if err, ok := err.(error); ok {
+		event := client.eventFromException(err, LevelFatal)
+		return client.CaptureEvent(event, hint, scope)
+	}
+	if err, ok := err.(string); ok {
+		event := client.eventFromMessage(err, LevelFatal)
+		return client.CaptureEvent(event, hint, scope)
+	}
 	return nil
 }
 
--- a/client.go
+++ b/client.go
@@ -314,8 +314,13 @@ func (client *Client) RecoverWithContext(
 		return nil
 	}
 
-	if hint.Context == nil && ctx != nil {
-		hint.Context = ctx
+	if ctx != nil {
+		if hint == nil {
+			hint = &EventHint{}
+		}
+		if hint.Context == nil {
+			hint.Context = ctx
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -310,24 +303,29 @@ func (client *Client) RecoverWithContext(
...
 
-	return nil
+	var event *Event
+	switch err := err.(type) {
+	case error:
+		event = client.eventFromException(err, LevelFatal)
+	case string:
+		event = client.eventFromMessage(err, LevelFatal)
+	default:
+		event = client.eventFromMessage(fmt.Sprint(err), LevelFatal)
```
kamilogorek, line 326: Calling `fmt.Sprint` on string is basically a pass-through so maybe we can merge these two cases?

rhcarvalho: Good catch, that's possible, yes.

At some point I was uncertain how this case would get converted into a string. Depending on the type/value, `fmt.Sprint` can produce output that is hard to decipher, whereas something like `fmt.Sprintf("%#v", err)` would make it more verbose and less ambiguous.

E.g. if the value is a pointer to a struct -- https://play.golang.org/p/KJ0gVNq7jV_T

What do you think? If the format is anything other than `"%v"` then we cannot unify the two cases.

rhcarvalho: In fact, `"%#v"` works reasonably for numbers, and might be a better default:

https://play.golang.org/p/GHCYJ1Vyqsm

```
type         %v                    %#v
---          ---                   ---
*main.Person &{}                   &main.Person{Name:""}
main.Person  {}                    main.Person{Name:""}
int          42                    42
float64      5.6                   5.6
[]uint8      [104 101 108 108 111] []byte{0x68, 0x65, 0x6c, 0x6c, 0x6f}
[2]int       [-5 10]               [2]int{-5, 10}
[]string     [  hello]             []string{"", "", "hello"}
string       a string              "a string"
```

When getting a report for a panic with an arbitrary type, what would you rather see in Sentry?

rhcarvalho: `"%+v"` is also a generic option:

https://play.golang.org/p/f7eeaX6v3Zr

```
type         %v                    %+v                   %#v
---          ---                   ---                   ---
*main.Person &{}                   &{Name:}              &main.Person{Name:""}
main.Person  {}                    {Name:}               main.Person{Name:""}
int          42                    42                    42
float64      5.6                   5.6                   5.6
[]uint8      [104 101 108 108 111] [104 101 108 108 111] []byte{0x68, 0x65, 0x6c, 0x6c, 0x6f}
[2]int       [-5 10]               [-5 10]               [2]int{-5, 10}
[]string     [  hello]             [  hello]             []string{"", "", "hello"}
string       a string              a string              "a string"
*main.Foo    &{0x593df8}           &{Bar:0x593df8}       &main.Foo{Bar:(*main.Bar)(0x593df8)}
```

`"%#v"` is the most informative in the general case, without handling deep nested values. Pretty-printing deeply nested values is a more elaborate task, I'd leave it out-of-scope, specially because we don't want to add yet another dependency just for this.

rhcarvalho: Updated this line to `fmt.Sprintf("%#v", err)` and added a few extra test cases that exercise the need for `"%#v"` formatting.

PR #280, opened 2020-10-15
feat: Add types and rename fields for tracing

rhcarvalho:
This overhauls the tracing-related types and their serialization to
JSON.

We rename Event.StartTimestamp, Span.StartTimestamp and
Span.EndTimestamp to remove the "stamp" suffix in preparation to use
StartTime and EndTime for spans when full tracing support is added.

Preferring the shorter name because it is shorter and equally clear, and
the type is time.Time.

This is a breaking change, but should not affect most users, as those
names were only used by the Sentry OpenTelemetry Exporter (we can update
it) and are marked as experimental APIs.

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -186,12 +186,18 @@ type Event struct {
 	Request     *Request               `json:"request,omitempty"`
 	Exception   []Exception            `json:"exception,omitempty"`
 
-	// Experimental: This is part of a beta feature of the SDK. The fields below
-	// are only relevant for transactions.
-	Type           string    `json:"type,omitempty"`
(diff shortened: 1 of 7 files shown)


kamilogorek (approved):

rhcarvalho:
I'm holding off merging this pending the direction that #285 takes.

In particular, `Span` => `RawSpan` may or may not be necessary.

rhcarvalho:
@kamilogorek, sorry for the trouble. I've changed this PR to reflect changes in the initial tracing implementation, should still be easier to review than #285.

HazAT (approved):

PR #285, opened 2020-10-16
feat: Initial tracing implementation

rhcarvalho:
This adds the StartSpan function and related APIs to the SDK.

The initial support focuses on manual instrumentation and HTTP servers
based on net/http.

Tracing is opt-in. Use one of the new options, TracesSampleRate or
TracesSampler, when initializing the SDK to enable sending transactions
and spans to Sentry.

The tracing APIs rely heavily on the standard Context type from Go, and
integrate with the SDKs notion of scopes.

See example/http/main.go for an example of how the new APIs are meant to
be used in practice.

While the basic functionality should be in place, more features are
planned for later.

image

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1,6 +1,9 @@
 /*
 Package sentry is the official Sentry SDK for Go.
 
+Use it to report errors and track application performance through distributed
+tracing.
+
 For more information about Sentry and SDK features please have a look at the
 documentation site https://docs.sentry.io/platforms/go/.
 
@@ -16,6 +19,28 @@ Sentry project. This step is accomplished through a call to sentry.Init.
 
 A more detailed yet simple example is available at
 https://github.com/getsentry/sentry-go/blob/master/example/basic/main.go.
+
+Error Reporting
+
+The Capture* functions report messages and errors to Sentry.
+
+	sentry.CaptureMessage(...)
+	sentry.CaptureException(...)
+	sentry.CaptureEvent(...)
+
+Use similarly named functions in the Hub for concurrent programs like web
+servers.
+
+Performance Monitoring
+
+You can use Sentry to monitor your application's performance. More information
+on the product page https://docs.sentry.io/product/performance/.
+
+The StartSpan function creates new spans.
+
+	span := sentry.StartSpan(ctx, "operation")
+	...
+	span.Finish()
 
 Integrations
 
--- a/http/sentryhttp.go
+++ b/http/sentryhttp.go
@@ -4,6 +4,7 @@ package sentryhttp
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"time"
 
@@ -83,11 +84,23 @@ func (h *Handler) handle(handler http.Handler) http.HandlerFunc {
 		hub := sentry.GetHubFromContext(ctx)
 		if hub == nil {
 			hub = sentry.CurrentHub().Clone()
+			ctx = sentry.SetHubOnContext(ctx, hub)
 		}
+		span := sentry.StartSpan(ctx, "http.server",
+			sentry.TransactionName(fmt.Sprintf("%s %s", r.Method, r.URL.Path)),
+			sentry.ContinueFromRequest(r),
+		)
+		defer span.Finish()
+		// TODO(tracing): if the next handler.ServeHTTP panics, store
+		// information on the transaction accordingly (status, tag,
+		// level?, ...).
+		r = r.WithContext(span.Context())
 		hub.Scope().SetRequest(r)
-		ctx = sentry.SetHubOnContext(ctx, hub)
 		defer h.recoverWithSentry(hub, r)
-		handler.ServeHTTP(w, r.WithContext(ctx))
+		// TODO(tracing): use custom response writer to intercept
+		// response. Use HTTP status to add tag to transaction; set span
+		// status.
+		handler.ServeHTTP(w, r)
 	}
 }
 
--- a/hub.go
+++ b/hub.go
@@ -367,6 +367,15 @@ func GetHubFromContext(ctx context.Context) *Hub {
 	return nil
 }
 
+// hubFromContext returns either a hub stored in the context or the current hub.
+// The return value is guaranteed to be non-nil, unlike GetHubFromContext.
+func hubFromContext(ctx context.Context) *Hub {
+	if hub, ok := ctx.Value(HubContextKey).(*Hub); ok {
+		return hub
+	}
+	return currentHub
+}
+
 // SetHubOnContext stores given Hub instance on the Context struct and returns a new Context.
 func SetHubOnContext(ctx context.Context, hub *Hub) context.Context {
 	return context.WithValue(ctx, HubContextKey, hub)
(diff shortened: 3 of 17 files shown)


http/sentryhttp.go
```diff
@@ -83,11 +83,20 @@ func (h *Handler) handle(handler http.Handler) http.HandlerFunc {
 		hub := sentry.GetHubFromContext(ctx)
 		if hub == nil {
 			hub = sentry.CurrentHub().Clone()
+			ctx = sentry.SetHubOnContext(ctx, hub)
```
rhcarvalho, line 86: ⚠️ This line moved.

Before, we'd unconditionally called `SetHubOnContext`, perhaps creating one unnecessary level of contexts (contexts are like a stack, and there would be two stack items with values pointing to the same hub).

Now, we only update the context if there was no hub previously stored. This is the common case and the reason why the old behavior probably did not affect many people.

What should not have changed is that every HTTP handler in user code will be guaranteed to have a hub ready for use. That is, `GetHubFromContext` never returns `nil` inside of an handler.

hub.go
```diff
@@ -367,6 +367,15 @@ func GetHubFromContext(ctx context.Context) *Hub {
 	return nil
 }
 
+// hubFromContext returns either a hub stored in the context or the current hub.
+// The return value is guaranteed to be non-nil, unlike GetHubFromContext.
+func hubFromContext(ctx context.Context) *Hub {
```
rhcarvalho, line 372: This variation of `GetHubFromContext` is useful in various span methods, because when doing manual instrumentation it makes sense that `StartSpan`, etc, should always find a hub.

I decided not to export for now as users can still use `GetHubFromContext` which has more appropriate semantics in HTTP handlers: handlers should never accidentally use the global hub.

internal/crypto/randutil/randutil_test.go
```diff
@@ -0,0 +1,16 @@
...
+
+func TestFloat64(t *testing.T) {
+	const total = 1 << 24
+	for i := 0; i < total; i++ {
+		n := Float64()
+		if !(n >= 0 && n < 1) {
+			t.Fatalf("out of range [0.0, 1.0): %f", n)
+		}
+	}
+	// TODO: verify that distribution is uniform
```
rhcarvalho, line 15: FWIW, I've verified manually, didn't put time onto a formal validation.

traces_sampler.go
```diff
@@ -0,0 +1,56 @@
...
+// A SamplingContext is passed to a TracesSampler to determine a sampling
+// decision.
+type SamplingContext struct {
+	Span   *Span // The current span, always non-nil.
+	Parent *Span // The parent span, may be nil.
+}
+
+// The TracesSamplerFunc type is an adapter to allow the use of ordinary
+// functions as a TracesSampler.
+type TracesSamplerFunc func(ctx SamplingContext) bool
```
rhcarvalho, line 31: This should make it easy for people to write their own sampling logic if they wish.

traces_sampler.go
```diff
@@ -0,0 +1,56 @@
...
+
+var _ TracesSampler = TracesSamplerFunc(nil)
+
+func (f TracesSamplerFunc) Sample(ctx SamplingContext) bool {
+	return f(ctx)
+}
+
+// fixedRateSampler is a TracesSampler that samples root spans randomly at a
+// uniform rate.
+type fixedRateSampler struct {
```
rhcarvalho, line 41: We'll probably export this sampler in a later release to allow people to compose logic in their custom sampler.

tracing_test.go
```diff
@@ -0,0 +1,315 @@
...
+	// SpanCheck{
+	// 	ZeroTraceID: true,
+	// 	ZeroSpanID:  true,
+	// }.Check(t, span)
+
+	// // Should create a transaction
+	// child := span.StartChild("top")
+	// SpanCheck{
+	// 	RecorderLen: 1,
+	// }.Check(t, child)
```
rhcarvalho, line 314: This test is just a stub for `SpanFromContext`. Contents are commented out because `spanFromContext` doesn't return a "good enough dummy" to allow for the flow described by the test.

In a future pass through the tracing implementation, we either remove `spanFromContext` and everything related to it, or have to implement it such to make this test pass.

rhcarvalho (commented):
Notes for myself / reviewers.

doc.go
```diff
@@ -17,6 +20,28 @@ Sentry project. This step is accomplished through a call to sentry.Init.
 A more detailed yet simple example is available at
 https://github.com/getsentry/sentry-go/blob/master/example/basic/main.go.
 
+Error Reporting
+
+The Capture* functions report messages and errors to Sentry.
+
+	sentry.CaptureMessage(...)
+	sentry.CaptureException(...)
+	sentry.CaptureException(...)
```
kamilogorek, line 29:
```suggestion
	sentry.CaptureEvent(...)
```

tracing.go
```diff
@@ -0,0 +1,587 @@
...
+		panic(err)
+	}
+	if hasParent {
+		span.ParentSpanID = parent.SpanID
+	}
+
+	// Apply options to override defaults.
+	for _, option := range options {
+		option(&span)
+	}
```
kamilogorek, line 138: How does it work exactly? I've seen you do `scope.SetTransaction()` somewhere as the option to this call.

tracing.go
```diff
@@ -0,0 +1,587 @@
...
+// sentryTracePattern matches either
+//
+// 	TRACE_ID - SPAN_ID
+// 	[[:xdigit:]]{32}-[[:xdigit:]]{16}
+//
+// or
+//
+// 	TRACE_ID - SPAN_ID - SAMPLED
+// 	[[:xdigit:]]{32}-[[:xdigit:]]{16}-[01]
+var sentryTracePattern = regexp.MustCompile(`^([[:xdigit:]]{32})-([[:xdigit:]]{16})(?:-([01]))?$`)
```
kamilogorek, line 242: Nice, TIL `:xdigit:`

doc.go
rhcarvalho: Good catch, thanks!

tracing.go
rhcarvalho: Short version:

```go
span := sentry.StartSpan(ctx, "http.server",
	sentry.TransactionName(fmt.Sprintf("%s %s", r.Method, r.URL.Path)),
	sentry.ContinueFromRequest(r),
	func(s *sentry.Span) {
		s.SpanID[4] = 9
		s.Description = "my description"
		// ...
	},
)
```

`StartSpan` takes zero or more functions that take a pointer of span argument (a.k.a. `SpanOption` type), and those functions can mutate the span.

We provide some helpers like `TransactionName` and `ContinueFromRequest` (the equivalent of "FromTraceparent"), but users can also roll their own functions to customize the span, get/set any exported fields.

---

Long version is https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

HazAT (commented):
Why don't we have `startTransaction`?

HazAT (dismissed):

PR #286, opened 2020-10-16
Tracing examples

rhcarvalho:
This builds upon #285 and contains some examples of how to use the new tracing APIs.

In practice, the relationship is the inverse. The examples were written first, before any implementation existed, as a way to inform and guide the new APIs exposed by the SDK.

As is the case with the implementation in #285, this is still a WIP draft.


rhcarvalho:
Decided to merge the initial examples with #285 so it's all in the same place. We may add more examples later, in which case we'll use new PRs.

PR #293, opened 2020-11-11
fix: avoid unnecessary allocation in client.go

derekperkins:
This is a simple fix that avoids allocating an event hint if one was provided to CaptureEvent


rhcarvalho (approved):
Thanks @derekperkins!

PR #298, opened 2020-11-24
fix: only use a single location per frame from xerrors

### LINKED ISSUE #306: 2 frames shown per xerrors wrap instead of 1
## Summary
Stack trace shown on sentry for wrapped xerrors show 2 frames per wrap instead of 1.

## Steps To Reproduce

```go
package main

import (
	"fmt"
	"log"
	"time"

	"golang.org/x/xerrors"
)

func main() {
		Dsn: "THE DSN",
	})
	if err != nil {
		log.Fatal(err)
	}

	err = xerrors.Errorf("context in main: %w", A())
	fmt.Printf("%+v\n", err)

	sentry.CaptureException(err)
}

func A() error {
	err := B{}.Boo()
	return xerrors.Errorf("context in A: %w", err)
}

type B struct{}

func (B) Boo() error {
	err := C()
	return xerrors.Errorf("context in B: %w", err)
}

func C() error {
	return xerrors.New("the xerror")
}
```
Prints:
```
context in main:
    main.main
        /Users/yonderblue/go/github.com/yonderblue/tmpsentry/tmpsentry.go:21
  - context in A:
    main.A
        /Users/yonderblue/go/github.com/yonderblue/tmpsentry/tmpsentry.go:29
  - context in B:
    main.B.Boo
        /Users/yonderblue/go/github.com/yonderblue/tmpsentry/tmpsentry.go:36
  - the xerror:
    main.C
        /Users/yonderblue/go/github.com/yonderblue/tmpsentry/tmpsentry.go:40
```
Shown in UI:
screenshot

## Expected Behavior

A single frame per wrap as in the `%+v` output above. There is little point to showing two, which is not a full stack and at the same time is redundant, adding confusion.

The only reason I can see that two is being shown is simply an artifact/mistake of grabbing all the frames for the PCs that were stored here https://github.com/golang/xerrors/blob/master/frame.go#L31, since that is what was being done for the other error providers. However you can clearly see that while xerrors.Frame does store 3 PCs, it is only because of what runtime.CallersFrames requires. location() only returns a single func/file/line. So in that spirit and for what xerror users are likely looking for in sentry, showing a single frame makes the most sense.

Please see #298

yonderblue:
if you look at https://github.com/golang/xerrors/blob/master/frame.go#L31
you'll notice that while xerrors.Frame does store 3 PCs, it is only because
of what runtime.CallersFrames requires. location() only returns a single
func/file/line. So in that spirit and for what xerror users are likely
looking for in sentry, this change makes a non duplicated stack. In other
words what %+v will print for an error that has wraps underneath, will
also be what sentry shows.

Fixes #306
Related to issue #246

### DIFF
--- a/stacktrace_external_test.go
+++ b/stacktrace_external_test.go
@@ -7,6 +7,7 @@ import (
 	goErrors "github.com/go-errors/errors"
 	pingcapErrors "github.com/pingcap/errors"
 	pkgErrors "github.com/pkg/errors"
+	"golang.org/x/xerrors"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
@@ -49,6 +50,15 @@ func BlueGoErrorsRanger() error {
 	return goErrors.New("this is bad from goErrors")
 }
 
+func RedXErrorsRanger() error {
+	err := BlueXErrorsRanger()
+	return xerrors.Errorf("context in RedXErrorsRanger: %w", err)
+}
+
+func BlueXErrorsRanger() error {
+	return xerrors.New("this is bad from xerrors")
+}
+
 //nolint: scopelint // false positive https://github.com/kyoh86/scopelint/issues/4
 func TestNewStacktrace(t *testing.T) {
 	tests := map[string]struct {
@@ -60,7 +70,7 @@ func TestNewStacktrace(t *testing.T) {
 				{
 					Function: "f1",
 					Module:   "github.com/getsentry/sentry-go_test",
-					Lineno:   18,
+					Lineno:   19,
 					InApp:    true,
 				},
 			},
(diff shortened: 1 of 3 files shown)


stacktrace_external_test.go
```diff
@@ -112,125 +122,71 @@ func TestNewStacktrace(t *testing.T) {
...
 				{
 					Function: "BlueGoErrorsRanger",
 					Module:   "github.com/getsentry/sentry-go_test",
-					Lineno:   49,
+					Lineno:   50,
 					InApp:    true,
 				},
 			},
 		}},
-		// FIXME: The tests below are commented out to avoid introducing a
```
yonderblue, line 168: I saw this comment, but it seems like instead of leaving the test out, it would be best to keep it and remove them all at once in the future if the desire is to move to their own module.

rhcarvalho:
Hi @yonderblue, thanks for the PR. As per our [contribution guidelines](https://github.com/getsentry/sentry-go/blob/master/CONTRIBUTING.md#sending-a-pull-request), could you please file a new issue explaining what problem you have?

On first look this seems to be an opinionated change that deserves some discussion, as the behavior change might not be what other users expect.

yonderblue:
@rhcarvalho just a ping that the issue is there now, thanks

PR #299, opened 2020-11-24
Fixes to JSON payload serialization

rhcarvalho:
- Include [`event_id` (required)](https://develop.sentry.dev/sdk/envelopes/#transaction) and [`length` (recommended)](https://develop.sentry.dev/sdk/envelopes/#items) in transaction envelopes.
- Handle marshaling of `time.Time` values in transactions the same way as for error events. While start and end times are required, the server-side ingestion error shown in Sentry is more clear when the values are omitted than when they are `0001-01-01T00:00:00Z`.
- Stop trying to convert all `time.Time` values to UTC. Instead of doing it inconsistently, allow use of local time. Local time will be serialized with the timezone offset included, and that's handled by Relay (client and server conform to the RFC3339 spec).


CHANGELOG.md
```diff
@@ -151,7 +151,7 @@ Please verify the usage of `sentry.Flush` in your code base.
 
 ## v0.3.0
 
-- feat: Retry event marshalling without contextual data if the first pass fails
+- feat: Retry event marshaling without contextual data if the first pass fails
```
kamilogorek, line 154: It's double `L` :P

https://en.wikipedia.org/wiki/Marshalling_(computer_science)

transport.go
```diff
@@ -72,62 +72,84 @@ func getRequestBodyFromEvent(event *Event) []byte {
...
-		Logger.Println(partialMarshallMessage)
+		Logger.Println(msg)
 		return body
 	}
 
 	// This should _only_ happen when Event.Exception[0].Stacktrace.Frames[0].Vars is unserializable
 	// Which won't ever happen, as we don't use it now (although it's the part of public interface accepted by Sentry)
 	// Juuust in case something, somehow goes utterly wrong.
-	Logger.Println("Event couldn't be marshalled, even with stripped contextual data. Skipping delivery. " +
+	Logger.Println("Event couldn't be marshaled, even with stripped contextual data. Skipping delivery. " +
```
kamilogorek, line 94: same here

transport.go
```diff
@@ -72,62 +72,84 @@ func getRequestBodyFromEvent(event *Event) []byte {
...
+		return nil, err
+	}
+	return &b, nil
 }
 
 func getRequestFromEvent(event *Event, dsn *Dsn) (*http.Request, error) {
 	body := getRequestBodyFromEvent(event)
 	if body == nil {
-		return nil, errors.New("event could not be marshalled")
+		return nil, errors.New("event could not be marshaled")
```
kamilogorek, line 135: and here

transport_test.go
```diff
@@ -130,23 +129,21 @@ func TestGetRequestBodyFromEventCompletelyInvalid(t *testing.T) {
...
-	got := env.String()
-	//nolint: lll
-	want := strings.Join([]string{
-		`{"sent_at":"1970-01-01T00:00:06Z"}`,
-		`{"type":"transaction"}`,
-		`{"sdk":{},"timestamp":"1970-01-01T00:00:05Z","user":{},"type":"transaction","start_timestamp":"1970-01-01T00:00:03Z"}`,
-	}, "\n")
+func TestTransactionEnvelopeFromBody(t *testing.T) {
+	const eventID = "b81c5be4d31e48959103a1f878a1efcb"
+	sentAt := time.Unix(0, 0).UTC()
```
kamilogorek, line 134: Do you use `UTC` on purpose here? Because a default for `transactionEnvelopeFromBody` is not using it.

rhcarvalho: We need either UTC or any other fixed location, otherwise the test output will depend on the time zone of the computer running the test.

CHANGELOG.md
rhcarvalho: AFAICT this is UK vs US spelling: https://wikidiff.com/marshaled/marshalled.
Occurrences in [encoding/json](https://golang.org/pkg/encoding/json/) all use single L.

kamilogorek (approved):

PR #304, opened 2020-11-27
doc: Revamp sentryhttp documentation

rhcarvalho:
The major point is to promote instrumenting higher up in the HTTP
handler hierarchy, such that Sentry instrumentation takes only one line,
instead of wrapping many individual handlers.

- Add go doc example with minimal steps for instrumenting web servers.
- Overhaul doc comments.
- Replace executable example with something a little bit more realistic,
which we can build upon to show tracing instrumentation later.

### DIFF
--- a/example/http/main.go
+++ b/example/http/main.go
@@ -1,7 +1,23 @@
+// This is an example web server to demonstrate how to instrument web servers
+// with Sentry.
+//
+// Try it by running:
+//
+// 	go run main.go
+//
+// To actually report events to Sentry, set the DSN either by editing the
+// appropriate line below or setting the environment variable SENTRY_DSN to
+// match the DSN of your Sentry project.
 package main
 
 import (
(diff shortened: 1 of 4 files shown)


example/http/main.go
```diff
@@ -10,67 +26,156 @@ import (
...
+	return img
+}
+
+// configureLoggers configures the standard logger and the logger used by the
+// Sentry SDK.
+//
+// The only reason to change logger configuration in this example is aesthetics.
+func configureLoggers() {
+	logFlags := log.Ldate | log.Ltime | log.Lmsgprefix
+	sentry.Logger.SetPrefix("[sentry sdk  ] ")
```
kamilogorek, line 177:
```suggestion
	sentry.Logger.SetPrefix("[sentry sdk] ")
```

kamilogorek (approved):

Issue #253, opened 2020-06-23
Add ability to add custom stacktraces


Mokto:
Hi,

# Summary
Add ability to add custom stacktraces to the ones that sentry-go is finding automatically

## Motivation
I'm using github.com/ztrue/tracerr that contains all my traces when anything goes wrong.

I'd like to add that information to Sentry, as a stacktrace. The information is not as accurate as the one from tracerr.

I'm currently pushing it as breadcrumbs, which is definitely not ideal (especially since we can't "clear" breadcrumbs)

##

Thanks!

PS : it might be out of scope, I might misunderstand something about the plugin. If so, I'm sorry :)

rhcarvalho:
This is a valid request and something we intend to support.

I have a WIP branch to rework the way how stack traces are extracted and it should make room for extraction from custom error types.

> I'm currently pushing it as breadcrumbs, which is definitely not ideal (especially since we can't "clear" breadcrumbs)

@Mokto, doesn't [Scope.ClearBreadcrumbs](https://pkg.go.dev/github.com/getsentry/sentry-go@v0.7.0?tab=doc#Scope.ClearBreadcrumbs) work for you?

Mokto:
It should work indeed ;) I will try soon.

Thanks!

derekperkins:
This is important for us. We're capturing stacktraces and annotating them with meta (breadcrumbs) as we move up the stack, but aren't reporting until we get back up to a root level handler. In `raven-go`, we have a simple conversion function that works great.

```
// sentryStacktrace converts our stack representation to Sentry's
func (err *Err) sentryStacktrace() *sentry.Stacktrace {
	frames := make([]*sentry.StacktraceFrame, 0, len(err.frames))
	for i := len(err.frames) - 1; i >= 0; i-- {
		f := err.frames[i]
		frames = append(frames, &sentry.StacktraceFrame{
			Filename:     f.file,
			Function:     f.fn,
			Lineno:       f.line,
			AbsolutePath: f.path,
			InApp:        f.class == classApp,
		})
	}
	return &sentry.Stacktrace{
		Frames: frames,
	}
}
```

What I can't do today is annotate those frames with their "breadcrumb" data, so that would be very nice. At the end of the day, we're wrapping our errors and just want to convert those and send them to Sentry without a ton of SDK magic. Is that in scope for this SDK or should we look at just using the api directly?

derekperkins:
[...]

PR #310, opened 2020-12-01
Allow custom error types to emit more specific Stacktraces

jamescun:
This change creates a new interface, `StacktraceGetter`, which allows
custom error types to embed a more specific Stacktrace within
themselves to be caught by more generic error handlers, and updates
`ExtractStacktrace` to rely on this if implemented.

In addition to more specific Stacktraces, this partially de-couples
embedded Stacktraces from their implementation and creates an opt-in
integration of third-party error packages, rather than expanding the
scope of reflection in `ExtractStacktrace`. The `GetSentryStracktrace`
in `StacktraceGetter` explicitly references Sentry to ensure this
interface is opt-in and unlikely to conflict with existing Stacktrace
implementations.

This should also work with `Recover*` handlers without modification, in
the case where custom error types are being thrown.


Duffleman (approved):

yonderblue:
Would be great to convert the existing support for the error pkgs to use this.

rhcarvalho:
Related to #253, but only solves the problem if you control the error types you're dealing with (won't work for third-party error libraries).

Issue #240, opened 2020-06-03
gRPC interceptors


rcmachado:
## Summary

Provide gRPC interceptors to automatically capture errors and panics from servers and clients, both for unary and stream requests.

## Motivation

gRPC is a high-performance, open source universal RPC framework originally developed by Google. It supports languages/platforms such as Python, Ruby, Kotlin and Go.

Go SDK already has some integrations with standard `net/http` package and some popular frameworks (like [Gin](https://docs.sentry.io/platforms/go/gin/), [Martini](https://docs.sentry.io/platforms/go/martini/) and others).

Similar to other frameworks, having standard Sentry integration will make it easier for users to integrate Sentry with gRPC.

## Additional Context

gRPC has the concept of interceptor (middleware), which could make it easier to implement an official Sentry integration. An interceptor is a function with a specific signature/type:

```go
// Server interceptors (middlewares)
type UnaryServerInterceptor func(ctx context.Context, req interface{}, info *UnaryServerInfo, handler UnaryHandler) (resp interface{}, err error)
type StreamServerInterceptor func(srv interface{}, ss ServerStream, info *StreamServerInfo, handler StreamHandler) error

// Client interceptors (middlewares)
type UnaryClientInterceptor func(ctx context.Context, method string, req, reply interface{}, cc *ClientConn, invoker UnaryInvoker, opts ...CallOption) error
type StreamClientInterceptor func(ctx context.Context, desc *StreamDesc, cc *ClientConn, method string, streamer Streamer, opts ...CallOption) (ClientStream, error)
```

Using an interceptor is also straightforward (pseudo-code):

```go
// Sentry initialization
err := sentry.Init(sentry.ClientOptions{
	// Either set your DSN here or set the SENTRY_DSN environment variable.
	Dsn: "https://7a0d561762b749e79de56c58bb0d898b@o24338.ingest.sentry.io/1307339",
	// Enable printing of SDK debug messages.
	// Useful when getting started or trying to figure something out.
	Debug: true,
})
if err != nil {
	log.Fatalf("sentry.Init: %s", err)
}
// Flush buffered events before the program terminates.
// Set the timeout to the maximum duration the program can afford to wait.
defer sentry.Flush(2 * time.Second)

// Initialize gRPC server
myServer := grpc.NewServer(
    grpc.StreamInterceptor(
        sentry.StreamServerInterceptor(),
    ),
    grpc.UnaryInterceptor(
        sentry.UnaryServerInterceptor(),
    ),
[...]

PR #312, opened 2020-12-18
Add grpc unary server interceptor

shouichi:
This is a draft PR to support grpc (see #240). The main goal is to gather feedback and agree on the interface. Thus tests are not written yet. Any feedback/comments are welcomed. Thanks!


rhcarvalho (commented):
@shouichi thanks for the PR. Sorry, I didn't have the bandwidth to review it over the holiday season.

The current draft looks promising, would love to see some test cases and examples how it will be used.

shouichi:
@rhcarvalho Added tests. Please take a look.

grpc/server.go
```diff
@@ -0,0 +1,84 @@
...
+	if opts.ReportOn == nil {
+		opts.ReportOn = ReportAlways
+	}
+
+	return func(
+		ctx context.Context,
+		req interface{},
+		info *grpc.UnaryServerInfo,
+		handler grpc.UnaryHandler,
+	) (_ interface{}, err error) {
```
rhcarvalho, line 24: Either name all return values or none. I glanced over other unary interceptor implementations, I think the common case is the former:

```suggestion
	) (resp interface{}, err error) {
```

grpc/server.go
```diff
@@ -0,0 +1,84 @@
...
+
+		defer func() {
+			if r := recover(); r != nil {
+				hub.RecoverWithContext(ctx, r)
+
+				if opts.Repanic {
+					panic(r)
+				}
+
+				err = status.Errorf(codes.Internal, "%s", r)
```
rhcarvalho, line 39: Is it intentional that this line will not run if repanic is true?

grpc/server_test.go
```diff
@@ -0,0 +1,293 @@
...
+			err:   status.Error(codes.Aborted, ""),
+			codes: []codes.Code{codes.Canceled},
+			want:  false,
+		},
+	} {
+		tt := tt
+
+		t.Run(tt.name, func(t *testing.T) {
+			if w, g := tt.want, sentrygrpc.ReportOnCodes(tt.codes...)(tt.err); w != g {
+				t.Fatalf("ReportOnCodes: want %t, got %t", w, g)
```
rhcarvalho, line 270: "ReportOnCodes" is already part of the test name, so we don't need to repeat it in the error message:

```suggestion
			if got := sentrygrpc.ReportOnCodes(tt.codes...)(tt.err); got != tt.want {
				t.Fatalf("got %t, want %t", got, tt.want)
```

grpc/server_test.go
```diff
@@ -0,0 +1,293 @@
...
+			}
+		})
+	}
+}
+
+func ExampleUnaryServerInterceptor() {
```
rhcarvalho, line 276: Thanks for adding the example 👍

grpc/server_test.go
```diff
@@ -0,0 +1,293 @@
...
+	"testing"
+	"time"
+
+	"github.com/getsentry/sentry-go"
+	sentrygrpc "github.com/getsentry/sentry-go/grpc"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	grpchealth "google.golang.org/grpc/health/grpc_health_v1"
```
rhcarvalho, line 16: Is it relevant for the tests that we use this package, or can we use `google.golang.org/grpc` by itself?

rhcarvalho (commented):
Looks very good!

A couple of questions / suggestions. I think we can merge this soon, thank you!

grpc/server.go
shouichi: Yes, this is intentional. The idea is to re-panic and to let users do whatever they want. Actually, I copied this re-panic feature from sentryhttp package but maybe we want to remove this feature?

grpc/server_test.go
shouichi: Can you elaborate?

As far as I understand, to test grpc server we need a service to serve (in this case, I used grpc_health_v1 service). The service doesn't matter but grpc_health_v1 looked the easiest because it is bundled with grpc.

grpc/server_test.go
shouichi: > "ReportOnCodes" is already part of the test name, so we don't need to repeat it in the error message:

Got it.

I personally prefer to use the variable `w` not to repeat `tt.want`. But I'm OK to change if the suggested style is widely used in the sentry-go codebase.

grpc/server.go
shouichi: Sounds reasonable. I'll fix 👍

grpc/server.go
rhcarvalho: Never mind, my mistake. If repanic is true the function will not return and then it doesn't matter if we set `err` or not.

grpc/server_test.go
rhcarvalho: I just looked again at pkg.go.dev, for some reason I thought this was in a separate module, but it is not.
It's okay to use it.

grpc/server_test.go
```diff
@@ -0,0 +1,293 @@
...
+			codes.OutOfRange,
+			codes.Internal,
+		),
+		// Recovers from panic, reports it and returns internal error.
+		Repanic: false,
+	}
+
+	// This middleware sets *sentry.Hub to context. You can set user to
+	// hub's scope in the later interceptor for example.
+	sentry := sentrygrpc.UnaryServerInterceptor(opts)
```
rhcarvalho, line 289: @shouichi do you think the default options is what will be most used?

I know `sentryhttp` and friends follow this pattern of taking an option struct, but for new code I often consider if we can use the [functional options pattern](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis).

As an example, there's a similar interceptor in [opentracing-contrib/go-grpc](https://github.com/opentracing-contrib/go-grpc/blob/4a6760c71486b5e9aa6faf22b11d579e13a119d9/server.go#L27).

```go
func OpenTracingServerInterceptor(tracer opentracing.Tracer, optFuncs ...Option) grpc.UnaryServerInterceptor {
```

What the different style would allow is for:

- No empty struct as required argument
  ```go
  sentrygrpc.UnaryServerInterceptor()
  ```
  vs
  ```go
  sentrygrpc.UnaryServerInterceptor(sentrygrpc.UnaryServerInterceptorOptions{})
  ```

The `grpc` library itself uses (a form of) functional options.

- Possibility to extend the options and having defaults that do not necessarily match Go zero values.

grpc/server.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/getsentry/sentry-go"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// UnaryServerInterceptor is a grpc interceptor that reports errors and panics
+// to sentry. It also sets *sentry.Hub to context.
+func UnaryServerInterceptor(opts UnaryServerInterceptorOptions) grpc.UnaryServerInterceptor {
```
rhcarvalho, line 14: @shouichi do you envision there being some non-unary server interceptor later? If not, we could perhaps shorten the name to `ServerInterceptor`, so that call sites become `sentrygrpc.ServerInterceptor`.

Asking because I see in OpenTracing they name a similar interceptor [`OpenTracingServerInterceptor`](https://github.com/opentracing-contrib/go-grpc/blob/4a6760c71486b5e9aa6faf22b11d579e13a119d9/server.go#L27) (used as `otgrpc.OpenTracingServerInterceptor`, the "OpenTracing" prefix being a bit redundant).

grpc/server_test.go
shouichi: > do you think the default options is what will be most used?

Currently yes, but in the future, I'm not sure. The functional options pattern might be friendlier to users. My personal preference is to be consistent. So if the future direction is to use the functional options pattern, I'm happy to use it.

I often find myself wondering about options struct vs functional options. It seems that the go stdlib embraces zero values but I see a lot of usage of functional options pattern in the wild. It's hard to pick one!

grpc/server.go
shouichi: Honestly not in the near future (because we don't use grpc streams). I'm happy to contribute tho ;)

OpenTelemetry (which I believe a successor of OpenTracing) uses `UnaryServerInterceptor`. I personally prefer to prefix "Unary" because it is explicit and will be consistent after adding a stream interceptor.

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/65367f2b373efda45942c5c37b4332a31b226b98/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go#L304

grpc/server_test.go
rhcarvalho: > It seems that the go stdlib embraces zero values but I see a lot of usage of functional options pattern in the wild.

The pattern is something that emerged many years later when the stdlib was already solidified by the Go 1 compatibility promise.
But new experiments like the [new HTTP client](https://github.com/bradfitz/exp-httpclient) use functional options.

grpc/server.go
rhcarvalho: 👍 

Indeed it shall be `UnaryServerInterceptor`, so later we can also have `StreamServerInterceptor`.

grpc/server_test.go
shouichi: Cool! Didn't know that.

OK, I'll change to use the functional options pattern.

grpc/options.go
```diff
@@ -0,0 +1,64 @@
...
+
+	for _, f := range ff {
+		f(&opts)
+	}
+
+	return opts
+}
+
+// Option configures reporting behavior.
+type Option func(*options)
```
shouichi, line 26: I wonder if we should separate options for server/client and unary/stream. Probably they won't diverge too much?

shouichi (commented):
@rhcarvalho Changed to use the functional options pattern. Please take a look.

grpc/options.go
rhcarvalho: Good point, ~~perhaps what would give us more flexibility is~~

```suggestion
type Option interface{
	apply(*options)
}
```

~~Later down the road it could become~~

```go
type Option interface{
	applyServer(*serverOptions)
	applyClient(*clientOptions)
	// ...
}
```

It would allow us to share options between servers and clients, unary and stream.
Though I'm not sure that's actually a good idea. One downside is discoverability and documentation. If server/client/etc options have the same type, all of them will be grouped together in godoc. If an option doesn't make sense to a certain type (client/server/unary/stream), that would also require documentation to explain instead of being impossible to compile because of incompatible types.

I think the options we have right now are "interceptor options", and should probably be the same for other types of interceptors? Are there other existing concepts we'd expand into?

```suggestion
type InterceptorOption interface{
	apply(*options)
}
```

If there is only reason to be one type of option, for interceptors, then we can keep the name simple, `Option`.

---

Sorry, busy day, didn't have much time to think this through but wanted to share and continue the conversation.

grpc/options.go
```diff
@@ -0,0 +1,63 @@
...
+		c := status.Code(err)
+		for i := range cc {
+			if c == cc[i] {
+				return true
+			}
+		}
+
+		return false
+	}
+}
```
spaceweasel, line 63:
```suggestion
	cm := make(map[codes.Code]bool)	
	for _, c := range cc {
		cm[c] = true
	}
	return func(err error) bool {
		return cm[status.Code(err)]
	}
}
```

grpc/options.go
shouichi: I'm not really sure but both of the following changes shouldn't break user code because they don't directly use `Option`. So can we start simple?

- Changing `type Option func(*options)` to `type Option interface {}`.
- Renaming `Option` to `InterceptorOption`.

> Sorry, busy day, didn't have much time to think this through but wanted to share and continue the conversation.

No worries! We don't need to rush. I (we?) just want to agree on a good design ;)

grpc/options.go
shouichi: Thanks, using a map should be faster.

shouichi:
@rhcarvalho @spaceweasel Hi, it's been a while but could you take another look at this PR?

becoded:
I was looking for this. Glad to see it's only a review away from getting it merged!

odannyc:
Hi, it's been some time since this has been open. Are there plans to merge this in?

@rhcarvalho @spaceweasel

Vesquen:
This is exactly what I'm looking for. The pull request has been open for a while. Any chance this still can get a review? @rhcarvalho @spaceweasel

koteesy:
Any news?

serbanmarti:
Really looking forward to this getting merged!

legosx:
Can't wait to see it's merged. This is exactly I was looking for! Thanks a lot

prasad-marne:
Any updates.  If it needs any help, i can do it.

pablodz:
Still relevant, please update status

flyte:
Could this maybe be turned into a separate package if it's not going to be merged?

HurSungYun:
there's a seperate package here but code is not exactly same.

https://github.com/johnbellone/grpc-middleware-sentry

using this package until this PR merged can be a temporal solution.

cleptric:
Closing, as we'll add this in https://github.com/getsentry/sentry-go/pull/938.

Issue #316, opened 2021-01-13
Resources leak through net/http transport goroutines


laoleesch:
## Summary
Hi! I found that there is a resource leak problem in case of lot of events. The amount of goroutines, used fd and consumed memory is growing and does not try to decrease.
In my case I use sentry-go in logrus hook plugin and got thousands goroutines and fd limit in case of thousands of events.

For each transport flush there hung a couple "zombie"-routines
net/http.(*persistConn).readLoop
net/http.(*persistConn).writeLoop

## Steps To Reproduce

Any code with big amount events to flush.
https://github.com/onrik/logrus/blob/master/sentry/sentry.go

## Expected Behavior

Http.Client using keep-alive connections.

## Screenshots

<img width="445" alt="image" src="https://user-images.githubusercontent.com/6000186/104446957-3103ca80-55ac-11eb-8605-de7a2eb903d1.png">

### SDK
 - `sentry-go` version: 0.9.0
 - Go version: 1.15
 - Using Go Modules? yes

laoleesch:
I suppose a simple fix in PR [#317](https://github.com/getsentry/sentry-go/pull/317)

rhcarvalho:
Dang... thanks for catching it @laoleesch and thanks for the PR.

Reinforces https://github.com/bradfitz/exp-httpclient/blob/master/problems.md#too-easy-to-not-call-responsebodyclose

PR #317, opened 2021-01-13
fix: net/http transport goroutines leak

laoleesch:
Hi! For each transport flush there hung a couple "zombie"-routines 
net/http.(*persistConn).readLoop
net/http.(*persistConn).writeLoop

I suppose it because there is no response.Body.Close() in transport cycles. This not only leads to a zombie-routines but also because of the lack of re-using connections to the loss of performance (no keep-alive).

[net/http/#Client.Do](https://golang.org/pkg/net/http/#Client.Do)

> If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close. If the Body is not both read to EOF and closed, the Client's underlying RoundTripper (typically Transport) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.

It solves resource leak issue in my cases.

Thanks!

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -386,7 +386,7 @@ func (t *HTTPTransport) worker() {
 				Logger.Printf("Too many requests, backing off till: %s\n", deadline)
 			}
 
-			io.CopyN(ioutil.Discard, response.Body, 512)
+			_, err = io.CopyN(ioutil.Discard, response.Body, 512)
 			response.Body.Close()
 		}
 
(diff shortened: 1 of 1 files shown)


laoleesch:
Issue #316

rhcarvalho (commented):
I think we also need to drain the response body, could use `io.CopyN` (see https://github.com/google/go-github/pull/317#issuecomment-204099820).
Not sure this is still relevant in recent Go versions, the docs suggest so.

https://golang.org/pkg/net/http/#Client.Do
> If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close. **If the Body is not both read to EOF and closed,** the Client's underlying RoundTripper (typically Transport) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.

https://golang.org/pkg/net/http/#Response
> // [...] It is the caller's responsibility to
> // close Body. The default HTTP client's Transport may not
> // reuse HTTP/1.x "keep-alive" TCP connections if the Body is
> // **not read to completion and closed**.

PR #322, opened 2021-01-18
fix: drain and close API response bodies

### LINKED ISSUE #316: Resources leak through net/http transport goroutines

rhcarvalho:
This fixes the two transport implementations to drain (up to a safe limit) and close the response body from API calls, allowing the net/http client to reuse connections.

Fixes #316.
Closes #317 (replaces it).

Thanks @laoleesch.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -6,6 +6,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -16,6 +18,17 @@ import (
 const defaultBufferSize = 30
 const defaultRetryAfter = time.Second * 60
 const defaultTimeout = time.Second * 30
+
+// maxDrainResponseBytes is the maximum number of bytes that transport
+// implementations will read from response bodies when draining them.
+//
+// Sentry's ingestion API responses are typically short and the SDK doesn't need
+// the contents of the response body. However, the net/http HTTP client requires
+// response bodies to be fully drained (and closed) for TCP keep-alive to work.
+//
+// maxDrainResponseBytes strikes a balance between reading too much data (if the
+// server is misbehaving) and reusing TCP connections.
+const maxDrainResponseBytes = 16 << 10
 
 // Transport is used by the Client to deliver events to remote server.
 type Transport interface {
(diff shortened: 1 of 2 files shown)


transport.go
```diff
@@ -17,6 +19,17 @@ const defaultBufferSize = 30
...
+// maxDrainResponseBytes is the maximum number of bytes that transport
+// implementations will read from response bodies when draining them.
+//
+// Sentry's ingestion API responses are typically short and the SDK doesn't need
+// the contents of the response body. However, the net/http HTTP client requires
+// response bodies to be fully drained (and closed) for TCP keep-alive to work.
+//
+// maxDrainResponseBytes strikes a balance between reading too much data (if the
+// server is misbehaving) and reusing TCP connections.
+const maxDrainResponseBytes = 16 << 10
```
kamilogorek, line 31: Not sure why `16 << 10` instead of `1024 * 16` or just `16384`, but ok :p

transport.go
```diff
@@ -472,15 +488,18 @@ func (t *HTTPSyncTransport) SendEvent(event *Event) {
 	)
 
 	response, err := t.client.Do(request)
-
 	if err != nil {
```
kamilogorek, line 491: I assume that `err` and `response` are mutually exclusive, and only one of them can be `nil` at the time?

transport_test.go
```diff
@@ -361,3 +363,102 @@ func TestHTTPTransport(t *testing.T) {
...
+
+	// If server responses are too large, the SDK should close the
+	// connection instead of consuming an arbitrarily large number of bytes.
+	largeResponse = true
+
+	// Next event, first one to get a large response, reuses the connection.
+	tr.SendEvent(event)
+	checkLastConnReuse(true)
+
+	// All future events create a new TCP connection.
```
kamilogorek, line 450: Should they all create new TCP connections, or only the first one after the previous one was closed (due to too large drainage), and remaining 9 should reuse it?

kamilogorek: Ah, `largeResponse` is still kept `true` here, so they all should use new I guess.

kamilogorek (approved):

transport.go
rhcarvalho: https://golang.org/pkg/net/http/#Client.Do

>  On error, any Response can be ignored. A non-nil Response with a non-nil error only occurs when CheckRedirect fails [...]

transport_test.go
rhcarvalho: Yes. The tests cover two types of server behavior, first a server that behaves like current-times Relay, and second simulates sending more data, more than we're willing to read and discard. In the latter case, keep-alive should not work unless the Go HTTP client changes in a future release.

PR #323, opened 2021-01-18
feat: debug connection reuse

rhcarvalho:
Add a new SENTRYGODEBUG option httptrace to print DSN resolution and TCP
connection information, including whether connections are being reused.

Rename internal debug option dumphttp -> httpdump for symmetry with
httptrace (that derives from the net/http/httptrace package name).

The new option was used to reproduce / debug #316.

### DIFF
--- a/client.go
+++ b/client.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"reflect"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -212,10 +213,26 @@ func NewClient(options ClientOptions) (*Client, error) {
...
+	// options.Debug, which is also available for SDK users.
+	dbg := strings.Split(os.Getenv("SENTRYGODEBUG"), ",")
+	sort.Strings(dbg)
+	// dbgOpt returns true when the given debug option is enabled, for
+	// example SENTRYGODEBUG=someopt=1.
+	dbgOpt := func(opt string) bool {
+		s := opt + "=1"
+		return dbg[sort.SearchStrings(dbg, s)%len(dbg)] == s
+	}
+	if dbgOpt("httpdump") || dbgOpt("httptrace") {
```
kamilogorek, line 230: That's hell of a strange way to parse :D

internal/debug/transport.go
```diff
@@ -1,38 +1,72 @@
...
+	Dump bool
+	// Trace enables usage of net/http/httptrace.
+	Trace bool
 }
 
 func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
-	b, err := httputil.DumpRequestOut(req, true)
-	if err != nil {
-		return nil, err
+	var buf bytes.Buffer
```
kamilogorek, line 24: Maybe rename `buf` to something more explicit? I had hard time follow between `b` and `buf` interchangeably.
`debugLogBuffer`? `logBuffer`?

kamilogorek (approved):

PR #327, opened 2021-02-13
fix: Do not override trace context of transactions

### LINKED ISSUE #318: [0.9.0] wrong root span operation name in UI
## Summary
See title

## Steps To Reproduce

```go
func main() {
	ctx := context.Background()

	opts := sentry.ClientOptions{
		Debug:            true,
		TracesSampleRate: 1.0,
		Dsn:              "https://THE_DSN",
	}

	if err := sentry.Init(opts); err != nil {
		panic(err)
	}
	defer sentry.Flush(3 * time.Second)

	sp1 := sentry.StartSpan(ctx, "span 1", sentry.TransactionName("the transaction"))
	defer sp1.Finish()
	for i := 0; i < 2; i++ {
		sp2 := sp1.StartChild("span 2")
		sp2.SetTag("a name", "a value")

		time.Sleep(time.Millisecond * 100)

		sp3 := sp2.StartChild("span 3")

		time.Sleep(time.Millisecond * 100)

		sp4 := sp3.StartChild("span 4")

		time.Sleep(time.Millisecond * 400)

		sp4.Finish()

		sp3.Finish()

		sp2.Finish()
	}
}
```

## Expected Behavior
The first span operation name in the code is `span 1` but in the UI is showing as `span 4`.
Note that if you remove the `sentry.TransactionName("the transaction")` the `span 1` appears but of course the transaction name is `unamed transaction` which is no good when you need transaction grouping, so not really a workaround.

## Screenshots

Screen Shot 2021-01-13 at 4 31 45 PM

## Environment
sentry pkg at 0.9.0

### Sentry

rhcarvalho:
Otherwise the trace context of child spans or other unrelated value in
the scope ends up breaking the transaction representation.

Fixes #318

### DIFF
--- a/tracing_test.go
+++ b/tracing_test.go
@@ -159,7 +159,11 @@ func TestStartSpan(t *testing.T) {
 }
 
 func TestStartChild(t *testing.T) {
-	ctx := NewTestContext(ClientOptions{TracesSampleRate: 1.0})
+	transport := &TransportMock{}
+	ctx := NewTestContext(ClientOptions{
+		TracesSampleRate: 1.0,
+		Transport:        transport,
+	})
 	span := StartSpan(ctx, "top", TransactionName("Test Transaction"))
 	child := span.StartChild("child")
 	child.Finish()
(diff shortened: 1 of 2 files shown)


tracing_test.go
```diff
@@ -171,6 +175,54 @@ func TestStartChild(t *testing.T) {
...
+			"trace": &TraceContext{
+				TraceID: span.TraceID,
+				SpanID:  span.SpanID,
+				Op:      span.Op,
+			},
+		},
+		Tags: nil,
+		// TODO(tracing): Set Transaction.Data here or in
+		// Contexts.Trace, or somewhere else. Currently ignored.
+		Extra: nil,
```
kamilogorek, line 196: Same comment as in https://github.com/getsentry/sentry-go/pull/329/ once merged.

kamilogorek (dismissed):

tracing_test.go
rhcarvalho: We're not really exercising tags and extra here, I'll just remove it.

PR #328, opened 2021-02-13
fix: Do not double sample transactions

rhcarvalho:
Transactions should be sampled only by `TracesSampleRate` or `TracesSampler`, and the general `SampleRate` does not apply.

See https://develop.sentry.dev/sdk/performance/#sampling
image

### DIFF
--- a/client.go
+++ b/client.go
@@ -503,6 +503,9 @@ func (client *Client) processEvent(event *Event, hint *EventHint, scope EventMod
 		options.SampleRate = 1.0
 	}
 
+	// Transactions are sampled by options.TracesSampleRate or
+	// options.TracesSampler when they are started. All other events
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -503,7 +503,7 @@ func (client *Client) processEvent(event *Event, hint *EventHint, scope EventMod
 		options.SampleRate = 1.0
 	}
 
-	if !sample(options.SampleRate) {
+	if event.Type != transactionType && !sample(options.SampleRate) {
```
kamilogorek, line 506: Can we leave a comment here that transactions use their own sampler/samplerate?

kamilogorek (dismissed):

client.go
rhcarvalho: Done.

PR #329, opened 2021-02-13
fix: Send root span data as Event.Extra

### LINKED ISSUE #311: [0.9.0] Setting data & status on root span has no impact
## Summary
Hi, i've been using the new tracing system from the golang library, good job!
I've been unable to set data & level on a trace, though

## Steps To Reproduce

```go
span := sentry.StartSpan(ctx, name)
span.Status = sentry.SpanStatusInternalError

span.Data = map[string]interface{}{
"body":        c.Request().Body,
"path":        c.Path(),
"queryParams": c.QueryParams(),
"paramValues": c.ParamValues(),
}
span.Finish()
```

## Expected Behavior
See the data in the UI

### SDK
 - `sentry-go` version: 0.9.0
 - Go version: 1.15.0
 - Using Go Modules? yes

### Sentry
 - Using hosted Sentry in sentry.io? no
 - Using your own Sentry installation? Version: 20.12.1

rhcarvalho:
Pending spec clarification since implementation in Relay and different
SDKs diverged.

For now sending data in Event.Extra gets the information into the
Transaction view in Sentry.

Fixes #311.

### DIFF
--- a/tracing_test.go
+++ b/tracing_test.go
@@ -139,8 +139,9 @@ func TestStartSpan(t *testing.T) {
 			},
 		},
 		Tags: nil,
-		// TODO(tracing): Set Transaction.Data here or in
-		// Contexts.Trace, or somewhere else. Currently ignored.
+		// TODO(tracing): the root span / transaction data field is
+		// mapped into Event.Extra for now, pending spec clarification.
+		// https://github.com/getsentry/develop/issues/244#issuecomment-778694182
 		Extra:     span.Data,
 		Timestamp: endTime,
 		StartTime: startTime,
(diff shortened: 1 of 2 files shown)


kamilogorek (approved):

tracing_test.go
```diff
@@ -137,7 +141,7 @@ func TestStartSpan(t *testing.T) {
 		Tags: nil,
 		// TODO(tracing): Set Transaction.Data here or in
 		// Contexts.Trace, or somewhere else. Currently ignored.
-		Extra:     nil,
+		Extra:     span.Data,
```
kamilogorek, line 144: Does the comment above still stands if we set it now?

rhcarvalho: Thanks for noticing, it was stale. I moved the comment over from `tracing.go` here.

PR #343, opened 2021-05-17
ref(transports): Simplify retryAfter arguments

rhcarvalho:
Internal change.

We only need to look at HTTP headers, not the full response value.

### DIFF
--- a/transport_test.go
+++ b/transport_test.go
@@ -199,36 +199,29 @@ func TestGetRequestFromEvent(t *testing.T) {
 }
 
 func TestRetryAfterNoHeader(t *testing.T) {
-	r := http.Response{}
-	assertEqual(t, retryAfter(time.Now(), &r), time.Second*60)
+	assertEqual(t, retryAfter(time.Now(), nil), time.Second*60)
 }
 
 func TestRetryAfterIncorrectHeader(t *testing.T) {
-	r := http.Response{
-		Header: map[string][]string{
-			"Retry-After": {"x"},
-		},
+	h := http.Header{
+		"Retry-After": {"x"},
 	}
-	assertEqual(t, retryAfter(time.Now(), &r), time.Second*60)
+	assertEqual(t, retryAfter(time.Now(), h), time.Second*60)
 }
 
 func TestRetryAfterDelayHeader(t *testing.T) {
-	r := http.Response{
-		Header: map[string][]string{
-			"Retry-After": {"1337"},
-		},
+	h := http.Header{
+		"Retry-After": {"1337"},
 	}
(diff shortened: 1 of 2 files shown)


transport_test.go
```diff
@@ -199,36 +199,29 @@ func TestGetRequestFromEvent(t *testing.T) {
...
+	assertEqual(t, retryAfter(time.Now(), h), time.Second*1337)
 }
 
 func TestRetryAfterDateHeader(t *testing.T) {
 	now, _ := time.Parse(time.RFC1123, "Wed, 21 Oct 2015 07:28:00 GMT")
-	r := http.Response{
-		Header: map[string][]string{
-			"Retry-After": {"Wed, 21 Oct 2015 07:28:13 GMT"},
-		},
+	h := map[string][]string{
```
AbhiPrasad, line 221:
```suggestion
	h := http.Header{
```

transport_test.go
```diff
@@ -199,36 +199,29 @@ func TestGetRequestFromEvent(t *testing.T) {
...
-	assertEqual(t, retryAfter(time.Now(), &r), time.Second*60)
+	assertEqual(t, retryAfter(time.Now(), h), time.Second*60)
 }
 
 func TestRetryAfterDelayHeader(t *testing.T) {
-	r := http.Response{
-		Header: map[string][]string{
-			"Retry-After": {"1337"},
-		},
+	h := map[string][]string{
```
AbhiPrasad, line 213:
```suggestion
	h := http.Header{
```

transport_test.go
```diff
@@ -199,36 +199,29 @@ func TestGetRequestFromEvent(t *testing.T) {
...
-	assertEqual(t, retryAfter(time.Now(), &r), time.Second*60)
+	assertEqual(t, retryAfter(time.Now(), nil), time.Second*60)
 }
 
 func TestRetryAfterIncorrectHeader(t *testing.T) {
-	r := http.Response{
-		Header: map[string][]string{
-			"Retry-After": {"x"},
-		},
+	h := map[string][]string{
```
AbhiPrasad, line 206:
```suggestion
	h := http.Header{
```

AbhiPrasad (approved):
IMO, I think we should type as `http.Header` for better clarity, it's an alias for `map[string][]string` anyway.

transport_test.go
rhcarvalho: Absolutely. Updated those too, thanks!

rhcarvalho:
Rebased on top of master to pick up CI fixes from #345.

kamilogorek (approved):

PR #346, opened 2021-05-18
ci: Rewrite GitHub Action workflows

rhcarvalho:
- Break up Test and Lint into separate workflows (could be simply
separate jobs for parallelism, but separate workflows also allow
independent re-runs, independent triggers and independent concurrency
group)
- Use minimal permissions for GITHUB_TOKEN
- Cancel outdated jobs when PR/branch is updated
- Test on Linux/Windows/macOS
- Use actions/cache
- Stop testing against Go tip (was never a hard requirement and is
slower because requires compiling the toolchain)
- Stop testing against Go 1.13, start testing against Go 1.16 (because
we focus support on the last 3 releases of Go, more than the community
practice of supporting the last 2)
- Use the official golangci-lint action and the latest version of the
tool at the moment


.github/workflows/test.yml
```diff
@@ -0,0 +1,107 @@
...
+      GO111MODULE: "off"
+      WORKDIR: ${{ github.workspace }}/main/src/github.com/getsentry/sentry-go
+    defaults:
+      run:
+        working-directory: ${{ env.WORKDIR }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          path: ${{ env.WORKDIR }}
+      # TODO: cache dependencies
```
rhcarvalho, line 75: Didn't enable this step for now because it requires a bit more thought on the cache key or how to update dependencies.

In GOPATH Mode we download all dependencies in their latest commits. With too aggressive caching or missing an update step (a la `go get -u`)  we'd be testing against stale dependencies.

We can't use `go get -d -t -u -v ./...` because that would also try to update sentry-go itself. The solution probably requires resorting to `go list` and would need to work no matter if we got a cache hit or miss. Deferring that for later.

This job is obviously prone to breakage (dependencies are changing out of our control) and not going to be required to merge changes.

kamilogorek (approved):
There's nothing outstanding in the config, so I think it's fine as long as all workflows passes.

PR #348, opened 2021-05-19
Clone EventProcessors when cloning scope

ajpetersons:
When cloning a Scope (or a Hub), EventProcessors were not copied. This caused them not to be applied when using Sentry HTTP middleware.

This change copies EventProcessors from original Scope to the cloned Scope.

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -49,11 +49,12 @@ type Scope struct {
 // NewScope creates a new Scope.
 func NewScope() *Scope {
 	scope := Scope{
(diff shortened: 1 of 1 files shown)


rhcarvalho:
@ajpetersons thanks for the report/PR. Does #349 work for you? I think we can avoid some unnecessary allocations.

ajpetersons:
Yes, that seems to achieve what I'm looking for. Thanks!

PR #350, opened 2021-05-19
docs: Fix mentions to 'all events'

rhcarvalho:
When the examples and accompanying documentation were written, 'all events' meant 'error events'. 

With transactions added to the SDK, `BeforeSend` is only applied to error events while all `EventProcessor` functions (on client, scope or global) are applied to all outgoing events (errors and transactions).

### DIFF
--- a/example/with_extra/main.go
+++ b/example/with_extra/main.go
@@ -67,7 +67,7 @@ func main() {
 		Dsn:   "https://hello@world.io/1337",
 		BeforeSend: func(event *sentry.Event, hint *sentry.EventHint) *sentry.Event {
 			// Solution 1 (use beforeSend, which will be applied to
-			// all events and is usually application specific):
+			// error events and is usually application specific):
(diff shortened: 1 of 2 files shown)


example/with_extra/main.go
```diff
@@ -80,9 +80,9 @@ func main() {
 		},
 		Transport: &devNullTransport{},
 
-		// Solution 2 (use custom integration, which will also be
-		// applied to all events, but can be easily extracted even as a
-		// separate utility and reused across all your projects):
+		// Solution 2 (use custom integration, which will be
+		// applied to all events, and can be extracted as a
+		// separate utility and reused across projects):
```
AbhiPrasad, line 85: nit: Think this flows better

```suggestion
		// Solution 2 (use custom integration, which will be
		// applied to all events, can be extracted as a
		// separate utility, and reused across projects):
```

AbhiPrasad (approved):
Feel free to merge without addressing comment, just a nit. LGTM.

PR #363, opened 2021-05-31
feat: Automatic Release detection (#335)

### LINKED ISSUE #335: Automatic Release detection
## Summary

Make SDK pick up release name from known environment variables if not explicitly defined.

## Motivation

Having a release set is a required step to use Release Health and generally improves the product experience. Apart from taking the release name from `SENTRY_RELEASE`, look at other common environment variables defined in platforms where Go code is deployed. Look at Python/JS or other SDKs for inspiration.

## Additional Context

While there is no support yet for reporting sessions, this is a small change and useful by itself.

AramisHM:
Attempts to guess the release info, either by getting the last git
commit hash or by looking up some know environment variables that tracks
release identification.

Had to make changes on some tests, though I am not sure if what I wrote there is appropriated.

Closes #335.

### DIFF
--- a/http/sentryhttp_test.go
+++ b/http/sentryhttp_test.go
@@ -214,7 +214,7 @@ func TestIntegration(t *testing.T) {
 		cmpopts.IgnoreFields(
 			sentry.Event{},
 			"Contexts", "EventID", "Extra", "Platform",
-			"Sdk", "ServerName", "Tags", "Timestamp",
+			"Release", "Sdk", "ServerName", "Tags", "Timestamp",
 		),
 		cmpopts.IgnoreFields(
 			sentry.Request{},
--- a/util.go
+++ b/util.go
@@ -1,11 +1,14 @@
 package sentry
 
 import (
+	"bytes"
 	"crypto/rand"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"os"
+	"os/exec"
+	"strings"
 	"time"
 )
 
@@ -36,4 +39,44 @@ func monotonicTimeSince(start time.Time) (end time.Time) {
 func prettyPrint(data interface{}) {
 	dbg, _ := json.MarshalIndent(data, "", "  ")
 	fmt.Println(string(dbg))
+}
+
+// attempts to guess a default release.
+func defaultRelease() string {
+	// Search environment variables (EV) known to hold release info.
+	envs := []string{
+		"SENTRY_RELEASE",
+		"HEROKU_SLUG_COMMIT",
+		"SOURCE_VERSION",
+		"CODEBUILD_RESOLVED_SOURCE_VERSION",
+		"CIRCLE_SHA1",
+		"GAE_DEPLOYMENT_ID",
+		"GITHUB_SHA",             // GitHub Actions - https://help.github.com/en/actions
+		"COMMIT_REF",             // Netlify - https://docs.netlify.com/
+		"VERCEL_GIT_COMMIT_SHA",  // Vercel - https://vercel.com/
+		"ZEIT_GITHUB_COMMIT_SHA", // Zeit (now known as Vercel)
+		"ZEIT_GITLAB_COMMIT_SHA",
+		"ZEIT_BITBUCKET_COMMIT_SHA"}
+	for _, e := range envs {
+		if val := os.Getenv(e); val != "" {
+			return val // Stop at first non-empty variable.
+		}
+	}
+
+	// No EV's, attempt to get the last commit hash with git.
+	var stdout bytes.Buffer
+	cmd := exec.Command("git", "rev-parse", "HEAD")
+	cmd.Stdout = &stdout
+	err := cmd.Run()
+	if err != nil {
+		Logger.Println("Failed attempt to run git rev-parse.")
+	} else {
+		shastr := strings.TrimSpace(stdout.String())
+		if len(shastr) == 40 { // sha1 hash length
+			return shastr
+		}
+	}
+
+	// Not able to find a release name at all.
+	return ""
 }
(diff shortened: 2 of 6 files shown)


util.go
```diff
@@ -37,3 +40,44 @@ func prettyPrint(data interface{}) {
 	dbg, _ := json.MarshalIndent(data, "", "  ")
 	fmt.Println(string(dbg))
 }
+
+// GetDefaultRelease attempts to guess a default release.
+func GetDefaultRelease() string {
```
messer, line 45: If calling this when creating `Event`, wouldn't that mean that a service which suddenly generates a large number of events, will spin up `n` number of `git rev-parse HEAD` processes, making it kind of a fork bomb? If so, I feel like it might be better to setup some sort of situation where the call of this function is locked from sharing, and the result cached.

messer (commented):
If so inclined, please see the note I left above on the `GetDefaultRelease()` function declaration.

util.go
AramisHM: > If so inclined, please see the note I left on the `GetDefaultRelease()` function declaration.

Howdy!
Thanks for giving a look at this pr! Much appreciated!

When I read your comment, at first glance, I didn't think twice, this certainly would blow, like a fork bomb ! 😂

But, the function `GetDefaultRelease` is initialized on `Event`'s on test files only. I imagine this could be an exception for calling `GetDefaultRelease`?

I think that in a normal situation, `GetDefaultRelease` is called only once in `NewClient` on main.go.

"[...] some sort of situation where the call of this function is locked from sharing, [...]"
I am not sure if I understood, by locking, you mean not exporting `GetDefaultRelease`? I didn't export at first, but wasn't able to call this on the test cases, since they are in different packages.

Feel free to point anything else you find strange in this pr!

messer: > Howdy!
> Thanks for giving a look at this pr! Much appreciated!

You're very welcome.

> I think that in a normal situation, `GetDefaultRelease` is called only once in `NewClient` on main.go.

You're right, I'm not sure where my mind was with that. I retract.

rhcarvalho: Happy to see the positive engagement here :)

- [x] Re: how often this is called;

  I think we are fine for now. `GetDefaultRelease` is called once per `NewClient` call, and we call `NewClient` once when the SDK is initialized with `Init`, which should be only once in the lifetime of a program. If this ever becomes a problem we can cache the result, but we don't need that to land this PR.

- [ ] Re: exporting this name;

  Please let's not export it. It is easy enough to export it later, it is hard to remove something later. SDK users don't really need to call this function themselves. We can deal with tests, see other comment.

http/sentryhttp_test.go
```diff
@@ -33,6 +33,7 @@ func TestIntegration(t *testing.T) {
 			}),
 
 			WantEvent: &sentry.Event{
+				Release: sentry.GetDefaultRelease(),
```
rhcarvalho, line 36: For the tests that live outside of `package sentry`, we don't really need to test the behavior of getting a default release string.

```suggestion
```

We can remove again the `Release` field from the tests, and explicitly ignore it in the [comparison options down below](https://github.com/getsentry/sentry-go/pull/363/files#diff-f5c2234d1ccf4d9b66809ca2e4c7a9e1a29b0e3742b110222ab843a896359fa0R219-R223).

```go
cmpopts.IgnoreFields(
	sentry.Event{},
	"Contexts", "EventID", "Extra", "Platform", "Release",       // <--- added "Release" here
	"Sdk", "ServerName", "Tags", "Timestamp",
),
```

util.go
```diff
@@ -37,3 +40,44 @@ func prettyPrint(data interface{}) {
...
+	if release := os.Getenv("SENTRY_RELEASE"); release == "" {
```
rhcarvalho, line 46: I think we can reduce the amount of indented code here by reversing the condition:

```go
if release := os.Getenv("SENTRY_RELEASE"); release != "" {
	return release
}

// continue trying to figure out a release name...
```

util.go
```diff
@@ -37,3 +40,44 @@ func prettyPrint(data interface{}) {
...
+				"COMMIT_REF",             // Netlify - https://docs.netlify.com/
+				"VERCEL_GIT_COMMIT_SHA",  // Vercel - https://vercel.com/
+				"ZEIT_GITHUB_COMMIT_SHA", // Zeit (now known as Vercel)
+				"ZEIT_GITLAB_COMMIT_SHA",
+				"ZEIT_BITBUCKET_COMMIT_SHA"}
+			for _, env := range envVars {
+				if envVal := os.Getenv(env); envVal != "" {
+					return envVal // Stop at first non-empty variable.
+				}
+			}
```
rhcarvalho, line 79: I think it might be a better end result to check env vars first, and only as a last resort look at VCS/git, because env vars might give a more informative version name than `git`.

@AramisHM, what do you think?

If we pursue that route, this whole function could become shorter too, along the lines of:

```go
func defaultRelease() string {
	s := []string{"SENTRY_RELEASE", "GITHUB_SHA", ...}
	for _, e := range s {
		if v := os.Getenv(e); v != "" {
			return v
		}
	}
	// ... no env var had a suitable release name, try git...
	// ...

	// not able to find a release name at all
        return ""
}

rhcarvalho (changes requested):
@AramisHM thanks for your contribution, muito obrigado! Only required change is to not introduce any new exported symbol. We can always revisit how exactly the automatic detection works later.

util.go
AramisHM: I took the approach from your next note and discarded this block entirely

AramisHM:
Thank you for taking the time to deal with a newbie.

Wasn't sure if create a new commit, ended up force pushing the new code on top of the old one.

rhcarvalho (approved):
Thanks @AramisHM!

PR #366, opened 2021-06-22
More ClientOptions.Release documentation

rhcarvalho:
- Briefly explain importance of setting a release
- Document that release may be derived from env or git
- Recommend setting the release during build
- Apply build-time release pattern to examples

This is a follow up on https://github.com/getsentry/sentry-go/pull/363.

### DIFF
--- a/example/basic/main.go
+++ b/example/basic/main.go
@@ -22,6 +22,13 @@ import (
 	"github.com/getsentry/sentry-go"
 )
 
+// release is the release of this program that will be reported to Sentry.
+// Use go build -ldflags='-X main.release=VALUE' to set this value in build
+// time.
+// If not set, a default release value will be derived in runtime from
+// environment variables or the Git repository in the current working directory.
+var release string
+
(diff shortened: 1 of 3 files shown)


example/basic/main.go
```diff
@@ -22,6 +22,13 @@ import (
 	"github.com/getsentry/sentry-go"
 )
 
+// release is the release of this program that will be reported to Sentry.
+// Use go build -ldflags='-X main.release=VALUE' to set this value in build
+// time.
+// If not set, a default release value will be derived in runtime from
+// environment variables or the Git repository in the current working directory.
+var release string
```
rhcarvalho, line 30: If we're happy with this suggestion, I'd add the same to other examples, and a similar bit of explanation to https://docs.sentry.io.

AbhiPrasad (approved):
I like the added context, and agree that we should update the docs site.

kamilogorek (approved):

PR #369, opened 2021-07-19
feat: Add debug logging for release detection

rhcarvalho:
This enables anyone to understand better how the SDK decided to report a
given Release value.

Also:
- Simplify code to use exec.Cmd.Output instead of exec.Cmd.Run with a
  manually created bytes.Buffer
- Remove check of commit hash length returned from Git -- we can use any
  length for Release.

### DIFF
--- a/util.go
+++ b/util.go
@@ -1,7 +1,6 @@
 package sentry
 
 import (
-	"bytes"
 	"crypto/rand"
 	"encoding/hex"
 	"encoding/json"
@@ -43,8 +42,8 @@ func prettyPrint(data interface{}) {
 
 // defaultRelease attempts to guess a default release for the currently running
 // program.
-func defaultRelease() string {
-	// Search environment variables (EV) known to hold release info.
+func defaultRelease() (release string) {
+	// Return first non-empty environment variable known to hold release info, if any.
 	envs := []string{
 		"SENTRY_RELEASE",
 		"HEROKU_SLUG_COMMIT",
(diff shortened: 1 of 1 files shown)


kamilogorek (approved):

util.go
```diff
@@ -57,27 +56,32 @@ func defaultRelease() string {
...
+		Logger.Print(s.String())
+		Logger.Print("Some Sentry features will not be available. See https://docs.sentry.io/product/releases/.")
+		Logger.Print("To stop seeing this message, pass a Release to sentry.Init or set the SENTRY_RELEASE environment variable.")
+		return ""
 	}
-
-	// Not able to find a release name at all.
-	return ""
+	release = strings.TrimSpace(string(b))
+	Logger.Printf("Release from Git: %s", release)
```
AbhiPrasad, line 85:
```suggestion
	Logger.Printf("Using release from Git: %s", release)
```

util.go
```diff
@@ -57,27 +56,32 @@ func defaultRelease() string {
...
 		"ZEIT_GITHUB_COMMIT_SHA", // Zeit (now known as Vercel)
 		"ZEIT_GITLAB_COMMIT_SHA",
-		"ZEIT_BITBUCKET_COMMIT_SHA"}
+		"ZEIT_BITBUCKET_COMMIT_SHA",
+	}
 	for _, e := range envs {
-		if val := os.Getenv(e); val != "" {
-			return val // Stop at first non-empty variable.
+		if release = os.Getenv(e); release != "" {
+			Logger.Printf("Release from environment variable %s: %s", e, release)
```
AbhiPrasad, line 63:
```suggestion
			Logger.Printf("Using release from environment variable %s: %s", e, release)
```

AbhiPrasad (approved):

util.go
rhcarvalho: I was purposely trying to be terse with the messages. Is the current form understandable enough?

AbhiPrasad: I feel the use of the action word makes the intent a lot more clear so I prefer having `Using` + I don’t think adding an extra word increases length or complexity by too much.

PR #377, opened 2021-09-01
Add Package to the frame when creating stacktrace

grongor:
When creating stack trace, the Package field is completely omitted, so that information isn't present in the Sentry UI withing Exception box.

I tried simpler approach first: to just fill Package field with the same content as the Module field, and the result is identical (I guess Sentry server/frontend does the same thing I did here with stripping content before last slash), but this way we send smaller payload...I'm not sure which you prefer but I just wanted to mention this fact so you can decide on your own.

Here are before and after screenshots from my dummy application:

before
after


grongor:
> @grongor in case you want to have this and other customizations to your events you can write an [EventProcessor](https://pkg.go.dev/github.com/getsentry/sentry-go#EventProcessor) that walks event stack traces and set `Frame.Package` (and in this case you can always derive it from `Frame.Module`).

Thanks for the suggestion. Yeah, I know there are ways (it's great that it's so extensible and people are not limited by the SDK :heart: ), I just thought that the "default" should be to include it :) I'll do just that for my projects (as I do for other fixes like for https://github.com/getsentry/sentry-go/issues/378 ).

PR #389, opened 2021-10-31
fix!: Separate sentryiris into its own module.

elliotcourant:
This is a follow up on discussion https://github.com/getsentry/sentry-go/discussions/388

This separates the sentry code for iris into its own go module. Allowing it to be imported directly; and preventing side effects from importing sentry-go on its own.


rhcarvalho:
Closing as per https://github.com/getsentry/sentry-go/discussions/388#discussioncomment-1583213.

Splitting into multiple modules require considerations involving the release process, GitHub Actions setup, developer environment, etc. Putting that infrastructure together is the hard part.

We are aware of the problems the single-module layout brings to downstream consumers, and also aware this is not limited to `sentry-go` but a common difficulty in the ecosystem.

For most concerns we have solutions with Go 1.17. Perhaps improving dependabot notifications is the next big thing.
The `go.work` workspaces can also help with multi-module repositories and I'm closely tracking that too.

Issue #391, opened 2021-12-02
Please support StackTraces which return a slice of `runtime.Frame`


kat-co:
I am working with an errors package that reports a stack trace as a slice of `runtime.Frame`. Sentry does not look for the `PC` [field](https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/runtime/symtab.go;l=31) which `runtime.Frame` contains:

https://github.com/getsentry/sentry-go/blob/bd116d6ce79b604297c6497aa07d7ac01768adbb/stacktrace.go#L123-L127

This makes utilizing `sentry.ReportException` less useful in this case since it will only generate a new stacktrace from the place `sentry.ReportException` was called.

Thank you!

kat-co:
Apologies, I need to rebase this off of the last release. I'm closing this PR in favor of another.

rhcarvalho:
Hi @kat-co, is the error package you're working with publicly available?

kat-co:
Sorry, @rhcarvalho, it is not. I think the salient bit is that our error type satisfies this interface:
```go
type MyError interface {
  StackTrace() []runtime.Frame
}
```

If it helps, I can tell you that a fork of `sentry-go` with the addition of my commit has been running in production for a few weeks now and is working well.

PR #392, opened 2021-12-02
fix: When extracting PCs from stack frames, try the `PC` field

kat-co:
Some error types may return stack traces already defined as `[]runtime.Frame`.
In this case, the program counter will be in the `PC` field.

fixes: #391


kat-co:
Apologies, I need to rebase this off of the last release. I'm closing this PR in favor of another.

PR #393, opened 2021-12-02
fix: When extracting PCs from stack frames, try the `PC` field

kat-co:
Some error types may return stack traces already defined as `[]runtime.Frame`.
In this case, the program counter will be in the `PC` field.

fixes: #391


kamilogorek (approved):

kamilogorek:
Thanks! The change looks trivial, so there should be no issues getting it out there, especially that iteration order preserves `ProgramCounter` as the initial check.

PR #408, opened 2022-01-13
fix: marshal exceptions into an object with a values node, as per spec

bastianccm:
This change sets the `exception` property of an event to an object with a `values` property, as described in the docs: https://develop.sentry.dev/sdk/event-payloads/types/#exception


rhcarvalho:
Thanks for the additional context.

I've sent https://github.com/getsentry/develop/pull/491 to update the dev docs.

PR #410, opened 2022-02-08
build: Bump Go version in go.mod

rhcarvalho:
Commit generated with this command

```
go mod tidy -go=1.17 -compat=1.15
```

Using Go version 1.17.5.

The idea is to enable consumers of sentry-go to be able to use module
graph pruning (Go 1.17+ required), such that dependencies of sentry-go
that are otherwise unused by end users can be pruned away.

See https://github.com/getsentry/sentry-go/issues/156#issuecomment-1027175141

Also update CI workflow to check for go.mod tidiness using the same line
as used to generate this commit. Update data race tests that were still
accidentally being run on Go 1.16 instead of 1.17.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,8 +46,8 @@ jobs:
       - name: Vet
         run: go vet ./...
       - name: Check go.mod Tidiness
-        run: go mod tidy && git diff --exit-code
-        if: ${{ matrix.go == '1.16' }}
+        run: go mod tidy -go=1.17 -compat=1.15 && git diff --exit-code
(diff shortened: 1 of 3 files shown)


rhcarvalho:
The CodeQL failure can likely be ignored. It should have been addressed by https://github.com/github/codeql-go/pull/658 (which closed https://github.com/github/codeql-go/issues/650).

@kamilogorek do you know who's taking a look here these days?

kamilogorek:
> such that dependencies of sentry-go that are otherwise unused by end users can be pruned away.

At last 😌

> @kamilogorek do you know who's taking a look here these days?

Noone just yet 😔

PR #434, opened 2022-04-27
feat: :sparkles: Added function to continue from trace string

karatekaneen:
**tl;dr: This allows users to easily add traces from sources other than http.Request**

Recently ran in to a situation where we want to add tracing between services where not all of them uses HTTP to communicate (in this case, rabbitMQ) so I refactored a bit and added a new exposed function that continues from a trace header.

This does not introduce any breaking changes from what I can see and the tests does of course still pass.

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -545,9 +545,18 @@ func TransactionName(name string) SpanOption {
 // ContinueFromRequest returns a span option that updates the span to continue
 // an existing trace. If it cannot detect an existing trace in the request, the
 // span will be left unchanged.
+//
+// ContinueFromRequest is an alias for:
+// 	ContinueFromTrace(r.Header.Get("sentry-trace"))
 func ContinueFromRequest(r *http.Request) SpanOption {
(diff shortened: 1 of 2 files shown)


tracing.go
```diff
@@ -545,9 +545,18 @@ func TransactionName(name string) SpanOption {
...
+//
+// ContinueFromRequest is an alias for:
+// 	ContinueFromTrace(r.Header.Get("sentry-trace"))
 func ContinueFromRequest(r *http.Request) SpanOption {
+	return ContinueFromTrace(r.Header.Get("sentry-trace"))
+}
+
+// ContinueFromTrace returns a span option that updates the span to continue
+// an existing trace. If it cannot detect an existing trace in the request, the
+// span will be left unchanged.
```
kamilogorek, line 557:
```suggestion
// ContinueFromTrace returns a span option that updates the span to continue
// an existing TraceID.
```

kamilogorek:
Lovely, thanks!

PR #443, opened 2022-05-20
Tweak readme, assume new users are using go modules

jeffwidman:
Anyone reading the readme is likely using go modules. 
Plus the last three versions of go are supported, which means `1.15` is the oldest supported version, so 99% of users on go 1.15 will be using go modules.

Also replaced the word `"line"` with `"generation"` because at first glance I was confused on how the next line in the file... took a moment to realize this is the next iteration or generation.


jeffwidman:
Gentle nudge

jeffwidman:
Nudge @bruno-garcia

PR #444, opened 2022-05-25
ref: Use a Context type mapping to map[string]interface{}

phacops:
When Sentry processes an event, it only accepts a context being a key/value object. This PR aims to enforce we're passing a `map[string]interface{}` instead of an `interface{}`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- ref: Use a Context type mapping to a map[string]interface{} for all contexts
+
 ## v0.13.0
 
 - ref: Change DSN ProjectID to be a string (#420)
--- a/interfaces.go
+++ b/interfaces.go
@@ -161,10 +161,12 @@ type Exception struct {
 // An EventID must be 32 characters long, lowercase and not have any dashes.
 type EventID string
 
+type Context map[string]interface{}
+
 // Event is the fundamental data structure that is sent to Sentry.
 type Event struct {
 	Breadcrumbs []*Breadcrumb          `json:"breadcrumbs,omitempty"`
-	Contexts    map[string]interface{} `json:"contexts,omitempty"`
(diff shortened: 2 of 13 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,9 @@
+repos:
+- repo: https://github.com/dnephin/pre-commit-golang
+  rev: 96221dc741cb30cc0136999083dc6bd0e2113000
+  hooks:
+    - id: go-build
+    - id: go-fmt
+    - id: go-imports
+    - id: golangci-lint
+    - id: no-go-testing
```
kamilogorek, line 9: Can you move this to a separate PR?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- ref: Use a Context type mapping to a map[string]interface{} for all contexts
```
kamilogorek, line 5:
```suggestion
- ref: Use a `Context` type mapping to a `map[string]interface{}` for all event contexts (#444)
```

interfaces.go
```diff
@@ -161,10 +161,12 @@ type Exception struct {
 // An EventID must be 32 characters long, lowercase and not have any dashes.
 type EventID string
 
+type Context map[string]interface{}
```
kamilogorek, line 164: I'm not sure if we should introduce a concrete `Context` type. It'll force people to pass `sentry.Context` explicitly. Where if we'd use type alias, `type Context = map[string]interface{}`, it'd still work with most cases where people already pass a valid dictionary.

kamilogorek (changes requested):

interfaces.go
phacops: I defined a new type since you mention on Slack it was a v0 and we should strive to make the API as it should. The type alias works too and the difference is minimal so that'll works well too.

PR #446, opened 2022-06-01
fix: Scope values should not override Event values

### LINKED ISSUE #326: Scope values override Event values
## Summary

In the Go SDK, setting certain fields on the scope override the equivalent fields in events, while in other SDKs the event fields take precedence.

## Steps To Reproduce

Compare [Scope.ApplyToEvent (Go)](https://github.com/getsentry/sentry-go/blob/aed9115503d3bc31fd9be452423831c2bc7d19c7/scope.go#L339-L425) and [Scope.applyToEvent (JS)](https://github.com/getsentry/sentry-javascript/blob/1fb0b4c3f6533c43ead2d2cb3f0dc4bc8a84e544/packages/hub/src/scope.ts#L372-L408).

In particular:

### Go

```go
	if len(scope.contexts) > 0 {
		if event.Contexts == nil {
			event.Contexts = make(map[string]interface{})
		}


		for key, value := range scope.contexts {
			event.Contexts[key] = value
		}
	}
```

### JavaScript

```javascript
    if (this._contexts && Object.keys(this._contexts).length) {
      event.contexts = { ...this._contexts, ...event.contexts };
    }
```

## Expected Behavior

SDKs should handle this type of detail consistently.

### SDK
 - `sentry-go` version: `0.9.0`
 - Go version: any
 - Using Go Modules? [yes/no] either

## Additional context

[The spec](https://develop.sentry.dev/sdk/unified-api#scope) is not clear about precedence, but seems that the JavaScript behavior is the one replicated elsewhere.

craigpastro:
Fixes #326.


cleptric (approved):
LGTM

AbhiPrasad (approved):

PR #447, opened 2022-06-01
style: Deprecate "exception" from the API and replace with "error"

### LINKED ISSUE #161: Remove "Exception" from the API
I propose deprecating and removing the word "exception" from methods in the exported API.
Given that there are no exceptions in Go, it is alien to the language to write `sentry.CaptureException(someError)`.

Note that `raven-go` used `sentry.CaptureError` which sounds more idiomatic.

Note also that there is an `Exception` type, but that one matches the Server's Payload API https://docs.sentry.io/development/sdk-dev/event-payloads/exception/.

`Exception` is part of an `Event`. We may talk about all of these in terms of "error" in Go, and only have "exception" in the serialized JSON payload, as that's what the server understand and what we use for reporting.

craigpastro:
Closes #161.

Deprecates methods with the word "exception" from the exported API and adds the same method with the word "error".

There are still some private methods with "exception" and exported structs with "exception" fields.


cleptric:
When the event gets sent to Sentry, we still have to use the exception key.

Btw, our Rust SDK uses [sentry::capture_error(&err)](https://docs.sentry.io/platforms/rust/usage/#capturing-errors) as well, so I'm 👍 on this and will take a look!

PR #450, opened 2022-06-22
Add StripSpanContextKeyFromContext function

GTB3NW:
StripSpanContextKeyFromContext returns a new context with the parent span removed
but critically retaining the parentSpanID, allowing for new transactions
to be formed within the same running process while retaining all the current
context benefits, including existing sentry context keys (apart  from
`spanContextKey`) and timeouts etc.


GTB3NW:
I can implement the logic needed without this PR. This currently isn't useful

PR #451, opened 2022-06-22
Export `spanFromContext`

GTB3NW:
Matches the logic from `TransactionFromContext` and returns nil.
This is idiomatic and does what it says on the tin, if the context
does not have a span contained within it, I would not expect anything
to be returned. An error seems... extra.


GTB3NW:
Opened a new one to switch branch. Sorry for the PR spam - https://github.com/getsentry/sentry-go/pull/452

PR #452, opened 2022-06-22
Export `spanFromContext`

GTB3NW:
Matches the logic from `TransactionFromContext` and returns nil.
This is idiomatic and does what it says on the tin, if the context
does not have a span contained within it, I would not expect anything
to be returned. An error seems... extra.

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -600,27 +600,9 @@ func TransactionFromContext(ctx context.Context) *Span {
 	return nil
 }
 
-// spanFromContext returns the last span stored in the context or a dummy
+// SpanFromContext returns the last span stored in the context or a dummy
 // non-nil span.
-//
-// TODO(tracing): consider exporting this. Without this, users cannot retrieve a
-// span from a context since spanContextKey is not exported.
-//
-// This can be added retroactively, and in the meantime think better whether it
-// should return nil (like GetHubFromContext), always non-nil (like
-// HubFromContext), or both: two exported functions.
-//
-// Note the equivalence:
-//
-// 	SpanFromContext(ctx).StartChild(...) === StartSpan(ctx, ...)
-//
-// So we don't aim spanFromContext at creating spans, but mutating existing
-// spans that you'd have no access otherwise (because it was created in code you
-// do not control, for example SDK auto-instrumentation).
-//
-// For now we provide TransactionFromContext, which solves the more common case
-// of setting tags, etc, on the current transaction.
-func spanFromContext(ctx context.Context) *Span {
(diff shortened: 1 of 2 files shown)


rhcarvalho:
> cc: @rhcarvalho - Not sure if you work for sentry anymore; If you do please could you weigh in on this? The function was originally created by you an annotated with some ideas for future plans.

Hi @GTB3NW! What are you going to use `SpanFromContext` for? You can't do much with spans without a parent transaction, so I believe `TransactionFromContext` is still the right choice for most cases.

rhcarvalho:
Would be interesting to see more details on the PoC :-)

> If I want to create a new transaction linked to an existing one and as a child of a span it's impossible without exposing the trace id **&** most importantly the span id, this allows me to do it in one fell swoop.

Trace propagation is done with a `sentry-trace` string / HTTP header, and includes both trace and span IDs, so exposing those is expected. If you want to hide that info within a `context.Context`, you can do so using your own context key and then fetch it and start a new transaction using the recently added `ContinueFromTrace` (https://github.com/getsentry/sentry-go/pull/434).

> As you noted in the todo there's other use cases such as modifying instrumented spans from middleware etc.

That's right. The "transaction" concept is however so central to Sentry's tracing such that without an associated transaction all span data is garbage collected. In this model, user code can always poke at the spans starting off a transaction.
So some form of "middleware" calling `TransactionFromContext` to mutate the transaction/spans would be essentially the same as setting up an event processor to go over transactions before they are sent to Sentry.

GTB3NW:
> Would be interesting to see more details on the PoC :-)

Just packaging it up now



> Trace propagation is done with a `sentry-trace` string / HTTP header, and includes both trace and span IDs, so exposing those is expected. If you want to hide that info within a `context.Context`, you can do so using your own context key and then fetch it and start a new transaction using the recently added `ContinueFromTrace` (#434).

Yeah I saw that function but without any way to the data required for the function it's useless. I wouldn't mind adding a function to extract that data from a context as you say. I think my next PR may be a bit more controversial than that change :P



> That's right. The "transaction" concept is however so central to Sentry's tracing such that without an associated transaction all span data is garbage collected. In this model, user code can always poke at the spans starting off a transaction.
> So some form of "middleware" calling `TransactionFromContext` to mutate the transaction/spans would be essentially the same as setting up an event processor to go over transactions before they are sent to Sentry.

I'll be honest I'm less bothered about that use case, I personally wouldn't be a user of it, however I am working on some middleware which does need the span/traceid, so less about mutation and more about information :)

PR #454, opened 2022-06-26
Replace deprecated ioutil pkg with os & io

abhinavnair:
As of Go 1.16, the same functionality is now provided by package io or
package os, and those implementations should be preferred in new code.

So replacing all usage of `ioutil` pkg with `io` & `os`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"log"
 	"math/rand"
 	"net/http"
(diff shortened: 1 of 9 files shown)


YaroslavPodorvanov (approved):

YaroslavPodorvanov:
@kamilogorek this PR safe

cleptric (commented):
@abhinavnair Would you be so kind as to update the Changelog.md and include a note similar to https://github.com/getsentry/sentry-go/blob/master/CHANGELOG.md#v0120  letting folks know that versions before 1.16 won't work anymore?

kamilogorek (approved):

kamilogorek:
LGTM. @cleptric we will update support note once https://github.com/getsentry/sentry-go/pull/470 is merged, just prior to releasing next version.

smeubank:
i was just looped in here, and possibly I am being a bit overcautious 😅  . And I'm no expert in Go, we just see some significant number of events coming in for sub 1.16 Go versions. So I am wondering if this breaking change can be based on some version dependency check from the SDK?

smeubank:
any thoughts on my post above? @kamilogorek maybe we just need to make a decision here on what to do

vladanpaunovic:
Together with @smeubank , we'll follow up on this, respecting our depreciation policy.

kamilogorek (approved):

PR #459, opened 2022-07-08
Add configurable stack frame offset

### LINKED ISSUE #458: Allow customizable stack trace offset when using abstracted interface
## Summary
It's very likely that as an application grows, developers will seek to abstract some foundational things they may need such as logging, etc. And additionally, manually capturing Sentry breadcrumbs or events would likely get abstracted as well. If an error has a stacktrace already, great. Otherwise, using `sentry.NewStacktrack()` is the way to go. But this hard-codes the offset to `1`. 

I'd like it if we can configure the offset in the pcs from `runtime.Callers` so that the trace starts where the problem was, and not in the logger or other abstraction.

## Motivation
I already have a logger based on Zap and want to add Sentry integration. My logger is the first few frames of the stack trace which is clutter.

mbrancato:
This adds a new exported function to return a `sentry.Stacktrace` with a customized stack frame offset.
Fixes #458

### DIFF
--- a/stacktrace.go
+++ b/stacktrace.go
@@ -25,8 +25,18 @@ type Stacktrace struct {
 
 // NewStacktrace creates a stacktrace using runtime.Callers.
 func NewStacktrace() *Stacktrace {
+	return newStacktrace(2)
+}
+
+// NewCustomStacktrace creates a stacktrace using runtime.Callers while skipping the specified
+// number of stack frames.
+func NewCustomStacktrace(skip int) *Stacktrace {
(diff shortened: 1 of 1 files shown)


tmc:
It'd be useful to demonstrate use of this capability.
there's also a chance to add this to the more general interfaces.

YaroslavPodorvanov:
similar skip in previous PR https://github.com/getsentry/sentry-go/pull/396

smeubank:
Yes, is the goal here to influence grouping for example?

PR #460, opened 2022-07-19
Make maximum amount of spans configurable

fsrv-xyz:
In one of my projects, I need to do performance monitoring for complex tasks with the highest possible accuracy.
Due to some of the tasks are long running, the current `const` span limit of 1000 per transaction will be not enough.

I saw, that some other SDKs (for example python) have an option for that.


YaroslavPodorvanov:
I think exists config on server too and user can not change config on server

fsrv-xyz:
I have tested on one of our self hosted instances. Works fine with a higher limit (as far as I can check).

cleptric:
@fsrv-xyz Can you add some tests for this so we get it merged?

fsrv-xyz:
@cleptric Sure. I added some test cases covering the added code (and some additional things).

span_recorder.go
```diff
@@ -20,6 +15,10 @@ type spanRecorder struct {
 // record stores a span. The first stored span is assumed to be the root of a
 // span tree.
 func (r *spanRecorder) record(s *Span) {
+	maxSpans := defaultMaxSpans
+	if CurrentHub().Client() != nil {
+		maxSpans = CurrentHub().Client().options.MaxSpans
```
cleptric, line 20:
```suggestion
	if client := CurrentHub().Client(); client != nil {
		maxSpans = client.Options().MaxSpans
```

client.go
```diff
@@ -175,6 +180,8 @@ type ClientOptions struct {
 	Environment string
 	// Maximum number of breadcrumbs.
 	MaxBreadcrumbs int
+	// Maximum number of spans.
+	MaxSpans int
```
cleptric, line 184:
```suggestion
	// Maximum number of spans.
	//
	// See https://develop.sentry.dev/sdk/envelopes/#size-limits for size limits
	// applied during Event ingestion. Events that exceed these limits might get dropped.
	MaxSpans int
```

cleptric (commented):
Please update the `CHANGELOG.md` as well and add something like 

> feat: Make maximum amount of spans configurable (#460)

Thanks for your work on this 🙂

span_recorder_test.go
```diff
@@ -0,0 +1,63 @@
...
+			Logger.SetOutput(&logBuffer)
+			defer Logger.SetOutput(io.Discard)
+			spanRecorder := spanRecorder{}
+
+			currentHub.BindClient(&Client{
+				options: ClientOptions{
+					MaxSpans: tt.maxSpans,
+				},
+			})
+			// unbind client after test for not affecting other tests
```
cleptric, line 44:
```suggestion
			// Unbind the client afterwards, to not affect other tests
```

fsrv-xyz:
I applied the changes you suggested and updated the changelog file.

cleptric (approved):

PR #469, opened 2022-08-31
Cleanup for more readable + benchmark ExtractStacktrace

YaroslavPodorvanov:
In performance nothing change
```bash
benchstat old.txt new.txt
```
```text
name                  old time/op    new time/op    delta
ExtractStacktrace-12    6.73µs ±17%    6.59µs ±16%   ~     (p=0.057 n=98+98)

name                  old alloc/op   new alloc/op   delta
ExtractStacktrace-12    2.56kB ± 0%    2.56kB ± 0%   ~     (all equal)

name                  old allocs/op  new allocs/op  delta
ExtractStacktrace-12      26.0 ± 0%      26.0 ± 0%   ~     (all equal)
```


stacktrace.go
```diff
@@ -73,39 +73,38 @@ func ExtractStacktrace(err error) *Stacktrace {
 }
 
 func extractReflectedStacktraceMethod(err error) reflect.Value {
-	var method reflect.Value
+	errValue := reflect.ValueOf(err)
```
YaroslavPodorvanov, line 76: reflect.ValueOf once instead of multi

stacktrace_external_test.go
```diff
@@ -255,6 +255,21 @@ func TestExtractStacktrace(t *testing.T) {
 	}
 }
 
+func BenchmarkExtractStacktrace(b *testing.B) {
+	var fs = []func() error{
+		RedPkgErrorsRanger,
```
YaroslavPodorvanov, line 260: same as in tests

kamilogorek (approved):
Thanks 🙇

PR #471, opened 2022-09-06
Add logrus support

### LINKED ISSUE #43: Support logrus out-of-the-box?
Logrus is likely the most popular logging library in Go (https://github.com/sirupsen/logrus), with over 10k stars.

Since sentry-go already supports some web frameworks, it will be great if there can be out-of-the-box support for logrus too.

flimzy:
This PR adapts my free-standing logrus hook (https://gitlab.com/flimzy/logrusentry) for inclusion in this repo. Closes #43 

There's a good chance this hook doesn't take advantage of all available/relevant Sentry features.  But it's been serving me well for several years now, at at least three different companies.  Please advise if there are any functional or stylistic changes that are desired. I'm happy to do what I can to get this included in this repo.

### DIFF
--- a/logrus/logrusentry.go
+++ b/logrus/logrusentry.go
@@ -1,0 +1,240 @@
+// Package logrusentry provides a simple Logrus hook for Sentry.
+package logrusentry
+
+import (
+	"errors"
+	"net/http"
+	"time"
+
+	sentry "github.com/getsentry/sentry-go"
+	"github.com/sirupsen/logrus"
+)
+
+// These default log field keys are used to pass specific metadata in a way that
+// Sentry understands. If they are found in the log fields, and the value is of
+// the expected datatype, it will be converted from a generic field, into Sentry
+// metadata.
+//
+// These keys may be overridden by calling SetKey on the hook object.
+const (
+	// FieldRequest holds an *http.Request.
+	FieldRequest = "request"
+	// FieldUser holds a User or *User value.
+	FieldUser = "user"
+	// FieldTransaction holds a transaction ID as a string.
+	FieldTransaction = "transaction"
+	// FieldFingerprint holds a string slice ([]string), used to dictate the
+	// deduplication of this event
+	FieldFingerprint = "fingerprint"
+
+	// These fields are simply omitted, as they are duplicated by the Sentry SDK.
+	FieldGoVersion = "go_version"
+	FieldMaxProcs  = "go_maxprocs"
+)
+
+// ClientOptions is an alias for sentry.ClientOptions, and is all of the options
+// available for configuring the sentry SDK client.
+type ClientOptions = sentry.ClientOptions
+
+// User is an alias for sentry.User. Pass a this value with key FieldUser, and
+// it will be passed to Sentry appropriately.
+type User = sentry.User
+
+// Event is an alias for sentry.Event.
+type Event = sentry.Event
+
+// EventHint is an alias for sentry.EventHint
+type EventHint = sentry.EventHint
+
+// Hook is the logrus hook for Sentry.
+//
+// It is not safe to configure the hook while logging is happening. Please
+// perform all configuration before using it.
+type Hook struct {
+	client   *sentry.Client
+	scope    *sentry.Scope
+	hub      *sentry.Hub
+	fallback FallbackFunc
+	keys     map[string]string
+	levels   []logrus.Level
+}
+
+var _ logrus.Hook = &Hook{}
+
+// New initializes a new Sentry hook.
+func New(levels []logrus.Level, opts ClientOptions) (*Hook, error) {
+	client, err := sentry.NewClient(opts)
+	if err != nil {
+		return nil, err
+	}
+	h := &Hook{
+		levels: levels,
+		client: client,
+		scope:  sentry.NewScope(),
+		keys:   make(map[string]string),
+	}
+	h.setHub()
+	return h, nil
+}
+
+// setHub refreshes the hub, to be used any time client or scope changes.
+func (h *Hook) setHub() {
+	h.hub = sentry.NewHub(h.client, h.scope)
+}
+
+// AddTags adds tags to the hook's scope.
+func (h *Hook) AddTags(tags map[string]string) {
+	h.scope.SetTags(tags)
+	h.setHub()
+}
+
+// A FallbackFunc can be used to attempt to handle any errors in logging, before
+// resorting to Logrus's standard error reporting.
+type FallbackFunc func(*logrus.Entry) error
+
+// Fallback sets a fallback function, which will be called in case logging to
+// sentry fails. In case of a logging failure in the Fire() method, the
+// fallback function is called with the original logrus entry. If the
+// fallback function returns nil, the error is considered handled. If it returns
+// an error, that error is passed along to logrus as the return value from the
+// Fire() call. If no fallback function is defined, a default error message is
+// returned to Logrus in case of failure to send to Sentry.
+func (h *Hook) Fallback(fb FallbackFunc) {
+	h.fallback = fb
+}
+
+// SetKey sets an alternate field key. Use this if the default values conflict
+// with other loggers, for instance. You may pass "" for new, to unset an
+// existing alternate.
+func (h *Hook) SetKey(oldKey, newKey string) {
+	if oldKey == "" {
+		return
+	}
+	if newKey == "" {
+		delete(h.keys, oldKey)
+		return
+	}
+	delete(h.keys, newKey)
+	h.keys[oldKey] = newKey
+}
+
+func (h *Hook) key(key string) string {
+	if val := h.keys[key]; val != "" {
+		return val
+	}
+	return key
+}
+
+// Levels returns the list of logging levels that will be sent to
+// Sentry.
+func (h *Hook) Levels() []logrus.Level {
+	return h.levels
+}
+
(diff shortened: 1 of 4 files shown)


jeffwidman:
Be great to see this merged!

jeffwidman:
While this would be helpful in the short term for migrating a project from `raven-go`, in the long term it might be better to hold off on this until there's more clarity on [the discussion about adding structured logging to the `go` std lib](https://github.com/golang/go/discussions/54763). 

Because if that turns into a proposal that lands, then I suspect most applications will start migrating toward that or toward logging libraries that extend that. So usage of `logrus` would probably decline given that [it's in maintenance mode](https://github.com/sirupsen/logrus#logrus----) so unlikely to be updated to that.

Wouldn't want to add the `logrus` support for a few months to a year, then have to do a full version bump simply to drop it again. I expect that the logging discussion should result in a decision either way within a few months, so even if that discussion gets rejected then the worst case is it only delays the inclusion of the logrus hook by a few months.

flimzy:
Rebased and force-pushed after recent changes to master.

kamilogorek:
? Wouldn't want to add the logrus support for a few months to a year, then have to do a full version bump simply to drop it again. I expect that the logging discussion should result in a decision either way within a few months, so even if that discussion gets rejected then the worst case is it only delays the inclusion of the logrus hook by a few months.

> IMO, it wouldn't make sense to drop it until logrus is no longer widely used, which will probably be 5-10 years, even if no new projects ever adopt it starting immediately.

Agree that it's still fine to support libraries that are and probably will be very widely used for a long time.

If anything, I'd much rather drop it in the future, than not have support for it during that period.

We are also still pre-v1 release, so if it won't gain any traction, we'll always have the opportunity to not support it anymore.

logrus/logrusentry.go
```diff
@@ -0,0 +1,233 @@
...
+// These keys may be overridden by calling SetKey on the hook object.
+const (
+	// FieldRequest holds an *http.Request.
+	FieldRequest = "request"
+	// FieldUser holds a User or *User value.
+	FieldUser = "user"
+	// FieldTransaction holds a transaction ID as a string.
+	FieldTransaction = "transaction"
+	// FieldFingerprint holds a string slice ([]string), used to dictate the
+	// deduplication of this event.
```
kamilogorek, line 27:
```suggestion
	// grouping of this event.
```

logrus/logrusentry.go
```diff
@@ -0,0 +1,233 @@
...
+)
+
+// Hook is the logrus hook for Sentry.
+//
+// It is not safe to configure the hook while logging is happening. Please
+// perform all configuration before using it.
+type Hook struct {
+	client   *sentry.Client
+	scope    *sentry.Scope
+	hub      *sentry.Hub
```
kamilogorek, line 42: `hub` is already responsible for maintaining proper `scope` and `client` references. Storing them on the hook directly seems unnecessary.

logrus/logrusentry.go
```diff
@@ -0,0 +1,233 @@
...
+
+// setHub refreshes the hub, to be used any time client or scope changes.
+func (h *Hook) setHub() {
+	h.hub = sentry.NewHub(h.client, h.scope)
+}
+
+// AddTags adds tags to the hook's scope.
+func (h *Hook) AddTags(tags map[string]string) {
+	h.scope.SetTags(tags)
+	h.setHub()
```
kamilogorek, line 81: This line wont be necessary if we use `hub` directly anywhere.

logrus/logrusentry.go
```diff
@@ -0,0 +1,233 @@
...
+// NewFromClient initializes a new Logrus hook which sends logs to the provided
+// sentry client.
+func NewFromClient(levels []logrus.Level, client *sentry.Client) *Hook {
+	h := &Hook{
+		levels: levels,
+		client: client,
+		scope:  sentry.NewScope(),
+		keys:   make(map[string]string),
+	}
+	h.setHub()
```
kamilogorek, line 69: `hub: sentry.NewHub(client, sentry.NewScope())` in case we drop `scope/client` pair.

logrus/logrusentry.go
```diff
@@ -0,0 +1,233 @@
...
+type FallbackFunc func(*logrus.Entry) error
+
+// Fallback sets a fallback function, which will be called in case logging to
+// sentry fails. In case of a logging failure in the Fire() method, the
+// fallback function is called with the original logrus entry. If the
+// fallback function returns nil, the error is considered handled. If it returns
+// an error, that error is passed along to logrus as the return value from the
+// Fire() call. If no fallback function is defined, a default error message is
+// returned to Logrus in case of failure to send to Sentry.
+func (h *Hook) Fallback(fb FallbackFunc) {
```
kamilogorek, line 95: In Go, this would be a getter. `Fallback/SetFallback` pairs should be used instead.

logrus/logrusentry.go
```diff
@@ -0,0 +1,233 @@
...
+	logrus.TraceLevel: sentry.LevelDebug,
+	logrus.DebugLevel: sentry.LevelDebug,
+	logrus.InfoLevel:  sentry.LevelInfo,
+	logrus.WarnLevel:  sentry.LevelWarning,
+	logrus.ErrorLevel: sentry.LevelError,
+	logrus.FatalLevel: sentry.LevelFatal,
+	logrus.PanicLevel: sentry.LevelFatal,
+}
+
+func (h *Hook) entry2event(l *logrus.Entry) *sentry.Event {
```
kamilogorek, line 149: Use `entryToEvent` instead, as we already have `applytoEvent` in the codebase, as well as std libs are using this convention, eg `Atoi`

kamilogorek: We have all necessary types/fields for such data - https://github.com/getsentry/sentry-go/blob/2503eee7cb11316861b52c5c68c6f364092c5d78/interfaces.go#L167-L196
Using it, will make it way more readable in the UI.

kamilogorek (changes requested):

logrus/logrusentry.go
flimzy: Thanks for the suggestion. I think I've mad the correct change in b922f35abe499ac9b81f7d9208433004bab029f1, but please let me know if I've misunderstood something.

flimzy: Thanks for the suggestion. I think I've mad the correct change in b922f35abe499ac9b81f7d9208433004bab029f1, but please let me know if I've misunderstood something.

logrus/logrusentry.go
flimzy: renamed.

logrus/logrusentry.go
flimzy: I'm not sure I follow.  This is just a setter, not a getter.  Are you saying a getter needs to be added? I can trivally add a getter, but is there any value in such a getter?

flimzy:
Thanks for the review @kamilogorek . I've addressed most of your comments, and added a follow-up question.

logrus/logrusentry.go
kamilogorek: No no, I mean that if it is a setter (which it is), then it should be named `SetFallback`. Without `Set`, its implied getter.

flimzy: ah, got it.  I'll update.

kamilogorek:
Thanks, there's one thing left. You did the reverse of what I'd prefer 😅
Here https://github.com/getsentry/sentry-go/pull/471#discussion_r998070949 instead of holding onto `scope/client` separately, keep the reference to `hub` itself, not the other way around.

It's because `Hub` already has references to its client and scope via `hub.Scope()` and `hub.Client()` methods.

flimzy:
> You did the reverse of what I'd prefer sweat_smile

Ha! Oops. Thanks for the explanation. Will follow up soon.

flimzy:
Rebased and updated to use `hub` in place of `client` and `scope`.

cleptric:
@flimzy We might want this to be its own module, to not add further deps to the main package.

cc @kamilogorek What do you think?

flimzy:
> We might want this to be its own module, to not add further deps to the main package.

I don't think it adds any deps anyway.  The only change to `go.mod` is a promotion of `logrus` from an indirect to direct dependency, so won't actually change the deps that any consumer has to download.

cleptric:
btw, @flimzy test failures are caused by a new option added via https://github.com/getsentry/sentry-go/pull/490

flimzy:
I've added an example. Feedback welcome. It's hard to know how detailed to be in an example.

cleptric:
Thanks a lot, looks good to me!

kamilogorek:
All requests from the previous review were applied, thanks!
As for the separate package, I think we can postpone it with logrus, as it's already been our dep before, which you already notices.
We can also test it out in the wild, and verify whether we should pull the support to v1 of the SDK once we go for it.

kamilogorek (approved):

kamilogorek:
I just noticed that filename is called `logrusentry` instead of `logru*s*sentry` haha. But I think it's actually a good wordplay, and it's not affecting the behavior, so we can leave it as is if you want😅

flimzy:
Re: `logrusentry` vs `logrussentry`, it was an intentional wordplay, but perhaps that violates the principle of least surprise?  I can go either way ;)

kamilogorek:
Nah, I like the idea as well!

PR #482, opened 2022-10-20
feat: Add a method to start a transaction

phacops:
In the quest of having a similar API as the other SDKs, we wanted to add a `StartTransaction` function to start a transaction.

Since the Go SDK makes everything a span and the root span becomes the transaction, we don't need more than a function that calls `StartSpan` in order to make something compatible with a big difference being `TransactionName` is now required instead of `Op` and `Op` can be set as an option.

You might notice I chose to `panic` if there's already a transaction in progress in the context. I'm not 100% sure how other SDKs are handling this case. We could choose to return the current transaction, stop the current transaction and start a new one instead of a `panic`. This will still means several transactions can happen at the same time, just with different contexts.

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -5,6 +5,7 @@ import (
 	"crypto/rand"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"regexp"
@@ -567,6 +568,13 @@ func TransactionName(name string) SpanOption {
 	}
 }
 
+// OpName sets the operation name for a given span
+func OpName(name string) SpanOption {
+	return func(s *Span) {
+		s.Op = name
+	}
+}
+
 // ContinueFromRequest returns a span option that updates the span to continue
 // an existing trace. If it cannot detect an existing trace in the request, the
 // span will be left unchanged.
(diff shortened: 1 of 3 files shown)


tracing.go
```diff
@@ -626,3 +634,27 @@ func spanFromContext(ctx context.Context) *Span {
...
+// ErrTransactionAlreadyInProgress is returne when we try to start a transaction
+// when another one is in progress.
+var ErrTransactionAlreadyInProgress = errors.New("transaction already in progress")
+
+// StartTransaction will create a transaction (root span) if there's no existing
+// transaction in the context.
+func StartTransaction(ctx context.Context, name string, options ...SpanOption) *Span {
+	currentTransaction := ctx.Value(spanContextKey{})
+	if currentTransaction != nil {
+		panic(ErrTransactionAlreadyInProgress)
```
AbhiPrasad, line 647: Not sure if I like the idea of panicing here, we shouldn't cause possible runtime issues with our SDK methods.

Could we return `err` instead here? And have users handle this?

phacops: Right, I can return an error instead here. I'll be back to asking if I can `panic` in the HTTP handler though, and so, maybe you can shed some light on how other SDKs are dealing with this if that's a case that can happen?

Could we silently return the existing transaction with a `bool` to say it already was in progress instead? Could we just return `nil` and the user would have test for the transaction or get the existing one themselves?

AbhiPrasad: Yeah so there are the two options

1. We return `err`, so the users have to grab the existing transaction themselves.
2. We return the active transaction and don't create one. This does mean that the transaction won't have the context supplied in the constructor.

Option 2 is more "magic" behavior, but I think it's correct since two transactions running at once is basically useless to a person, so let us maybe do that. 

> Could we silently return the existing transaction with a bool to say it already was in progress instead

Sounds good to me, I think that's cleaner API

AbhiPrasad (approved):
:shipit:

phacops:
In the end, the compromise was to return the existing transaction to avoid any `panic` or additional error management. In the future, we'd need to align the behavior of this SDK on others to actually start a new transaction, referencing the parent one and sending both at the end.

cleptric (approved):

GoWithPlay:
Great work! So when will this feature be released as v0.15.0 of this package??

AbhiPrasad:
Yes @GoWithPlay, this will be part of the next release!

AbhiPrasad:
It'll be less than 7 months 😅, look forward to something within the next 2-3 weeks.

GoWithPlay:
@AbhiPrasad Sounds really great!

cleptric:
@GoWithPlay `0.15.0` was released yesterday, which includes `StartTransaction`.

AbhiPrasad:
> It'll be less than 7 months 😅, look forward to something within the next 2-3 weeks.

More like 2 days! :shipit:

PR #485, opened 2022-10-30
feat: Add ClientOptions.SendDefaultPii

### LINKED ISSUE #175: PII opt-in by default
We need to add to the SDK a new config bool such that anything that we know will identify a user is not sent on events by default. The entire user object essentially, auth headers, IP addresses, that sort of thing.

See `send_default_pii` in the Python SDK.
https://docs.sentry.io/error-reporting/configuration/?platform=python#send-default-pii

The current docs for Go suggests that it exists as `SendDefaultPii`, but it doesn't exist yet.
https://docs.sentry.io/error-reporting/configuration/?platform=go#send-default-pii

(see also https://github.com/getsentry/develop/issues/466)

cleptric:
As `scope.SetUser` has to be explicitly called by the user, I did not add a check for `SendDefaultPii` there. 

Closes #175

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 - feat: Extend User inteface by adding Data, Name and Segment (#483)
+- feat: Add ClientOptions.SendDefaultPii #485
 
 ## 0.14.0
 
--- a/client.go
+++ b/client.go
@@ -129,6 +129,9 @@ type ClientOptions struct {
 	// and if applicable, caught errors type and value.
 	// If the match is found, then a whole event will be dropped.
 	IgnoreErrors []string
+	// If this flag is enabled, certain personally identifiable information (PII) is added by active integrations.
+	// By default, no such data is sent.
+	SendDefaultPii bool
 	// BeforeSend is called before error events are sent to Sentry.
 	// Use it to mutate the event or return nil to discard the event.
 	// See EventProcessor if you need to mutate transactions.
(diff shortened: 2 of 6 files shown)


cleptric:
- I've noticed that the PHP SDK still sends all headers but sets their value to `[Filtered]`.
Other SDKs drop the keys completely.
- Do I need to add checks for the request body as well?

AbhiPrasad:
> Do I need to add checks for the request body as well?

Yes, it's a source of PII as well

cleptric:
While the `Request` struct contains a `Data` field, I can't find any usage of this field in the SDK. So we're good to go with this PR.

AbhiPrasad (approved):

CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
 # Changelog
 
+## Unreleased
+
 - fix: Scope values should not override Event values (#446)
 - feat: Extend User inteface by adding Data, Name and Segment (#483)
 - feat: Make maximum amount of spans configurable (#460)
+- feat: Add ClientOptions.SendDefaultPii #485
```
rhcarvalho, line 8: Happy to see the Sentry Go SDK evolving!

One suggestion: initialisms like "PII" are spelled with a consistent case as per Go community conventions https://github.com/golang/go/wiki/CodeReviewComments#initialisms.

The option could be called `ClientOptions.SendDefaultPII`.

cleptric: Thanks, I opened https://github.com/getsentry/sentry-go/pull/490

PR #487, opened 2022-10-30
docs: Run go fmt

cleptric:
Just a quick `go fmt` sweep.


AbhiPrasad (approved):
This isn’t checked in CI?

cleptric:
@AbhiPrasad We definitely run some linters (https://github.com/getsentry/sentry-go/blob/master/.golangci.yml) but I haven't yet looked deeply into this topic.

PR #490, opened 2022-11-02
ref: Rename ClientOptions.SendDefaultPii to ClientOptions.SendDefaultPII

cleptric:
re: https://github.com/getsentry/sentry-go/pull/485#discussion_r1011500255

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 - fix: Scope values should not override Event values (#446)
 - feat: Extend User inteface by adding Data, Name and Segment (#483)
 - feat: Make maximum amount of spans configurable (#460)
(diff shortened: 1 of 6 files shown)


AbhiPrasad (commented):
I understand this helps unify with the Go ecosystem - but in this case we use the lowercase version of `SendDefaultPii` in every single other SDK, I'd rather stay consistent with the rest of the SDK

cleptric:
I think it's ok to stay unified with the Go ecosystem, especially if we expand on better linting in the future.
In terms of DevEx, IDEs should autocomplete on this just fine.

AbhiPrasad (approved):
We'll have to update the docs about this as well then!

cleptric:
Yes Sir! https://github.com/getsentry/sentry-docs/pull/5708

PR #497, opened 2022-11-10
feat: add support for stack trace from github.com/pierrre/errors

pierrre:
Related to https://github.com/getsentry/sentry-go/issues/253#issuecomment-1284034722

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/google/go-cmp v0.5.9
 	github.com/kataras/iris/v12 v12.2.0-beta5
 	github.com/labstack/echo/v4 v4.9.0
+	github.com/pierrre/errors v0.0.5
 	github.com/pingcap/errors v0.11.4
 	github.com/pkg/errors v0.9.1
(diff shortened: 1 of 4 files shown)


cleptric:
I'm sorry, but we can't get this merged.
We only support libraries that are frequently used in the ecosystem.
Can you work around this by creating your own integration?

cleptric (changes requested):
I'm sorry, but we can't get this merged.
We only support libraries that are frequently used in the ecosystem.
Can you work around this by creating your own integration?

pierrre:
I'm willing to write the code that will add support for custom stack trace.
I just need to know what is an acceptable solution (the "user interface")

flimzy:
What if you make `github.com/pierrre/errors` support `github.com/pkg/errors` stack traces?  I believe it would be a fairly simple change to add a `Stacktrace() errors.Stacktrace` method on the errors in that package, no?  (See https://pkg.go.dev/github.com/pkg/errors#hdr-Retrieving_the_stack_trace_of_an_error_or_wrapper)

PR #498, opened 2022-11-14
ref: Unify TracesSampler

cleptric:
In order to implement Dynamic Sampling, I had to make changes to how the `TracesSampler` works in the SDK.
I'm aware that this is a breaking change, but it is required at this point.

For Dynamic Sampling, a `sample_rate` item needs to be present in the `trace` envelope header.
With the current implementation, accessing these values was not possible.

- Unify the `TracesSampler` to always return a `float64`. This is in line with the implementation of all other SDKs and https://develop.sentry.dev/sdk/performance/#tracessampler
- I made the `sample` method quite verbose, if `debug: true`, as this is a constant pain point for developers to figure out what is happening.
- As we can't denote `TracesSampleRate: 0.0/nil`, I introduced a new `ClientOption` called `EnableTracing`, which turns off tracing completely by default. This setting is required, as we need a way to allow developers to turn off sampling that could be started by a `sentry-trace` header with a `true` sampling decision. I did consider using a pointer, but this was too clunky.

Open questions:

- ~~The old and new implementations would also sample spans if the `TracesSampler` option is set. I hence would vote for only calling `sample()` on transactions. @AbhiPrasad Do I miss the use-case of sampling spans separately to a transaction?~~ I missed this when making my changes. Not an issue anymore, I kept the old logic in place.


client.go
```diff
@@ -235,10 +238,6 @@ type Client struct {
 // single goroutine) or hub methods (for concurrent programs, for example web
 // servers).
 func NewClient(options ClientOptions) (*Client, error) {
-	if options.TracesSampleRate != 0.0 && options.TracesSampler != nil {
```
cleptric, line 238: I removed this error, as `TracesSampler` only takes priority over `TracesSampleRate`

client.go
```diff
@@ -323,7 +322,7 @@ func (client *Client) setupTransport() {
 			// accommodate more concurrent events.
 			// TODO(tracing): consider using separate buffers per
 			// event type.
-			if opts.TracesSampleRate != 0 || opts.TracesSampler != nil {
+			if opts.EnableTracing {
```
cleptric, line 325: We might make this the default.

client.go
```diff
@@ -323,7 +322,7 @@ func (client *Client) setupTransport() {
...
 				httpTransport.BufferSize = 1000
```
cleptric, line 326: We might make this the default.

client.go
```diff
@@ -126,6 +125,8 @@ type ClientOptions struct {
 	// 0.0 is treated as if it was 1.0. To drop all events, set the DSN to the
 	// empty string.
 	SampleRate float64
+	// Enable performance tracing.
+	EnableTracing bool
```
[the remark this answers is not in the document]
cleptric: `TracesSampleRate` will always be defined, with a zero value of `0.0`.
Hence the need for this option arose 😕

tracing.go
```diff
@@ -276,38 +278,76 @@ func (s *Span) MarshalJSON() ([]byte, error) {
...
+
+	// https://develop.sentry.dev/sdk/performance/#sampling
+	// #1 tracing is not enabled.
+	if !clientOptions.EnableTracing {
+		Logger.Printf("Dropping transaction: EnableTracing is set to %t", clientOptions.EnableTracing)
+		return SampledFalse
+	}
+
+	// #2 explicit sampling decision via StartSpan/StartTransaction options.
+	if s.Sampled != SampledUndefined {
```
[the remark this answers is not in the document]
cleptric: Good point, will remove it and type `span.Sample` as `bool`.

client.go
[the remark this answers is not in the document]
AbhiPrasad: Hmm - I guess then we have performance turned on by default, just with no sample rate, which maybe isn't the end of the world. If users don't want it, they can turn it off explicitly.

But that might be too invasive, let's just stuck with the boolean then.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Changelog
 
+- *[breaking]* ref: Unify TracesSampler (#444)
```
AbhiPrasad, line 3: Can we add migration docs?

AbhiPrasad (approved):

PR #501, opened 2022-11-17
Added doc on how to handle logrus.Fatalf events

Mindstan:
This PR adds some doc on how to handle `logrus.Fatalf` events.

When calling `logrus.Fatalf()`, `logrus` calls `os.Exit(1)` immediately after treating the log. They are two consequences :
- all `defer` are ignored;
- it is almost certain that Sentry will not have time to send the `fatal` event to the remote server before the program stops.

Fortunately, [`logrus.RegisterExitHandler`](https://github.com/sirupsen/logrus#fatal-handlers) allows to run arbitrary code before the call to `os.Exit`. I propose to add it to the `logrus` example.

### DIFF
--- a/example/logrus/main.go
+++ b/example/logrus/main.go
@@ -43,9 +43,16 @@ func main() {
 	defer sentryHook.Flush(5 * time.Second)
 	logger.AddHook(sentryHook)
 
+	// Flushes before calling os.Exit(1) when using logger.Fatal
+	// (else all defers are not called, and Sentry does not have time to send the event)
+	logrus.RegisterExitHandler(func() { sentryHook.Flush(5 * time.Second) })
+
 	// The following line is logged to STDERR, but not to Sentry
 	logger.Infof("Application has started")
 
 	// The following line is logged to STDERR and also sent to Sentry
 	logger.Errorf("oh no!")
+
+	// The following line is logged to STDERR, sent to Sentry, then the program abruptly stops
+	logger.Fatalf("can't continue...")
 }


mbaumanndev (approved):

example/logrus/main.go
```diff
@@ -43,9 +43,16 @@ func main() {
 	defer sentryHook.Flush(5 * time.Second)
 	logger.AddHook(sentryHook)
 
+	// Flushes before calling os.Exit(1) when using logger.Fatal
+	// (else all defers are not called, and Sentry does not have time to send the event)
+	logrus.RegisterExitHandler(func() { sentryHook.Flush(5 * time.Second) })
```
flimzy, line 48: Would it not make sense to do this unconditionally, so the caller doesn't need to?

Mindstan: We could, but it is unnecessary if Sentry sends its events synchronously.

example/logrus/main.go
```diff
@@ -43,9 +43,16 @@ func main() {
 	defer sentryHook.Flush(5 * time.Second)
 	logger.AddHook(sentryHook)
 
+	// Flushes before calling os.Exit(1) when using logger.Fatal
+	// (else all defers are not called, and Sentry does not have time to send the event)
```
cleptric, line 47:
```suggestion
	// When using `logger.Fatal`, flush the event buffer before `logger.Exit(1)` is called.
	// 
```

example/logrus/main.go
```diff
@@ -43,9 +43,16 @@ func main() {
...
+	logrus.RegisterExitHandler(func() { sentryHook.Flush(5 * time.Second) })
+
 	// The following line is logged to STDERR, but not to Sentry
 	logger.Infof("Application has started")
 
 	// The following line is logged to STDERR and also sent to Sentry
```
cleptric, line 53:
```suggestion
	// Log an error to STDERR which is also send to Sentry
```

example/logrus/main.go
```diff
@@ -43,9 +43,16 @@ func main() {
 	defer sentryHook.Flush(5 * time.Second)
 	logger.AddHook(sentryHook)
 
+	// Flushes before calling os.Exit(1) when using logger.Fatal
+	// (else all defers are not called, and Sentry does not have time to send the event)
+	logrus.RegisterExitHandler(func() { sentryHook.Flush(5 * time.Second) })
+
 	// The following line is logged to STDERR, but not to Sentry
```
cleptric, line 50:
```suggestion
	// Log a InfoLevel entry STDERR which is not send to Sentry
```

cleptric:
@Mindstan Ping 🙂

cleptric (approved):

PR #507, opened 2022-12-05
[Dynamic Sampling] String()

cleptric:
This adds the implementation of creating the baggage header string.
The implementation relies mostly on the vendored-in OTel package, which takes care of length and encoding constrains.


dynamic_sampling_context_test.go
```diff
@@ -128,3 +129,22 @@ func TestHasEntries(t *testing.T) {
...
+
+	dsc = DynamicSamplingContext{
+		Frozen: true,
+		Entries: map[string]string{
+			"trace_id":    "d49d9bf66f13450b81f65bc51cf49c03",
+			"public_key":  "public",
+			"sample_rate": "1",
+		},
+	}
+	assertEqual(t, strings.Contains(dsc.String(), "sentry-trace_id=d49d9bf66f13450b81f65bc51cf49c03"), true)
```
cleptric, line 147: Comparing the whole DSC string was flaky, hence I split it up.

PR #516, opened 2022-12-13
dev: Add Makefile with common commands

tonyo:
We can later update CONTRIBUTING.md and Actions workflows to use "make" targets as the source of truth.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,38 @@
+.DEFAULT_GOAL := help
+
+# Parse Makefile and display the help
+help: ## Show help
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,38 @@
+.DEFAULT_GOAL := help
+
+# Parse Makefile and display the help
+help: ## Show help
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
```
tonyo, line 5: A bit of magic to auto-generate usage docs for make targets:

image

Credits go to https://marmelab.com/blog/2016/02/29/auto-documented-makefile.html

cleptric (approved):
:shipit:

PR #520, opened 2022-12-20
fix(dynamic-sampling): Do not crash in Span.Finish() when Client is empty

### LINKED ISSUE #518: panic because `nil` client in `DynamicSamplingContextFromTransaction`
## Summary
In a context where we do not have initialized sentry, when finishing a transaction that have `Sampled = sentry.SampledFalse` will trigger a panic. I have a piece of code that conditionally force the sampling of a given transaction, this is working well but if I run my unit tests on the function that panics.

In general, I think the SDK should not panic if sentry is not configured.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x635330]

goroutine 1 [running]:
github.com/getsentry/sentry-go.DynamicSamplingContextFromTransaction(0xc00015c000)
	/tmp/gopath1614006584/pkg/mod/github.com/getsentry/sentry-go@v0.16.0/dynamic_sampling_context.go:46 +0xb0
github.com/getsentry/sentry-go.(*Span).toEvent(0xc00015c000)
	/tmp/gopath1614006584/pkg/mod/github.com/getsentry/sentry-go@v0.16.0/tracing.go:405 +0x2e5
github.com/getsentry/sentry-go.(*Span).Finish(0xc00015c000)
	/tmp/gopath1614006584/pkg/mod/github.com/getsentry/sentry-go@v0.16.0/tracing.go:173 +0xb5
```

## Steps To Reproduce
https://go.dev/play/p/vzesJ4l9TJa

## Expected Behavior
No panic on `Finish()`

### SDK
 - `sentry-go` version: `v0.16.0`
 - Go version: 1.19
 - Using Go Modules? yes

### Sentry

Happy to shoot a PR if you quickly explain how would you like to handle that 🙏🏻

tonyo:
~Making a draft since I'm not entirely sure if there are any other places right now where we might have this issue, and also because I'll have to confirm what DynamicSamplingContext to return when Client or Scope are empty.~

Should be ready for review now.

Fixes #518.


cleptric (commented):
This will attach the DSC to the event here https://github.com/getsentry/sentry-go/blob/6588be354e21c7e5bd0307e80ee7a81bfa3dd371/tracing.go#L404

In the transport, we'll skip attaching the DSC as it's empty
https://github.com/getsentry/sentry-go/blob/master/transport.go#L103

I think this is a feasible solution.

dynamic_sampling_context.go
```diff
@@ -43,6 +43,14 @@ func DynamicSamplingContextFromTransaction(span *Span) DynamicSamplingContext {
 	hub := hubFromContext(span.Context())
 	scope := hub.Scope()
 	client := hub.Client()
+
+	if client == nil || scope == nil {
+		return DynamicSamplingContext{
+			Entries: map[string]string{},
+			Frozen:  false,
```
tonyo, line 50: Made it more explicit to make sure there are no default-value surprises.

cleptric (approved):

PR #524, opened 2023-01-09
feat: When ClientOptions.SendDefaultPii is false, send http headers without sensitive data

### LINKED ISSUE #523: No http headers are sent if `ClientOptions.SendDefaultPii` is set to false
## Summary
`SendDefaultPii` option was added in #485 .
This option behaves as follows:
- client is not nil and SendDefaultPii is true -> all headers are sent
- client is not nil and SendDefaultPii is false -> no headers are sent
- client is nil -> headers without sensitive data are sent

Is this expected behaviour?
In case of other language, it looks like all http headers except sensitive headers are sent when `SendDefaultPii` is false.

sentry-python:
https://github.com/getsentry/sentry-python/blob/2f67f12e405f8a6f89418d96071158367fcf516f/sentry_sdk/integrations/_wsgi_common.py#L160

sentry-spring:
https://github.com/getsentry/sentry-java/blob/8ade22549f328b48fd01aea726141f7186f3bd65/sentry-spring/src/main/java/io/sentry/spring/SentryRequestResolver.java#L33

## Steps To Reproduce

```go
func Run(addr string) error {
	_ = sentry.Init(sentry.ClientOptions{
		SendDefaultPII: false,
	})

	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		ctx := r.Context()

		hub.CaptureMessage("my message")
	})

	handler := sentryhttp.New(sentryhttp.Options{}).Handle(http.DefaultServeMux)
	return http.ListenAndServe(addr, handler)
}
 ```

## Expected Behavior
When `ClientOptions.SendDefaultPii` is false, http headers except sensitive data are sent.
I made a PR #524 .

### SDK
- sentry-go version: v0.16.0
- Go version: 1.19
- Using Go Modules? yes

### Sentry

### PR DESCRIPTION
**Pull request by ikorihn**
This fixes #523 .

Non-sensitive headers are sent even if `SendDefaultPII` is set to false.

I referred to this: 
https://github.com/getsentry/sentry-java/blob/8ade22549f328b48fd01aea726141f7186f3bd65/sentry-spring/src/test/kotlin/io/sentry/spring/SentrySpringFilterTest.kt#L196

### CODE REVIEW

File: interfaces.go [hunk 1]
```diff
@@ -165,35 +165,23 @@ func NewRequest(r *http.Request) *Request {
...
-
-			for k, v := range r.Header {
-				headers[k] = strings.Join(v, ",")
-			}
-
-			if addr, port, err := net.SplitHostPort(r.RemoteAddr); err == nil {
-				env = map[string]string{"REMOTE_ADDR": addr, "REMOTE_PORT": port}
-			}
-		}
-	} else {
```
cleptric, line 188: Thanks for the contribution and raising this issue!

Couldn't we move this block inside the client check instead? IMO, this would be more readable than introducing a new function.

ikorihn: Thanks for review.

> this would be more readable than introducing a new function

Right, moved it into if block.
8373d13221838d40aa3d23e841655fd07532049b

tonyo (approved):
LGTM, thanks!

cleptric (approved):

PR #530, opened 2023-01-12
ci: Add Codecov

cleptric:
> -count n
    Run each test, benchmark, and fuzz seed n times (default 1).
    If -cpu is set, run n times for each GOMAXPROCS value.
    Examples are always run once. -count does not apply to
    fuzz tests matched by -fuzz.
    
As count defaults to 1, I removed it.


.github/workflows/test.yml
```diff
@@ -49,14 +49,16 @@ jobs:
         run: go mod tidy -go=1.19 -compat=1.17 && git diff --exit-code
         if: ${{ matrix.go == '1.19' }}
       - name: Test
-        run: go test -count=1 ./...
```
tonyo, line 52: AFAIK we use `-count=1` to disable test result caching: https://stackoverflow.com/questions/73432741/why-does-count-1-ignores-caching-in-go-tests
But we could totally document this better.

tonyo (approved):

PR #531, opened 2023-01-12
Setup CodeQL

mdtro:
This sets up CodeQL to scan on PRs, the default branch, and weekly to identify security issues.

The new workflow in CI will not block PRs.


cleptric:
Shall we remove https://github.com/getsentry/sentry-go/blob/master/.github/workflows/codeql-analysis.yml then?

cleptric (approved):

mdtro:
> Shall we remove https://github.com/getsentry/sentry-go/blob/master/.github/workflows/codeql-analysis.yml then?

Good catch! That's my bad. I didn't see the existing workflow. It's using some of the older versions and the non-extended query packs -- so just opting to replace it. :)

PR #537, opened 2023-01-19
feat: Add initial OpenTelemetry support

### LINKED ISSUE #486: [GO] Add basic OpenTelemetry support
### Details
- SDK develop spec https://github.com/getsentry/develop/pull/686
- span processor - transforming OTEL spans into Sentry spans within Sentry transactions
  - span name and description
  - parent child relationships
- trace/context propagator - including DSC and trace context 
  - https://opentelemetry.io/docs/reference/specification/context/api-propagators/

## Reference
Ruby issues with initial implementation https://github.com/getsentry/sentry-ruby/issues/1907

### LINKED ISSUE #526: [OpenTelemetry] Implement OTel Propagator
Subtask of https://github.com/getsentry/sentry-go/issues/486

The goal is to implement trace/context OpenTelemetry propagator (including DSC and trace context): https://opentelemetry.io/docs/reference/specification/context/api-propagators/

JavaScript implementation can be used as a template: https://develop.sentry.dev/sdk/performance/opentelemetry/#step-2-implement-the-sentrypropagator-on-your-sdk-add-sentrypropagator

cleptric:
This PR adds two main integration components (along with all the main supporting functions to enrich Sentry spans with OTel data):
* [Span Processor](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-processor)
* [Propagator](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md)

Main discussion: https://github.com/getsentry/sentry/discussions/40712 

Closes #526, closes #486.

A list of open issues with the implementation:

1. Dynamic Sampling Context propagation from nested spans doesn't work properly. Main reason: no publicly exposed mechanism to get the transaction (as in, a root span) for the given nested span.
    #553
1. Baggage parsing bug is not yet fixed or acknowledged: https://github.com/open-telemetry/opentelemetry-go/issues/3601 We might want to use our vendored implementation and fix the bug there.
    #554
1. Associating errors with transactions is generally implemented (see `otel/event_processor.go`), but there's a caveat: the EventHint of the event processor has to contain a valid `Context` with a set OpenTelemetry span. This means that e.g. messages captured with `sentry.CaptureMessage` or `hub.CaptureMessage` won't be linked to Sentry spans at the moment.
    #555
1. "instrumenter" configuration in ClientOptions -- we need to confirm whether we need that.
    #556
1. Span attributes/description for special cases (HTTP, DB, RPC, etc.) might be incomplete.
    #557


otel/internal/utils/sentryrequest.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/getsentry/sentry-go"
+	"go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.12.0"
+)
+
+func IsSentryRequestSpan(ctx context.Context, s trace.ReadOnlySpan) bool {
+	attributes := s.Attributes()
+
+	// TODO(michi): can we access the attribute directly?
```
cleptric, line 17: I haven't found a better way, so we might roll with this instead. Will remove the comments if you're 👍

tonyo: Let's leave it like this for now, we can later implement something similar to what they do in sentry exporter: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/sentryexporter/sentry_exporter.go#L376-L394

otel/propagator.go
```diff
@@ -0,0 +1,137 @@
...
+
+func NewSentryPropagator() propagation.TextMapPropagator {
+	return &sentryPropagator{}
+}
+
+// Inject sets Sentry-related values from the Context into the carrier.
+//
+// https://opentelemetry.io/docs/reference/specification/context/api-propagators/#inject
+func (p sentryPropagator) Inject(ctx context.Context, carrier propagation.TextMapCarrier) {
+	sentry.Logger.Printf("\n--- Propagator Inject\nContext: %#v\nCarrier: %#v\n", ctx, carrier)
```
cleptric, line 24: This might be super noisy for users. I use `Debug: true` a lot, so maybe we could reduce the output a little bit.

tonyo (approved):
🚢 🚂 🚀

PR #538, opened 2023-01-19
ref(tracing): Add constants for "sentry-trace" and "baggage" header names

tonyo:
This is helpful to have for the ongoing OTel work (happening in #537).


cleptric (approved):
🚢 

Maybe add a one-liner to the changelog, so we save some time once we craft the release notes.

PR #543, opened 2023-01-22
feat: Add Span.IsTransaction()

cleptric:
This is needed for OTel (https://github.com/getsentry/sentry-go/pull/537).

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -211,6 +211,11 @@ func (s *Span) SetTag(name, value string) {
 	s.Tags[name] = value
 }
 
+// Check if the given span is a transaction.
+func (s *Span) IsTransaction() bool {
+	return s.isTransaction
+}
+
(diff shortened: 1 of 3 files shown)


tracing.go
```diff
@@ -211,6 +211,11 @@ func (s *Span) SetTag(name, value string) {
 	s.Tags[name] = value
 }
 
+// Check if the given span is a transaction.
```
tonyo, line 214:
```suggestion
// IsTransaction checks if the given span is a transaction.
```

tracing_test.go
```diff
@@ -313,6 +313,22 @@ func TestStartTransaction(t *testing.T) {
 	}
 }
 
+func TestIsTansaction(t *testing.T) {
```
tonyo, line 316:
```suggestion
func TestIsTransaction(t *testing.T) {
```

tonyo (approved):
A couple suggestions

PR #544, opened 2023-01-23
test: Run tests and collect coverage for all submodules, if present

tonyo:
At the moment we only run test/coverage commands for the root module. 
Given the ongoing OTEL work (#537), we want to run the CI steps also for the new submodule(s).

The general approach for `test`/`test-coverage` targets was taken from https://github.com/open-telemetry/opentelemetry-go/blob/main/Makefile, and we can further iterate, if necessary.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,6 +4,8 @@ on:
     branches:
       - master
       - release/**
+      # REMOVEME
+      - tonyo/otel-plus-testing-submodules
   pull_request:
 permissions:
   contents: read
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,6 +69,7 @@ jobs:
         go: ["1.19", "1.18", "1.17"]
         os: [ubuntu, windows, macos]
       fail-fast: false
+  # Test the GOPATH mode with the oldest Go version we support
   test-gopath:
     name: GOPATH Mode
     runs-on: ubuntu-latest
@@ -84,6 +85,8 @@ jobs:
         working-directory: ${{ env.WORKDIR }}
     steps:
       - uses: actions/setup-go@v3
+        with:
+          go-version: "1.17"
       - uses: actions/checkout@v3
         with:
           path: ${{ env.WORKDIR }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,9 +49,9 @@ jobs:
         run: go mod tidy -go=1.19 -compat=1.17 && git diff --exit-code
         if: ${{ matrix.go == '1.19' }}
       - name: Test
-        run: go test -count=1 ./...
-      - name: Test (race)
-        run: go test -count=1 -race ./... -coverprofile=coverage.out -covermode=atomic
+        run: make test-race
+      - name: Collect coverage
+        run: make test-coverage
         # The race detector adds considerable runtime overhead. To save time on
         # pull requests, only run this step for a single job in the matrix. For
         # all other workflow triggers (e.g., pushes to a release branch) run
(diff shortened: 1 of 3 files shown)


cleptric:
Can we somehow bring back the behavior that we only run one test with `--race` to bring down CI time on PRs? Or are we ok with it taking ~5 mins now?
Also, macOS tests are not happy 😢

tonyo:
@cleptric To speed up the Windows builds, added back the condition to run things with race detection.

.github/workflows/test.yml
```diff
@@ -109,5 +113,5 @@ jobs:
       - name: Vet
         run: go vet ./...
       - name: Test
-        run: go test -count=1 ./...
+        run: make test-race
```
cleptric, line 116: Here we're running `test-race` as well.

.github/workflows/test.yml
```diff
@@ -49,16 +49,20 @@ jobs:
...
-      - name: Test (race)
-        run: go test -count=1 -race ./... -coverprofile=coverage.out -covermode=atomic
+        run: make test
+      - name: Test (with race detection)
+        run: make test-race
         # The race detector adds considerable runtime overhead. To save time on
         # pull requests, only run this step for a single job in the matrix. For
         # all other workflow triggers (e.g., pushes to a release branch) run
         # this step for the whole matrix.
         if: ${{ github.event_name != 'pull_request' || (matrix.go == '1.19' && matrix.os == 'ubuntu') }}
```
cleptric, line 59: Could we only omit `-race` for PRs when the runner being windows?
We have to be careful to produce the same amount of coverage reports in any case though.

.github/workflows/test.yml
tonyo: That's fine, we're running this on Ubuntu only, which is pretty fast.

.github/workflows/test.yml
tonyo: We don't generate any coverage reports in this step.
"Collect coverage" (the next one) runs for all platforms all the time.

cleptric: For my confirmation, looking at the matrix

- `make test` runs tests for all Go versions on ubuntu/windows/macOS
- `make test-coverage` runs tests for all Go versions on ubuntu/windows/macOS and produces coverage reports
- `make test-race` runs tests for all Go versions on ubuntu/windows/macOS, if not in a PR
- `make test-race` (in `test-gopath`) runs tests for all Go versions on ubuntu

tonyo: * `make test` runs tests for all Go versions on ubuntu/windows/macOS
    Yes   

* `make test-coverage` runs tests for all Go versions on ubuntu/windows/macOS and produces coverage reports
    Yes 
* `make test-race runs` tests for all Go versions on ubuntu/windows/macOS, if not in a PR
    If not in a PR -- yes, if PR -- only for 1.19 and Linux.
* make test-race (in test-gopath) runs tests for all Go versions on ubuntu
    Only for the default `actions/setup-go@v3` Go version, which now seems to be 1.18. We can pin that more explicitly.

cleptric: So we could combine `make test` & `make test-coverage` in the CI? Or is there something I'm missing?

tonyo: We could, yes. 
I was thinking about this more like about separate stages (first -- just run tests, second -- run them with race detection, third -- run with coverage), and generally, I see the value in running the tests more than once, at the very least to identify flaky tests (e.g. those that implicitly rely on randomness of map iteration, or, say, order of Baggage elements). But that can be easily changed by passing `-count 10` to `go test` (not doing it here, though)
So if you feel strongly about this, I can merge `test` and `test-coverage` steps.

cleptric (approved):

.github/workflows/test.yml
cleptric: It's ok, was just curious about the reasoning.

PR #552, opened 2023-01-27
feat: Add "transaction" pointer on Span

tonyo:
A proposal that we haven't really discussed in details, but was straightforward to implement to showcase. Needed for OTel work (#537).

### Problem

Spans know whether they are transactions (e.g. root spans) thanks to `isTransaction` flag, but there's no nice way to actually "reach" the transaction from a child span. It is needed, for instance, to get a dynamic sampling context for a nested span, in case we want to propagate it to other services.

### Proposal

Introduce `transaction` (name tentative) attribute on `Span`, that always points to the root span of the local span tree. For actual root spans, `transaction` is set to `nil`. With that attribute in place, we can also get rid of `isTransaction` flag.

### Notes and Caveats

1. A side-effect of removing `isTransaction` and introducing `transaction` is that naively created Spans will be transactions by default (since they'll have an empty `transaction`). We don't really promote or support this usecase anywhere (people should be using `StartTransaction` and `StartSpan`), but a thing to keep in mind.
1. Alternatively, for transaction spans, `transaction` can store a pointer to itself, and not `nil`. A small downside is that it will be impossible to create a root level `Span` instance in one go: it'll be something like `t := Span{..}; t.transaction = &t` in tests or elsewhere.

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -45,11 +45,11 @@ type Span struct { //nolint: maligned // prefer readability over optimal memory 
 	// parent refers to the immediate local parent span. A remote parent span is
 	// only referenced by setting ParentSpanID.
 	parent *Span
-	// isTransaction is true only for the root span of a local span tree. The
-	// root span is the first span started in a context. Note that a local root
-	// span may have a remote parent belonging to the same trace, therefore
-	// isTransaction depends on ctx and not on parent.
-	isTransaction bool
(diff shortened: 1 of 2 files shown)


tonyo:
Ah, scratch that. Just realized we already have the root span info in `spanRecorder`. Let me think this over again.

Issue #553, opened 2023-01-30
[OTel] Dynamic Sampling Context propagation does not work for nested spans


tonyo:
Extracted from #537.

### Problem

Dynamic Sampling Context propagation from nested spans doesn't work properly in OpenTelemetry integration.
Main reason: no publicly exposed mechanism to get the transaction (as in, a root span) for the given nested span. 
At the moment we're [doing this](https://github.com/getsentry/sentry-go/blob/2ece4821c5ab58cb10a385c8bbdc1270a5a0b432/otel/propagator.go#L45-L49), meaning that if we're doing a cross-service request from a nested span, the DSC won't be properly set.

### Potential Solutions

* Expose a method on Span that allows to get the root span (transaction) of the given span.

PR #558, opened 2023-01-30
feat(tracing): Add a method to get transaction for given span

tonyo:
Add a function to get the transaction span for the given span: `Span.GetTransaction()`
This is needed for a proper fix for #553.

~Things left to agree on: Function name~
~In Python, for example, there's a `containing_transaction` method on [Spans](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/tracing.py#L201-L208) and [Transactions](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/tracing.py#L641C9-L647).~


tracing.go
```diff
@@ -241,6 +241,26 @@ func (s *Span) IsTransaction() bool {
...
 
+// GetTransaction returns the root span (basically, a transaction) for the given span
+// and its local span tree.
+// FIXME(anton): Name -- TBC. In Python it's called "containing_transaction".
+// Caveat: there's already sentry.Transaction() function.
+func (s *Span) GetTransaction() *Span {
+	spanRecorder := s.spanRecorder()
+	if spanRecorder == nil {
+		// This probably means that the Span was created manually (not via
+		// StartTransaction or StartChild).
```
cleptric, line 252:
```suggestion
		// StartTransaction/StartSpan or StartChild).
```

cleptric (commented):
I think this is mostly OK for the time being.
No hard feelings about the name, it's `Span::getTransaction` in PHP as well.

Slightly OT, but I think this is a good example of why we should add proper abstractions for Transactions/Spans.
https://github.com/getsentry/sentry-php/blob/bfa3193944fe385af7d6b6b063962fc45956a817/src/Tracing/Span.php#L390-L406

I kinda like the way all parent properties are explicitly set on a child span.

cleptric (approved):

PR #559, opened 2023-01-31
fix(otel): Fix Dynamic Sampling context propagation for nested spans

### LINKED ISSUE #553: [OTel] Dynamic Sampling Context propagation does not work for nested spans

tonyo:
Thanks to #558, we can now fix the DSC/baggage propagation for non-transaction spans in the OTel propagator.

Fixes #553.

### DIFF
--- a/helpers_test.go
+++ b/helpers_test.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"reflect"
 	"testing"
+
+	"github.com/getsentry/sentry-go/internal/otel/baggage"
 )
 
 func assertEqual(t *testing.T, got, want interface{}, userMessage ...interface{}) {
@@ -49,4 +51,21 @@ func formatUnequalValues(got, want interface{}) string {
 	}
 
 	return fmt.Sprintf("\ngot: %s\nwant: %s", a, b)
+}
+
+func assertBaggageStringsEqual(t *testing.T, got, want string, userMessage ...interface{}) {
+	t.Helper()
+
+	baggageGot, err := baggage.Parse(got)
+	if err != nil {
+		t.Error(err)
+	}
+	baggageWant, err := baggage.Parse(want)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if !reflect.DeepEqual(baggageGot, baggageWant) {
+		logFailedAssertion(t, formatUnequalValues(got, want), userMessage...)
+	}
 }
--- a/otel/propagator.go
+++ b/otel/propagator.go
@@ -50,7 +50,7 @@ func (p sentryPropagator) Inject(ctx context.Context, carrier propagation.TextMa
 		containingTransaction := sentrySpan.GetTransaction()
 		dynamicSamplingContext := sentry.DynamicSamplingContextFromTransaction(containingTransaction)
 		containingTransaction.SetDynamicSamplingContext(dynamicSamplingContext)
-		sentryBaggageStr = dynamicSamplingContext.String()
+		sentryBaggageStr = containingTransaction.ToBaggage()
 	}
 	// FIXME: We're basically reparsing the header again, because internally in sentry-go
 	// we currently use a vendored version of "otel/baggage" package.
--- a/otel/propagator.go
+++ b/otel/propagator.go
@@ -46,7 +46,11 @@ func (p sentryPropagator) Inject(ctx context.Context, carrier propagation.TextMa
 	// Propagate baggage header
 	sentryBaggageStr := ""
 	if sentrySpan != nil {
-		sentryBaggageStr = sentrySpan.ToBaggage()
+		// Finalize/freeze the baggage
+		containingTransaction := sentrySpan.GetTransaction()
+		dynamicSamplingContext := sentry.DynamicSamplingContextFromTransaction(containingTransaction)
+		containingTransaction.SetDynamicSamplingContext(dynamicSamplingContext)
+		sentryBaggageStr = dynamicSamplingContext.String()
 	}
 	// FIXME: We're basically reparsing the header again, because internally in sentry-go
 	// we currently use a vendored version of "otel/baggage" package.
--- a/otel/propagator.go
+++ b/otel/propagator.go
@@ -46,11 +46,7 @@ func (p sentryPropagator) Inject(ctx context.Context, carrier propagation.TextMa
 	// Propagate baggage header
 	sentryBaggageStr := ""
 	if sentrySpan != nil {
-		// Dynamic sampling context is set only on the (root) transaction.
-		sentryTransaction := sentrySpan.GetTransaction()
-		if sentryTransaction != nil {
-			sentryBaggageStr = sentryTransaction.ToBaggage()
-		}
+		sentryBaggageStr = sentrySpan.ToBaggage()
 	}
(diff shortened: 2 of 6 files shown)


otel/propagator.go
```diff
@@ -45,10 +45,12 @@ func (p sentryPropagator) Inject(ctx context.Context, carrier propagation.TextMa
...
-	if sentrySpan != nil && sentrySpan.IsTransaction() {
-		// TODO(anton): Normally, this should return the DSC baggage from the transaction
-		// (not necessarily the current span). So this might break things in some cases.
-		sentryBaggageStr = sentrySpan.ToBaggage()
+	if sentrySpan != nil {
+		// Dynamic sampling context is set only on the (root) transaction.
+		sentryTransaction := sentrySpan.GetTransaction()
+		if sentryTransaction != nil {
+			sentryBaggageStr = sentryTransaction.ToBaggage()
+		}
```
tonyo, line 53: This is the actual fix.

cleptric: Tbh, I would actually fix `ToBaggage()`.

```go

// ToBaggage returns the serialized dynamic sampling context in the baggage format.
func (s *Span) ToBaggage() string {
	if s.IsTransaction() {
		return s.GetTransaction().dynamicSamplingContext.String()
	}

	return s.dynamicSamplingContext.String()
}
```

tracing.go
```diff
@@ -288,7 +288,10 @@ func (s *Span) ToSentryTrace() string {
 
 // ToBaggage returns the serialized dynamic sampling context in the baggage format.
 func (s *Span) ToBaggage() string {
-	return s.dynamicSamplingContext.String()
+	if containingTransaction := s.GetTransaction(); containingTransaction != nil {
+		return containingTransaction.dynamicSamplingContext.String()
+	}
+	return ""
```
tonyo, line 294: Moved the updated logic here.

tonyo:
Alright, found another issue: in the propagator's `Inject` we should actually generate/freeze the baggage.
It's getting more obvious that we should refactor/redo the DSC generation and freezing logic. I think we need a more explicit `FinalizeBaggage()` (or similar) method on the transaction object. 
I don't want to do it in this PR which already got messier than I would like.

cleptric:
What issue, in particular, you're seeing? Freezing the baggage during creation is possible today, what would a more explicit way improve?

helpers_test.go
```diff
@@ -50,3 +52,20 @@ func formatUnequalValues(got, want interface{}) string {
...
+	baggageGot, err := baggage.Parse(got)
+	if err != nil {
+		t.Error(err)
+	}
+	baggageWant, err := baggage.Parse(want)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if !reflect.DeepEqual(baggageGot, baggageWant) {
```
cleptric, line 68: I would always use https://pkg.go.dev/github.com/google/go-cmp/cmp, had quite a few issues with reflect in the past.

otel/propagator_test.go
```diff
@@ -28,37 +28,31 @@ type transactionTestContext struct {
...
-		Release:       "1.2.3",
-		EnableTracing: true,
-	})
-	hub := sentry.NewHub(client, sentry.NewScope())
-	ctx := sentry.SetHubOnContext(context.Background(), hub)
-
+func createTransactionAndMaybeSpan(transactionContext transactionTestContext, withSpan bool) trace.SpanContextConfig {
 	transaction := sentry.StartTransaction(
-		ctx,
+		emptyContextWithSentry(),
```
cleptric, line 33:
```suggestion
		emptyCtxWithSentry(),
```

otel/propagator_test.go
```diff
@@ -28,37 +28,31 @@ type transactionTestContext struct {
...
 	transaction.TraceID = TraceIDFromHex(transactionContext.traceID)
 	transaction.SpanID = SpanIDFromHex(transactionContext.spanID)
-	transaction.SetDynamicSamplingContext(sentry.DynamicSamplingContextFromTransaction(transaction))
 
-	sentrySpanMap.Set(trace.SpanID(transaction.SpanID), transaction)
 	if withSpan {
 		span := transaction.StartChild("op")
+		// We want the child to have the SpanID from transactionContext, so
+		// we "swap" span IDs from the transaction and the child span.
+		transaction.SpanID = span.SpanID
```
cleptric, line 44: Not sure I fully understand this.

cleptric (approved):
Minor asks/questions, else LGTM!

tonyo:
I see the following issues:
1. Given a transaction, there's no easy way now to generate the DSC, and then freeze it. We need to do stuff like 
    `transaction.SetDynamicSamplingContext(DynamicSamplingContextFromTransaction(transaction))`, which is cumbersome and error-prone
2. `Span.ToBaggage()` returns the serialized DSC, but doesn't generate it. Honestly, I don't see a good case of using it alone.

> Freezing the baggage during creation is possible today, what would a more explicit way improve?

Ideally we should have convenient ways of explicitly generating the DSC/baggage in a proper format (as baggage values, or as a string). I don't have a super clear picture yet how it will look like, though.

helpers_test.go
tonyo: Agreed, will include things like this in the planned test utils refactoring.

otel/propagator_test.go
tonyo: When `withSpan` is passed, we want to create a transation and a child span. We also want the child span to have the predefined SpanID (passed via `transactionContext`). It means that we have to update the SpanID for the transaction. We can generate a new random one, or we can just take the already-generated one from the child span (`span`), that will be replaced with `SpanIDFromHex(transactionContext.spanID)` in the next line anyway.

cleptric:
The DSC APIs are something I want to avoid to expose to publicly, as it has lots of nitty-gritty details. Taking some extra steps for Otel is fine IMO.

> Span.ToBaggage() returns the serialized DSC, but doesn't generate it. Honestly, I don't see a good case of using it alone.

You would use this function to get the value of the header string for outgoing HTTP requests.
Here is a real-world example where I'm using it: https://github.com/getsentry/gib-potato/blob/e7eda439ad93517a27f7c3201d838b769c766587/potal/httpclient.go#L37

otel/propagator_test.go
tonyo: It's a helper function that's used in multiple places already, do we really want this?

cleptric: Up to you, no strong feelings.

cleptric:
You are right, we actually need to create a DSC when calling `ToBaggage` if there is none. This is the case if we're the head SDK. Great catch!

tonyo:
Cool, will think about how to improve this (as a separate task)

PR #562, opened 2023-02-01
ref(test): Move assert* helpers to internal package

tonyo:
Here we move all assert* helper functions from the root and "otel" modules to a new `internal/testutils` package.
Also: add AssertTrue and AssertFalse helpers.

### DIFF
--- a/internal/testutils/asserts.go
+++ b/internal/testutils/asserts.go
@@ -1,0 +1,52 @@
+package testutils
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+)
+
+func AssertEqual(t *testing.T, got, want interface{}, userMessage ...interface{}) {
+	t.Helper()
+
+	if !reflect.DeepEqual(got, want) {
+		logFailedAssertion(t, formatUnequalValues(got, want), userMessage...)
+	}
+}
+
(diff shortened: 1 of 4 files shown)


internal/testutils/asserts.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+func AssertEqual(t *testing.T, got, want interface{}, userMessage ...interface{}) {
+	t.Helper()
+
+	// Ideally, we would switch to cmp.Diff. However, in a general case, cmp.Diff
+	// is not able to compare structs with unexported (private) fields by default. There
+	// are ways to override modify that behaviour (e.g. by passing AllowUnexported), but
+	// it significantly complicates its usage.
+	if !reflect.DeepEqual(got, want) {
```
cleptric, line 19: Let's have a quick offline discussion about this, as I would like us to stick to either or 😄

cleptric (approved):

internal/testutils/asserts.go
tonyo: Yeah, it's a rabbit hole :/
Maybe you'll have some ideas, but at the moment it doesn't look very feasible to use `go-cmp` only.

PR #564, opened 2023-02-02
feat: add support for exception mechanism metadata

wyattanderson:
This adds support for the exception `mechanism` field, based on the [`relay` schema of the field](https://github.com/getsentry/relay/blob/e785f9a22adef018cec8785a7dec289a6ae74165/relay-general/src/protocol/mechanism.rs#L94-L159).

Primarily, I'd like to be able to annotate exceptions with this data to make the UI more useful. For instance, we have a gRPC interceptor that recovers from `panic`, and I think it would be useful to set the `handled` bit to `false` so that we get the additional UI embellishment (the red skull and "Unhandled" flag) when endpoints panic (versus returning an explicit status or error). This would likely be useful to other HTTP integrations as well.

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -204,6 +204,22 @@ func NewRequest(r *http.Request) *Request {
 	}
 }
 
+// Mechanism is the mechanism by which an exception was generated and handled.
+type Mechanism struct {
+	Type        string                 `json:"type,omitempty"`
+	Description string                 `json:"description,omitempty"`
+	HelpLink    string                 `json:"help_link,omitempty"`
+	Handled     *bool                  `json:"handled,omitempty"`
+	Data        map[string]interface{} `json:"data,omitempty"`
+}
+
+// SetUnhandled indicates that the exception is an unhandled exception, i.e.
+// from a panic.
+func (m *Mechanism) SetUnhandled() {
+	h := false
+	m.Handled = &h
+}
+
 // Exception specifies an error that occurred.
 type Exception struct {
 	Type       string      `json:"type,omitempty"`  // used as the main issue title
@@ -211,6 +227,7 @@ type Exception struct {
 	Module     string      `json:"module,omitempty"`
 	ThreadID   string      `json:"thread_id,omitempty"`
 	Stacktrace *Stacktrace `json:"stacktrace,omitempty"`
+	Mechanism  *Mechanism  `json:"mechanism,omitempty"`
 }
 
 // SDKMetaData is a struct to stash data which is needed at some point in the SDK's event processing pipeline
--- a/interfaces_test.go
+++ b/interfaces_test.go
@@ -160,6 +160,55 @@ func TestEventMarshalJSON(t *testing.T) {
 	}
 }
 
+func TestMechanismMarshalJSON(t *testing.T) {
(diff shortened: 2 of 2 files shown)


cleptric (commented):
`meta` is missing https://github.com/getsentry/relay/blob/e785f9a22adef018cec8785a7dec289a6ae74165/relay-general/src/protocol/mechanism.rs#L154

There are also some docs that might explain all of this in more detail https://develop.sentry.dev/sdk/event-payloads/exception/

wyattanderson:
> `meta` is missing https://github.com/getsentry/relay/blob/e785f9a22adef018cec8785a7dec289a6ae74165/relay-general/src/protocol/mechanism.rs#L154
> 
> There are also some docs that might explain all of this in more detail https://develop.sentry.dev/sdk/event-payloads/exception/

Ah, yeah. That was an intentional omission because it didn't look like any of the `meta` fields would be relevant in a Go runtime, since it seems like most of the `meta` fields are specific to Apple, POSIX, or C syscalls. I'm not opposed to adding the relevant structs if you think they're worth adding, or I could leave a comment explaining the omission.

cleptric:
Ok, we can also start with `type`, `handled`, and `data` for now. Not sure if the others will add much value.
Some tests would be great as well 🙂

wyattanderson:
> Ok, we can also start with `type`, `handled`, and `data` for now. Not sure if the others will add much value. Some tests would be great as well 🙂

I'll add some tests.

Are you opposed to keeping `help_link` and `description`? At least for my own use cases, I could see some value in being able to populate those (since they're rendered in the UI) for common internal errors so that we could point to internal documentation, etc.

interfaces.go
```diff
@@ -204,13 +204,30 @@ func NewRequest(r *http.Request) *Request {
...
+	Handled     *bool                  `json:"handled,omitempty"`
+	Data        map[string]interface{} `json:"data,omitempty"`
+}
+
+// SetUnhandled indicates that the exception is an unhandled exception, i.e.
+// from a panic.
+func (m *Mechanism) SetUnhandled() {
+	h := false
+	m.Handled = &h
+}
```
wyattanderson, line 221: I wanted to add some way to make it easier to set the `Handled` field, since it's a bit annoying otherwise. AWS SDKs, for instance, have utility functions like `func Bool(b bool) *bool`, but that felt like overkill here. Adding a corresponding `SetHandled()` also felt like it wouldn't be useful. Happy to go either (or any other) way, though.

wyattanderson:
@cleptric updated with tests and a comment.

wyattanderson:
> Could you add a small snippet of how someone can use this feature to set an error as unhandled as well?

Something beyond the tests? I can add it if you can point me to where you'd like me to add it. Documentation? Nothing else in `interfaces.go` has anything like this, so I'm not quite sure what you're asking for. If you could be specific, that would be very helpful. I'd like to get this merged so we can avoid the need to maintain an internal fork of this package to take full advantage of Sentry's data model.

cleptric (approved):
Merging for now. @wyattanderson would still be cool if you could comment with a snippet of how you use it in your application. Thanks for the contribution!

PR #568, opened 2023-02-03
fix(baggage): Update baggage parsing and encoding in vendored otel package

### LINKED ISSUE #554: [OTel] Baggage parsed incorrectly if it contains encoded whitespaces ("%20")
Extracted from #537.

### Problem

Baggage header parsing raises an error (`invalid value`) when we try to parse values like `sentry-transaction=HTTP%20GET`, which essentially breaks the entire Dynamic Sampling Context propagation.
Opened an issue in `opentelemetry-go`: https://github.com/open-telemetry/opentelemetry-go/issues/3601 

### Proposed Solution

We might want to use our vendored implementation and fix the bug there.

tonyo:
These changes to `baggage.go` fix a few issues that are happening now in violation of [the baggage spec](https://www.w3.org/TR/baggage/) (or at least my understanding of it):

1. Baggage string value `one%20two` should be properly parsed as "one two"
2. Baggage string value `one+two` should be parsed as "one+two"
3. Go string value "one two" should be encoded as `one%20two` (percent encoding), and NOT as `one+two` (URL query encoding).
4. Go string value "1=1" might be encoded as `1=1` ([the spec says:](https://www.w3.org/TR/baggage/#value) "Note, value MAY contain any number of the equal sign (=) characters. Parsers MUST NOT assume that the equal sign is only used to separate key and value."). `1%3D1` is also valid, but to simplify the implementation we're not doing it.

### Notes

* [The original `baggage_test.go`](https://github.com/open-telemetry/opentelemetry-go/blob/5e8eb855bf3c6507715edd12ded5c6a950dd6104/baggage/baggage_test.go) was copied from [opentelemetry-go](https://github.com/open-telemetry/opentelemetry-go), commit 5e8eb855bf3c6507715edd12ded5c6a950dd6104. 
* ~Case number 2 is not fully fixed right now: baggage string value `one+two` will be parsed as "one two", and not as "one+two". To fix that, `url.QueryUnescape()` should be replaced by a function with the logic compatible with `percentEncodeValue()`, where the plus (and other allowed characters) are not treated as something special.~

Fixes #554.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/pingcap/errors v0.11.4
 	github.com/pkg/errors v0.9.1
 	github.com/sirupsen/logrus v1.9.0
+	github.com/stretchr/testify v1.8.0
 	github.com/urfave/negroni v1.0.0
 	github.com/valyala/fasthttp v1.40.0
 	golang.org/x/sys v0.0.0-20220928140112-f11e5e49a4ec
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/pingcap/errors v0.11.4
 	github.com/pkg/errors v0.9.1
 	github.com/sirupsen/logrus v1.9.0
+	github.com/stretchr/testify v1.8.0
```
tonyo, line 15: `testify` is currently used only in `baggage_test.go`, and also used by a bunch of dependencies.

PR #577, opened 2023-02-07
fix(ci): Run "build", "vet", and "mod tidy" for all submodules

tonyo:
In the new module/submodule world, it's not enough to do e.g. `go build ./...`, since it'll ignore submodules.


cleptric (approved):
Alle the Makefiles :shipit:

PR #579, opened 2023-02-10
fix(logrus-hook): Don't throw error if sentry event is dropped

### LINKED ISSUE #578: Logrus hook prints error if event has been dropped
When using the logrus hook to send logs to sentry, the following line is printed everytime the hook is triggered but the event is dropped (e.g. by using `BeforeSend`):

```
Failed to fire hook: failed to send to sentry
```

The log is triggered by [this particular error](<https://github.com/getsentry/sentry-go/blob/f03b31a4a314fda36229c131a20306ffd686f2e9/logrus/logrusentry.go#L124>).<br>I think this error should not be thrown at all since it's a valid scenario to drop sentry events client-side.

hnicke:
When using the logrus hook to send logrus logs to sentry, an error was logged when the corresponding sentry event is being dropped client-side. This happens for example when using the `BeforeSend` event transformation function.

This should not be considered an error.

Resolves #578

### DIFF
--- a/logrus/logrusentry.go
+++ b/logrus/logrusentry.go
@@ -121,7 +121,6 @@ func (h *Hook) Fire(entry *logrus.Entry) error {
 		if h.fallback != nil {
 			return h.fallback(entry)
 		}
-		return errors.New("failed to send to sentry")
 	}
 	return nil
 }


logrus/logrusentry.go
```diff
@@ -121,7 +121,6 @@ func (h *Hook) Fire(entry *logrus.Entry) error {
 		if h.fallback != nil {
 			return h.fallback(entry)
 		}
-		return errors.New("failed to send to sentry")
```
cleptric, line 124: I'm questioning the usefulness of handling the return value of `CaptureEvent` altogether. By default, the ID is always returned, even if the `HTTPTransport` does not successfully submit the event to Sentry, as we do not have a back channel to the main thread. So this only applies when using the `HTTPSyncTransport`.

hnicke:
I have figured out that the error can be suppressed by setting a fallback:

```go
hook.SetFallback(func(entry *log.Entry) error {
	return nil
})
```

Given this solution the PR is not needed anymore.

PR #580, opened 2023-02-15
fix(otel): Overwrite "trace" context in Sentry error

tonyo:
This patches the following issue: when [creating a transaction in the span processor](https://github.com/getsentry/sentry-go/blob/f03b31a4a314fda36229c131a20306ffd686f2e9/otel/span_processor.go#L50-L56), we later overwrite its SpanID/TraceID. The problem is, the trace context is also set on the scope when the span is started: https://github.com/getsentry/sentry-go/blob/f03b31a4a314fda36229c131a20306ffd686f2e9/tracing.go#L176

So ultimately we need to either/or:
1. Pass additional SpanOptions to StartTransaction that set the right SpanID/TraceID right away
2. Overwrite the trace context later.

Here I'm taking approach 2 because it's easier, but generally speaking I believe we should review setting trace context (and a few other tracing related things) on the scope.

### DIFF
--- a/otel/event_processor.go
+++ b/otel/event_processor.go
@@ -12,6 +12,8 @@ import (
 //
 // Caveat: hint.Context should contain a valid context populated by OpenTelemetry's span context.
 func linkTraceContextToErrorEvent(event *sentry.Event, hint *sentry.EventHint) *sentry.Event {
+	sentry.Logger.Printf("\nEvent Processor\nEvent: %#v\nHint: %#v", event, hint)
+
 	if hint == nil || hint.Context == nil {
 		return event
 	}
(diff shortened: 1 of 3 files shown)


cleptric (commented):
https://github.com/getsentry/sentry-go/blob/master/scope.go#L359-L374

Let's remove the logic in `StartSpan` and set the `trace` context in `scope.ApplyToEvent()` instead.

tonyo:
@cleptric Not entirely sure what you suggest, but it sounds larger/riskier than what was initially planned for this PR. Can you open an issue and explain your idea there please? 
And this PR will at least fix things for specific OTel use case.

Issue #596, opened 2023-03-07
OTel Context Missing


smeubank:
## Summary
Events sent to Sentry which were converted from OTel spans, should have an OTel context with attributes and other data sourced from OTel SDKs. It appears that some events do not have this. 

[Python event with context](https://sentry-sdks.sentry.io/performance/otel-demo-recommendationservice:0394771c82c54ec3890da6659659ae9b/?project=4504259429859328&query=&referrer=performance-transaction-summary&statsPeriod=14d&transaction=%2Fhipstershop.RecommendationService%2FListRecommendations&unselectedSeries=p100%28%29#context-otel)

<img width="845" alt="image" src="https://user-images.githubusercontent.com/47563310/223389391-71fe745f-19e6-4ca7-acd7-f0c68b01ec91.png">

[Go event without context](https://sentry-sdks.sentry.io/performance/otel-demo-accountingservice:274d0f0e4cbb4813995dd4234394d160/?project=4504536076582912&query=&referrer=performance-transaction-summary&statsPeriod=14d&transaction=orders+receive&unselectedSeries=p100%28%29)

Develop Spec: [Add OpenTelemetry Context](https://develop.sentry.dev/sdk/performance/opentelemetry/#step-5-add-opentelemetry-context)

## Expected Behavior
OTel SDK data populated in the OTel context and appearing in the product on the transaction detail page

### SDK
 - `sentry-go` with OTel Go SDKs

AbhiPrasad:
Why did we re-open?

cleptric:
So this is about "OTel Context Missing", and as we currently always get an empty Hub here https://github.com/getsentry/sentry-go/blob/master/otel/span_processor.go#L120-L123, so even with #599 being merged, it won't work just yet.

AbhiPrasad:
Alright I will push up a PR then!

PR #599, opened 2023-03-08
feat(tracing): Introduce span.SetContext

AbhiPrasad:
ref https://github.com/getsentry/sentry-go/issues/596

This PR adds the `SetContext` struct method to the span struct.

This allows us to remove the usage of setting context on the scope in the OpenTelemetry span processor

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -52,6 +52,8 @@ type Span struct { //nolint: maligned // prefer readability over optimal memory 
 	isTransaction bool
 	// recorder stores all spans in a transaction. Guaranteed to be non-nil.
 	recorder *spanRecorder
+	// span context, can only be set on transactions
+	contexts map[string]Context
 }
 
 // TraceParentContext describes the context of a (remote) parent span.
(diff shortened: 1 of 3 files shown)


tracing.go
```diff
@@ -484,17 +493,21 @@ func (s *Span) toEvent() *Event {
 		s.dynamicSamplingContext = DynamicSamplingContextFromTransaction(s)
 	}
 
+	contexts := map[string]Context{}
+	for k, v := range s.contexts {
+		contexts[k] = v
+	}
+	contexts["trace"] = s.traceContext().Map()
```
cleptric, line 500: This might overwrite the `trace` key, is this a concern?

tracing.go
```diff
@@ -484,17 +493,21 @@ func (s *Span) toEvent() *Event {
 		s.dynamicSamplingContext = DynamicSamplingContextFromTransaction(s)
 	}
 
+	contexts := map[string]Context{}
+	for k, v := range s.contexts {
+		contexts[k] = v
```
cleptric, line 498: Maybe worth adding a test here as well.

cleptric (commented):
Added some comments. We need to find a solution for https://github.com/getsentry/sentry-go/blob/master/otel/span_processor.go#L148 as well, as the initial issue with not having any hotel context being present will persit.

cleptric:
Might be worth adding a new field to `Span` which includes the transaction name.

AbhiPrasad:
> Might be worth adding a new field to `Span` which includes the transaction name.

Let's do this in a follow up PR?

tracing.go
AbhiPrasad: Nope, we want to make sure the `trace` key is always being set by the transaction itself, not the span.

cleptric (approved):

PR #605, opened 2023-03-21
fix: Introduce a new Span.Name field

cleptric:


pbennett:
So how do we override the transaction name in later calls?
I do this currently:
```
sentry.GetHubFromContext(ctx).Scope().SetTransaction("xxxx")
```

This is no longer possible.

cleptric:
```go
transaction := sentry.TransactionFromContext(ctx)
if transaction != nil {
    transaction.Name = "xxxx"
}
```

jerbob92:
How would this work for other/error events?  We automatically create and inject a Hub per request and then do:

```
		hub.ConfigureScope(func(scope *sentry.Scope) {
			scope.SetTransaction(request_info.Id)
		})
```

This would then automatically add the transaction to every event coming from this scope using the `ApplyToEvent` method, but the lines that added it are gone:
```
	if scope.transaction != "" {
		event.Transaction = scope.transaction
	}
```

And it doesn't look an alternative has been added?

cleptric:
@jerbob92 So, `transaction` on the event has a long history, even before Sentry had a performance product.
There might be some legit use cases for some users that only use our error product, but in your case, a tag would give you the same UX in the product. Tags are indexed, hence searchable and appear in the UI in the same place as `transaction` did.

pbennett:
Perhaps updating the documentation and SDKs are in order because frankly, transactions / spans / tags / things that show as 'database' events are all super obscure and vary from sdk to sdk.

Issue #291, opened 2020-10-30
Export function that builds error chain from wrapped errors


avgeorge:
## Summary

Please export a function that can build an error chain from a wrapped error.

## Motivation

This package has logic that builds an error chain for an event from errors that are wrapped using several different error packages. However, this logic is in a function that builds an event and submits it immediately. I would like to be able to use this logic so that I can make further changes to the event before submitting it.

## Additional Context

The simplest way to achieve this would be to export [Client.eventFromException](https://github.com/getsentry/sentry-go/blob/6ba7f9cc0855829323615e9ba57fb0eb13450cc3/client.go#L368) which builds an event from an error. The function doesn't depend on the client receiver so it's trivial to change it to a standalone function.
It might be better to create a function with a different signature though, e.g. `func ExtractErrorChain(err error) []Exception`

rhcarvalho:
> I would like to be able to use this logic so that I can make further changes to the event before submitting it.

Have you tried `BeforeSend` (global) or using an `EventProcessor` (if you need per-scope customization)? Those are the standard ways to inspect and modify events before sending, including the possibility to drop the event altogether.

- https://pkg.go.dev/github.com/getsentry/sentry-go#ClientOptions
- https://docs.sentry.io/platforms/go/configuration/filtering/#using-beforesend
- https://pkg.go.dev/github.com/getsentry/sentry-go#Scope.AddEventProcessor

avgeorge:
Thanks, I think the Scope.AddEventProcessor should allow me to do what I need.

I want to build a sentry hook for logrus so that errors are sent to sentry automatically when they are logged from logrus. A logrus error log would include the error (or error chain), a message and any additional context as key value pairs. I can set the additional data on the scope, but Hub.CaptureException doesn't allow adding a message. I think I can use an EventProcessor to set the message, but it seems a bit overkill to use it just to set a field I already had when the event was created. So I would still prefer access to this function if possible.

cottrellio:
^ +1 I want logging with the error chain, a custom message, and contextual data

KevinSnyderCodes:
Can we reopen this issue? I am trying to accomplish the same as [here](https://github.com/getsentry/sentry-go/issues/291#issuecomment-719456516): create a custom logger that automatically sends events to Sentry in addition to logging them.

[...]

PR #607, opened 2023-03-22
feat: Export client.EventFromMessage, client.EventFromException, and event.SetException

cleptric:
Export the existing `client.EventFromMessage`, `client.EventFromException`, and the new `event.SetException` method to facilitate logging integrations.

refs https://github.com/getsentry/sentry-go/issues/291#issuecomment-1478633701


KevinSnyderCodes:
https://github.com/getsentry/sentry-go/pull/608

This adds a new method `Event.SetException()` which decouples the `Event` creation and the setting of its `Exception` field. `Client.EventFromException()` now uses `Event.SetException()` under the hood.

This allows you to populate an `Event` with an exception and other data separately from the `Client` before calling `Client.CaptureEvent()`.

tonyo:
@cleptric merged https://github.com/getsentry/sentry-go/pull/608 into this PR (thanks @KevinSnyderCodes), and added a few docstrings. Should be good to go, I think 👍

tonyo (approved):

PR #614, opened 2023-03-30
doc: Update attributions for vendored "baggage"

tonyo:
"internal/otel/baggage" package was initially vendored from opentelemetry-go, and then later we made some updates to it. According to our new attribution policy, adding some links to the original files (copyright notes are already there).

Ref https://github.com/getsentry/team-webplatform-meta/issues/42, https://github.com/getsentry/team-ospo/issues/120

AFAIK this is the only place where we need to attribute 3rd party code.


tonyo:
cc @chadwhitacre

chadwhitacre (approved):

PR #631, opened 2023-05-04
feat: Remove store endpoint

cleptric:
To move forward on initiatives like https://github.com/getsentry/team-replay/issues/70 and https://github.com/getsentry/team-webplatform-meta/issues/45, we decided to remove the `/store` endpoint from the Go SDK.

These changes require a self-hosted version of Sentry 20.6.0 or higher. If you are using a version of [self-hosted Sentry](https://develop.sentry.dev/self-hosted/) (aka onpremise) older than 20.6.0 then you will need to [upgrade](https://develop.sentry.dev/self-hosted/releases/).

I removed quite a few things from `logrusentry_test.go`.
My reasoning is that we mostly care about testing `entryToEvent`, as all other functionality is wrapped into `CaptureEvent`. Making the tests work again would require copying a lot of functionality from `client_test.go`, something I'm not keen on maintaining.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -94,7 +94,7 @@ func getRequestBodyFromEvent(event *Event) []byte {
 	return nil
 }
 
-func transactionEnvelopeFromBody(event *Event, dsn *Dsn, sentAt time.Time, body json.RawMessage) (*bytes.Buffer, error) {
+func envelopeFromBody(event *Event, dsn *Dsn, sentAt time.Time, body json.RawMessage) (*bytes.Buffer, error) {
 	var b bytes.Buffer
 	enc := json.NewEncoder(&b)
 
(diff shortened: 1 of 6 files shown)


transport.go
```diff
@@ -127,13 +127,23 @@ func transactionEnvelopeFromBody(event *Event, dsn *Dsn, sentAt time.Time, body
...
-		Length int    `json:"length"`
+		Type        string `json:"type"`
+		Length      int    `json:"length"`
+		ContentType string `json:"content_type"`
 	}{
-		Type:   transactionType,
-		Length: len(body),
+		Type:        itemType,
+		Length:      len(body),
+		ContentType: "application/json",
```
tonyo, line 146: The ContentType part is new, is it obligatory now?

transport.go
```diff
@@ -157,21 +167,14 @@ func getRequestFromEvent(event *Event, dsn *Dsn) (r *http.Request, err error) {
...
-		b, err := transactionEnvelopeFromBody(event, dsn, time.Now(), body)
-		if err != nil {
-			return nil, err
-		}
-		return http.NewRequest(
-			http.MethodPost,
-			dsn.EnvelopeAPIURL().String(),
-			b,
-		)
+	b, err := envelopeFromBody(event, dsn, time.Now(), body)
```
tonyo, line 170: Let's rename `b` to something more human-friendly

tonyo (commented):
Looks good so far

transport.go
cleptric: Not required, same as `length`. But as pointed out in most examples on https://develop.sentry.dev/sdk/envelopes/, I opted to include it.

transport.go
cleptric: Don't make the Go gods angry 😬
Renamed to `envelope`.

transport.go
tonyo: Interestingly, the docs only cover the field in the context of attachments:

```
content_type
    String, optional. The content type of the attachment payload. Any [MIME type](https://www.iana.org/assignments/media-types/media-types.xhtml) may be used; the default is application/octet-stream.
```

Haven't checked, but we might be ignoring the content-type field for non-attachment payloads.

cleptric: I opted to remove it again.

tonyo (approved):

PR #632, opened 2023-05-04
fix: Improve isSentryRequestUrl

cleptric:
If we can't fetch a hub from the `ctx`, we now fallback to the global hub.
I also added the project id to not solely rely on the hostname.

### DIFF
--- a/otel/internal/utils/sentryrequest.go
+++ b/otel/internal/utils/sentryrequest.go
@@ -27,7 +27,10 @@ func IsSentryRequestSpan(ctx context.Context, s trace.ReadOnlySpan) bool {
 func isSentryRequestUrl(ctx context.Context, url string) bool {
 	hub := sentry.GetHubFromContext(ctx)
 	if hub == nil {
-		return false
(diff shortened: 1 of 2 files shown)


otel/internal/utils/sentryrequest.go
```diff
@@ -27,7 +27,10 @@ func IsSentryRequestSpan(ctx context.Context, s trace.ReadOnlySpan) bool {
 func isSentryRequestUrl(ctx context.Context, url string) bool {
 	hub := sentry.GetHubFromContext(ctx)
 	if hub == nil {
-		return false
+		hub = sentry.CurrentHub()
```
tonyo, line 30: Note: for better experience and to fix more cases like this, we might want to set the Hub on context a lot earlier, maybe even in StartSpan.

cleptric: We might not know which hub to set on the `ctx` in `StartSpan`.
I think it might be worth digging a little bit into how Otel handles go routines.
For now, we might still consider this as a quick win.

Issue #506, opened 2022-11-30
Default behavior of sentry-go silently swallows panics


claytoneast:
## Summary
The sentry-go http library by default recovers panics. The relevant library code is here, https://github.com/getsentry/sentry-go/blob/v0.13.0/http/sentryhttp.go#L107

It does this because it is an error reporting SDK, and thus it is assumed that the error has gone to Sentry and is visible there, but this is obviously untrue in environments where Sentry is not actually phoning home, such as test or development. Thus, for all intents and purposes, anything panic-y in non-production environments is silently swallowed.

I'm filing this as a bug because it seems counterproductive that the default behavior is "We will silently hide from you the most catastrophic kind of error you may encounter if you aren't in a production environment and checking sentry."

I realize that there is a `Repanic` option. I do not think Repanic should be false by default, since if you don't already know Sentry swallows the error, and that Repanic exists, you will spend hours trying to figure out what is happening in your program.

## Steps To Reproduce
Our sentry setup is very basic.

```go
// Sentry configures Sentry error logging then passes control to the next http.Handler
func Sentry(next http.Handler) http.Handler {
	sentryHandler := sentryhttp.New(sentryhttp.Options{}) // Yes, I realize that Repanic goes here

	return sentryHandler.Handle(next)
}

```

## Expected Behavior
I would expect the worst possible kind of error in my Go programs to be visible to me, the person writing the program, when I'm developing the program, and encountering said worst possible kinds of errors.

### SDK
 - `sentry-go` version: 0.13.0
 - Go version:  1.18
 - Using Go Modules? [yes/no] yes

### Sentry
 - Using hosted Sentry in sentry.io? Yes

cleptric:
Thanks for writing in.

I do agree that the UX is not great, especially in a local environment.
Our other SDKs normally NOP if you do not supply a DSN, which is the case in most local envs. I'm wondering if our integrations should have the same behavior.

Changing the default will be a big breaking change, something I would rather do in `1.0.0`.

claytoneast:
I think that would be a great thing to do (noop w/o DSN) 👍

cleptric:
So to confirm, you did not have a DSN set when encountering this locally?

claytoneast:
Yep, DSN is unset in the program locally.

cleptric:
As this is a breaking change, we won't do this prior 1.0. I also dislike this behaviour, but we'll keep it for now.

PR #635, opened 2023-05-05
feat: add NewDefault to Gin

codenoid:
I just found a scary (default) behavior

```
package main

import (
	"fmt"
	"net/http"

	sentrygin "github.com/getsentry/sentry-go/gin"
	"github.com/gin-gonic/gin"
)

func main() {
	r := gin.Default()
	r.Use(sentrygin.New(sentrygin.Options{}))
	r.Use(func(c *gin.Context) {
		panic(2)
	})
	r.GET("/", func(c *gin.Context) {
		c.JSON(http.StatusOK, gin.H{
			"message": "pong",
		})
	})
	r.Run("127.0.0.1:8080")
}
```

As a newcomer, I expected the `/` route to not be called when a panic occurred inside the middleware (auth middleware). However, due to Sentry's default configuration, the route continued to be called. This behavior made the c.Abort() calls inside the middleware useless, as they were no longer being used

I suggest safer function for a new comer

### DIFF
--- a/gin/sentrygin.go
+++ b/gin/sentrygin.go
@@ -46,6 +46,17 @@ func New(options Options) gin.HandlerFunc {
 	}).handle
 }
 
+// NewDefault uses optimized option and returns a function that satisfies gin.HandlerFunc interface
+// It can be used with Use() methods.
+func NewDefault() gin.HandlerFunc {
+	defaultOptions := Options{
+		Repanic:         true,
+		Timeout:         2 * time.Second,
+		WaitForDelivery: false,
+	}
+	return New(defaultOptions)
+}
+
 func (h *handler) handle(ctx *gin.Context) {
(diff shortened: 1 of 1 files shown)


codenoid:
on my case, this line of code should be break the world
image

And I'm sure I'm not the only one who having this kind of code, because I took this part of code from [StackOverflow Answer](https://stackoverflow.com/a/39460624)

cleptric:
Thanks for your suggestion!

I'm not in favour of increasing the API surface for this.
We should instead change the default. We're tracking this in https://github.com/getsentry/sentry-go/issues/506, which will apply to all integrations.

codenoid:
do you think I still have a chance to contribute on changing default behavior? could you assign me a task (if there is) ?

PR #636, opened 2023-05-07
ref: Improve isSentryRequestUrl

cleptric:
This is a follow-up on #632, to add better error handling as well as a better comparison of the span attribute URL.

Has to be merged after #631.


otel/internal/utils/sentryrequest.go
```diff
@@ -40,4 +40,10 @@ func isSentryRequestUrl(ctx context.Context, url string) bool {
 
 	dsn, _ := sentry.NewDsn(client.Options().Dsn)
 	return strings.Contains(url, dsn.GetHost()) && strings.Contains(url, dsn.GetProjectID())
```
tonyo, line 42: Probably you meant:
```suggestion
```

tonyo (changes requested):

tonyo (approved):

Issue #588, opened 2023-02-24
Invalid transaction source: echo


jjbayer:
## Summary
Relay is reporting an invalid transaction source emitted by the Go SDK version 0.17.0. If this was fixed in >= 0.18.0, feel free to close this issue.

```json
"sdk": {
    "name": "sentry.go",
    "version": "0.17.0",
    "integrations": [
        "ContextifyFrames",
        "Environment",
        "IgnoreErrors",
        "Modules"
    ],
    "packages": [
        {
            "name": "sentry-go",
            "version": "0.17.0"
        }
    ]
},
"transaction_info": {
    "source": "echo"
}
```

## Steps To Reproduce
Go to https://sentry.my.sentry.io/organizations/sentry/issues/391873, and use the `project` tag to find example events in that project.

## Expected Behavior
The SDK sends one of the transaction sources listed in 
https://develop.sentry.dev/sdk/event-payloads/transaction/#transaction-annotations.

### SDK
See JSON snippet

### Sentry
 - Using hosted Sentry in sentry.io? yes

tonyo:
Looks like we should add additional guards in places where we set transaction sources from external inputs, e.g.
* https://github.com/getsentry/sentry-go/blob/master/tracing.go#L774

cleptric:
We would also need to not export the `Span.Source` field then. Adding a getter/setter seems reasonable.

Issue #511, opened 2022-12-06
Set `transaction_info.source` on `scope.SetTransaction` call


cleptric:
When calling `span.SetTransaction()`, we should update the `Source` to `SourceCustom`.

tonyo:
`SetTransaction` is removed anyway, but here's also an improvement to make sure that we always have a proper transaction source: https://github.com/getsentry/sentry-go/pull/637

PR #637, opened 2023-05-09
fix(tracing): Make sure a valid source is set on a transaction event

tonyo:
Transaction source should be one of the predefined values as described by the envelopes protocol: https://develop.sentry.dev/sdk/event-payloads/transaction/#transaction-annotations

The safest way to do for Go seems to be checking the source in `Span.toEvent()`: we don't currently have setters for `Span.Name` and `Span.Source`, so adjusting the source earlier may be not enough.

Fixes https://github.com/getsentry/sentry-go/issues/588
Related to https://github.com/getsentry/sentry-go/issues/511

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -626,7 +626,7 @@ const (
 	SourceTask      TransactionSource = "task"
 )
 
-// A set of all valid transaction sources
+// A set of all valid transaction sources.
 var allTransactionSources = map[TransactionSource]struct{}{
 	SourceCustom:    {},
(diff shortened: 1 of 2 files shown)


tonyo:
Now I'm looking at this again, lemme think a bit more about making Span.Source private 🤔

tonyo:
Decided not to touch `Span.Source` for now: we should probably couple that with overall improvements around settings/getting transaction Source and Name.

cleptric (approved):

Issue #570, opened 2023-02-06
"fatal error: concurrent map writes" when submitting transaction


SupaHam:
## Summary
When `span.Finish` is called, there's a chance a fatal concurrent write could occur due to the forceful setting of device parameters `arch` and `cpu`.

## Steps To Reproduce
1. Set "device" context on scope concurrently using `scope.SetContext` 
2. In parallel, create and Finish a span as is.
3. `fatal error: concurrent map writes`

```go
scope.SetContext("device", map[string]interface{}{
  "name": model
})
```

## Expected Behavior
Don't crash the go application because integrations have no concept of context safety.

### SDK
 - `sentry-go` version: v0.16.0
 - Go version: 1.18
 - Using Go Modules? yes

### Sentry
 - Using hosted Sentry in sentry.io? yes
 - Using your own Sentry installation? Version: 
 - Anything particular to your environment that could be related to this issue?

## Additional context

```fatal error: concurrent map writes

goroutine 140293796 [running]:
runtime.throw({0x2935f76?, 0x24b8a60?})
        /usr/local/go/src/runtime/panic.go:992 +0x71 fp=0xc01795ddc0 sp=0xc01795dd90 pc=0x43bcb1
runtime.mapassign_faststr(0x24b8640, 0xc0297caff0, {0x2909cb4, 0x4})
        /usr/local/go/src/runtime/map_faststr.go:295 +0x38b fp=0xc01795de28 sp=0xc01795ddc0 pc=0x416acb
github.com/getsentry/sentry-go.(*environmentIntegration).processor(0x0?, 0xc009852a00, 0xc01795dd38?)
        /build/.go/pkg/mod/github.com/getsentry/sentry-go@v0.16.0/integrations.go:89 +0x21a fp=0xc01795ded0 sp=0xc01795de28 pc=0x8d62ba
github.com/getsentry/sentry-go.(*environmentIntegration).processor-fm(0xc006d61560?, 0xc009852a00?)
        <autogenerated>:1 +0x2f fp=0xc01795def8 sp=0xc01795ded0 pc=0x8e5b6f
github.com/getsentry/sentry-go.(*Client).prepareEvent(0xc000d6e000, 0xc009852a00, 0x0?, {0x30983c0, 0xc006d61560})
        /build/.go/pkg/mod/github.com/getsentry/sentry-go@v0.16.0/client.go:642 +0x6a3 fp=0xc01795ecf0 sp=0xc01795def8 pc=0x8d2683
[...]

PR #638, opened 2023-05-11
fix: Clone scope.Context in multiple places to avoid concurrent reads/writes

tonyo:
Fixes https://github.com/getsentry/sentry-go/issues/570

Here we introduce `cloneContext` that should be used in multiple places to avoid issues like in https://github.com/getsentry/sentry-go/issues/570. 

Note: `cloneContext` does not do a deep copy of `Context` values (which are of type `interface{}`), so there are cases where it's still possible to catch panics if you do weird things with `Context` values. But this, I believe, is out of scope of the SDK responsibility.

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -287,7 +287,7 @@ func (scope *Scope) Clone() *Scope {
 		clone.tags[key] = value
 	}
 	for key, value := range scope.contexts {
-		clone.contexts[key] = value
+		clone.contexts[key] = cloneContext(value)
 	}
 	for key, value := range scope.extra {
 		clone.extra[key] = value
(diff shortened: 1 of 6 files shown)


scope.go
```diff
@@ -404,3 +404,16 @@ func (scope *Scope) ApplyToEvent(event *Event, hint *EventHint) *Event {
 
 	return event
 }
+
+// cloneContext returns a new context with keys and values copied from the passed one.
+//
+// Note: a new Context (map) is returned, but the function does NOT do
+// a proper deep copy: if some context values are pointer types (e.g. maps),
+// they won't be properly copied.
+func cloneContext(c Context) Context {
```
cleptric, line 413: Is this correct? We already iterated over the contexts of the scope and event at this point, so here we would iterate over the value of the context only.

PR #644, opened 2023-05-29
feat(gin): Add transactions to Gin integration

zetaab:
add missing transactions for gin

### DIFF
--- a/gin/sentrygin.go
+++ b/gin/sentrygin.go
@@ -2,6 +2,7 @@ package sentrygin
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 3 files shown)


tonyo:
Hey @zetaab , thanks for the changes 👍 
Looks good; the only thing: could you add a test or two, to make sure it works as expected? 

Could be something like https://github.com/getsentry/sentry-go/blob/master/http/sentryhttp_test.go, but can be even simpler (at least to make sure it doesn't crash).

zetaab:
@tonyo yeah I will do those, just take some days :)

zetaab:
@tonyo sorry for copy pasting the tests, but now its almost similar than in normal http lib. This does not include transaction tests. I will continue with those tomorrow.

zetaab:
actually it was quite easy to add transaction tests, so those are included now as well

cleptric:
Thanks, we'll take a closer look soon. For now, please revert any dependencies bumps as we do not want to raise our base level.
Regarding Gin, you can read up on our [Security Policy](https://github.com/getsentry/sentry-go/security/policy) and why we do not force an upgrade from the SDK.

gin/sentrygin.go
```diff
@@ -46,15 +47,33 @@ func New(options Options) gin.HandlerFunc {
...
+		sentry.ContinueFromRequest(c.Request),
+		sentry.WithTransactionSource(sentry.SourceURL),
+	}
+
+	transaction := sentry.StartTransaction(ctx,
+		fmt.Sprintf("%s %s", c.Request.Method, c.Request.URL.Path),
+		options...,
+	)
+	defer func() {
+		transaction.Status = sentry.HTTPtoSpanStatus(c.Writer.Status())
```
zetaab, line 68: added Status and tests for that as well

tonyo (approved):

tonyo:
Thanks!

PR #651, opened 2023-06-12
refactor: use client.options internally to avoid copying

vaind:
Removes use of `client.Options()` from the internal code to avoid copying the options object.

#skip-changelog

### DIFF
--- a/client.go
+++ b/client.go
@@ -240,6 +240,28 @@ type Client struct {
 // single goroutine) or hub methods (for concurrent programs, for example web
 // servers).
 func NewClient(options ClientOptions) (*Client, error) {
+	// The default error event sample rate for all SDKs is 1.0 (send all).
+	//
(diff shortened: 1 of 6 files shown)


client.go
```diff
@@ -374,8 +396,8 @@ func (client *Client) AddEventProcessor(processor EventProcessor) {
 }
 
 // Options return ClientOptions for the current Client.
-// TODO don't access this internally to avoid creating a copy each time.
 func (client Client) Options() ClientOptions {
+	// Note: internally, use `client.options` instead of `client.Options()` to avoid creating the object copy each time.
```
cleptric, line 400:
```suggestion
	// Note: internally, we use `client.options` instead of `client.Options()` to avoid copying the object each time.
```

cleptric (approved):

cleptric:
Let me get you merge access, brb 🙂

PR #652, opened 2023-06-12
Read release info from build info

mattrobenolt:
Since go1.18, vcs.revision gets compiled into the binary and can be read.

Conveniently, debug.ReadBuildInfo has been around since go1.12, so there's no need to conditionally compile this unless sentry-go needs to support < go1.12 for some reason.

So in builds newer than 1.18, this key just simply won't exist and effectively will do nothing.

This is much more convenient to detect in anything go1.18 since it's much more common to have this compiled in.

### DIFF
--- a/util.go
+++ b/util.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"runtime/debug"
 	"strings"
 	"time"
 
@@ -63,6 +64,15 @@ func defaultRelease() (release string) {
 		if release = os.Getenv(e); release != "" {
 			Logger.Printf("Using release from environment variable %s: %s", e, release)
 			return release
+		}
+	}
+
+	if info, ok := debug.ReadBuildInfo(); ok {
+		for _, setting := range info.Settings {
+			if setting.Key == "vcs.revision" && setting.Value != "" {
+				Logger.Printf("Using release from debug info: %s", setting.Value)
+				return setting.Value
+			}
 		}
 	}
 


cleptric:
The SDK requires baseline Go 1.18, so this is no issue https://github.com/getsentry/sentry-go/blob/97a00a4a9d0b9170266b3d83b2199a526ab2aa74/go.mod#L3

Any chance you could add some tests?

util.go
```diff
@@ -66,6 +67,15 @@ func defaultRelease() (release string) {
...
 
+	if info, ok := debug.ReadBuildInfo(); ok {
+		for _, setting := range info.Settings {
+			if setting.Key == "vcs.revision" && setting.Value != "" {
+				Logger.Printf("Using release from debug info: %s", setting.Value)
+				return setting.Value
+			}
+		}
+	}
+
```
cleptric, line 78: We may even move this above the env detection.

mattrobenolt:
> Any chance you could add some tests?

No, I'm not being paid by Sentry.

util.go
cleptric: Yeah, env should be the first, you're right. nvm

kamilogorek:
Hi Matt.

mattrobenolt:
> Hi Matt.

👋 just trying to use your software over here.

cleptric:
> > Any chance you could add some tests?
> 
> No, I'm not being paid by Sentry.

I understand, but we would like to handle this PR as any other PRs we receive from contributors.
I'll leave this open in the meantime.

mattrobenolt:
Yeah and that's fine. You're welcome to add your own tests. I'm not sure how this is easily testable and it's a trivial change. But since you're insisting, there's no tests at all for this entire function.

We've already worked around this by simply using my code in our own config. Figured I'd try to upstream it since it makes sense for others.

Issue #587, opened 2023-02-23
`span.Finish()` should be a noop when span is already finished


tonyo:
At the moment we allow to call `Finish()` on spans (and, specifically, transactions) multiple times, and run the full `Finish()` logic every time.

It's not optimal, because finalizing the same span/transaction is not something that we officially support or recommend, and it also might cause a data race, as in the following example:
```go
transaction := StartTransaction(ctx, "op")

go func() {
	for {
		transaction.Finish()
	}
}()

go func() {
	for {
		transaction.Finish()
	}
}()

time.Sleep(100 * time.Second)
```

> 

<details>
<summary>Data race log</summary>

```
==================
WARNING: DATA RACE
Read at 0x00c00013c9a8 by goroutine 9:
  github.com/getsentry/sentry-go.(*environmentIntegration).processor()
      /Users/anton/reps/sentry-go/integrations.go:91 +0x3ad
  github.com/getsentry/sentry-go.(*environmentIntegration).processor-fm()
      <autogenerated>:1 +0x4d
  github.com/getsentry/sentry-go.(*Client).prepareEvent()
      /Users/anton/reps/sentry-go/client.go:651 +0xb69
  github.com/getsentry/sentry-go.(*Client).processEvent()
      /Users/anton/reps/sentry-go/client.go:571 +0x4a4
  github.com/getsentry/sentry-go.(*Client).CaptureEvent()
      /Users/anton/reps/sentry-go/client.go:397 +0xa8
  github.com/getsentry/sentry-go.(*Hub).CaptureEvent()
      /Users/anton/reps/sentry-go/hub.go:224 +0x92
  github.com/getsentry/sentry-go.(*Span).Finish()
      /Users/anton/reps/sentry-go/tracing.go:205 +0x2e4
  github.com/getsentry/sentry-go.TestSetContextAndFinishTransaction.func1()
      /Users/anton/reps/sentry-go/tracing_test.go:850 +0x30

Previous write at 0x00c00013c9a8 by goroutine 10:
  github.com/getsentry/sentry-go.(*environmentIntegration).processor()
      /Users/anton/reps/sentry-go/integrations.go:92 +0x3f8
  github.com/getsentry/sentry-go.(*environmentIntegration).processor-fm()
      <autogenerated>:1 +0x4d
  github.com/getsentry/sentry-go.(*Client).prepareEvent()
      /Users/anton/reps/sentry-go/client.go:651 +0xb69
[...]

PR #660, opened 2023-07-04
fix(tracing): Make Span.Finish a no-op when the span is already finished

tonyo:
Span.Finish() should only be executed once, all the further executions should be no-ops. This should also prevent potential data races when Finish() is (erroneously) called concurrently.

Closes https://github.com/getsentry/sentry-go/issues/587

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -60,7 +60,7 @@ type Span struct { //nolint: maligned // prefer readability over optimal memory 
 	contexts map[string]Context
 	// profiler instance if attached, nil otherwise.
 	profiler transactionProfiler
-	// a Once instance to make sure that Finish() is only called once
+	// a Once instance to make sure that Finish() is only called once.
 	finishOnce sync.Once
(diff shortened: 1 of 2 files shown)


tracing.go
```diff
@@ -194,12 +196,8 @@ func StartSpan(ctx context.Context, operation string, options ...SpanOption) *Sp
...
 }
 
-// Finish sets the span's end time, unless already set. If the span is the root
-// of a span tree, Finish sends the span tree to Sentry as a transaction.
-func (s *Span) Finish() {
-	// TODO(tracing): maybe make Finish run at most once, such that
-	// (incorrectly) calling it twice never double sends to Sentry.
-
+// doFinish runs the actual Span.Finish() logic.
+func (s *Span) doFinish() {
```
cleptric, line 200: For readability, can you move this below `Finish()`?

cleptric (approved):

PR #661, opened 2023-07-04
feat(crons): initial cron support

### LINKED ISSUE #659: Cron implementation
Implement this feature: https://docs.sentry.io/product/crons/

(@cleptric) I would imagine something similar to what we did in PHP https://github.com/getsentry/sentry-php/pull/1467 & https://github.com/getsentry/sentry-php/pull/1511

Node implementation: https://github.com/getsentry/sentry-javascript/pull/8039

It turns out there's a parent issue: https://github.com/getsentry/sentry/issues/48560

UPDATE: Wrong parent issue 🤦

aldy505:
Closes #659 

This implementations looks up to:
- https://github.com/getsentry/sentry-php/pull/1467
- https://github.com/getsentry/sentry-php/pull/1511
- https://github.com/getsentry/sentry-javascript/commit/a2cda4df089ad372f721c94b07dd752211acf820
- https://github.com/getsentry/sentry-javascript/pull/8039

### DIFF
--- a/check_in.go
+++ b/check_in.go
@@ -90,6 +90,8 @@ type MonitorConfig struct { //nolint: maligned // prefer readability over optima
 }
 
 type CheckIn struct { //nolint: maligned // prefer readability over optimal memory layout
+	// Check-In ID (unique and client generated)
+	ID string `json:"check_in_id"`
 	// The distinct slug of the monitor.
 	MonitorSlug string `json:"monitor_slug"`
 	// The status of the check-in.
--- a/check_in.go
+++ b/check_in.go
@@ -1,0 +1,110 @@
+package sentry
+
+type CheckInStatus string
+
+const (
+	CheckInStatusInProgress CheckInStatus = "in_progress"
+	CheckInStatusOK         CheckInStatus = "ok"
+	CheckInStatusError      CheckInStatus = "error"
+)
+
+type checkInScheduleType string
+
+const (
+	checkInScheduleTypeCrontab  checkInScheduleType = "crontab"
+	checkInScheduleTypeInterval checkInScheduleType = "interval"
+)
+
+type MonitorSchedule interface {
+	scheduleType() checkInScheduleType
+}
+
+type crontabSchedule struct {
+	Type  string `json:"type"`
+	Value string `json:"value"`
+}
+
+func (c crontabSchedule) scheduleType() checkInScheduleType {
+	return checkInScheduleTypeCrontab
+}
+
+// CrontabSchedule defines the MonitorSchedule with a cron format.
+// Example: "8 * * * *".
+func CrontabSchedule(scheduleString string) MonitorSchedule {
+	return crontabSchedule{
+		Type:  string(checkInScheduleTypeCrontab),
+		Value: scheduleString,
+	}
+}
+
+type intervalSchedule struct {
+	Type  string `json:"type"`
+	Value int64  `json:"value"`
+	Unit  string `json:"unit"`
+}
+
+func (i intervalSchedule) scheduleType() checkInScheduleType {
+	return checkInScheduleTypeInterval
+}
+
+type MonitorScheduleUnit string
+
+const (
+	MonitorScheduleUnitMinute MonitorScheduleUnit = "minute"
+	MonitorScheduleUnitHour   MonitorScheduleUnit = "hour"
+	MonitorScheduleUnitDay    MonitorScheduleUnit = "day"
+	MonitorScheduleUnitWeek   MonitorScheduleUnit = "week"
+	MonitorScheduleUnitMonth  MonitorScheduleUnit = "month"
+	MonitorScheduleUnitYear   MonitorScheduleUnit = "year"
+)
+
+// IntervalSchedule defines the MonitorSchedule with an interval format.
+//
+// Example:
+//
+//	IntervalSchedule(1, sentry.MonitorScheduleUnitDay)
+func IntervalSchedule(value int64, unit MonitorScheduleUnit) MonitorSchedule {
+	return intervalSchedule{
+		Type:  string(checkInScheduleTypeInterval),
+		Value: value,
+		Unit:  string(unit),
+	}
+}
+
+type MonitorConfig struct { //nolint: maligned // prefer readability over optimal memory layout
+	Schedule MonitorSchedule `json:"schedule,omitempty"`
+	// The allowed allowed margin of minutes after the expected check-in time that
+	// the monitor will not be considered missed for.
+	CheckInMargin int64 `json:"check_in_margin,omitempty"`
+	// The allowed allowed duration in minutes that the monitor may be `in_progress`
+	// for before being considered failed due to timeout.
+	MaxRuntime int64 `json:"max_runtime,omitempty"`
+	// A tz database string representing the timezone which the monitor's execution schedule is in.
+	// See: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
+	Timezone string `json:"timezone,omitempty"`
+}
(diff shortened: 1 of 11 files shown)


client.go
```diff
@@ -413,6 +413,12 @@ func (client *Client) CaptureException(exception error, hint *EventHint, scope E
 	return client.CaptureEvent(event, hint, scope)
 }
 
+// CaptureCheckIn captures a check in.
+func (client *Client) CaptureCheckIn(checkIn *CheckIn, monitorConfig *MonitorConfig, scope EventModifier) *EventID {
```
aldy505, line 417: There was a discussion in Discord (#off-topic) between me and @cleptric about what function name to implement (as the PHP and Node implementation is quite different), where Michi suggested to follow the PHP implementation as the event is the check in already. As for Go SDK, if this implementation folows PHP's, this would be something like:

```go
event := &sentry.Event{
 // fill payloads
}

event.SetCheckIn(checkInPayload)

eventId :=  sentry.CaptureEvent(event)
```

It's kind of awkward and confusing for me, so I went ahead with the `CaptureCheckIn` as a method of `Client` struct.

interfaces.go
```diff
@@ -315,6 +318,12 @@ type Event struct {
 	Spans           []*Span          `json:"spans,omitempty"`
 	TransactionInfo *TransactionInfo `json:"transaction_info,omitempty"`
 
+	// The fields below are only relevant for crons/check ins
+
+	CheckInID     string         `json:"check_in_id,omitempty"`
+	CheckIn       *CheckIn       `json:"check_in,omitempty"`
+	MonitorConfig *MonitorConfig `json:"monitor_config,omitempty"`
```
aldy505, line 325: If I follow how transactions are keeping their data, it would mean that every entry on `CheckIn` would be spread out and might mess with other fields in the future, so I keep it enclosed here with `omitempty` tag.

interfaces.go
```diff
@@ -449,6 +460,28 @@ func (e *Event) transactionMarshalJSON() ([]byte, error) {
...
+		MonitorConfig: nil,
+	}
+
+	if e.MonitorConfig != nil {
+		checkIn.MonitorConfig = &MonitorConfig{}
+		checkIn.MonitorConfig.Schedule = e.MonitorConfig.Schedule
+		checkIn.MonitorConfig.CheckInMargin = e.MonitorConfig.CheckInMargin
+		checkIn.MonitorConfig.MaxRuntime = e.MonitorConfig.MaxRuntime
+		checkIn.MonitorConfig.Timezone = e.MonitorConfig.Timezone
+	}
```
aldy505, line 480: Yes I know it looks bad. It's 10 PM on Tuesday and I can't think of anything better right now. So sorry 🙈

tonyo:
Thanks @aldy505 , we'll have a look this/next week 👍

check_in.go
```diff
@@ -0,0 +1,110 @@
...
+	Timezone string `json:"timezone,omitempty"`
+}
+
+type CheckIn struct { //nolint: maligned // prefer readability over optimal memory layout
+	// The distinct slug of the monitor.
+	MonitorSlug string `json:"monitor_slug"`
+	// The status of the check-in.
+	Status CheckInStatus `json:"status"`
+	// The duration of the check-in in seconds. Will only take effect if the status is ok or error.
+	Duration int64 `json:"duration,omitempty"`
```
tonyo, line 93: Wdyt about changing the type to `time.Duration` here? When serialized it should be a float (in seconds), but on the high level interface we could make it more developer-friendly.

check_in.go
```diff
@@ -0,0 +1,110 @@
...
+// See https://develop.sentry.dev/sdk/check-ins/
+type serializedCheckIn struct { //nolint: maligned
+	// Check-In ID (unique and client generated).
+	CheckInID string `json:"check_in_id"`
+	// The distinct slug of the monitor.
+	MonitorSlug string `json:"monitor_slug"`
+	// The status of the check-in.
+	Status CheckInStatus `json:"status"`
+	// The duration of the check-in in seconds. Will only take effect if the status is ok or error.
+	Duration      int64          `json:"duration,omitempty"`
```
tonyo, line 106: Let's make the serialized duration float64, please.

tonyo (changes requested):
Overall looking good 👍 
Left some comments, but nothing super major.

client.go
```diff
@@ -413,6 +413,12 @@ func (client *Client) CaptureException(exception error, hint *EventHint, scope E
...
+	event := client.EventFromCheckIn(checkIn, monitorConfig)
+	return client.processEvent(event, nil, scope)
```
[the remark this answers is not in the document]
aldy505: I'll have to reread the code before committing the suggested changes. I don't remember if it's okay to call that method from here.

check_in.go
```diff
@@ -0,0 +1,110 @@
...
+type MonitorSchedule interface {
+	scheduleType() checkInScheduleType
+}
+
+type crontabSchedule struct {
+	Type  string `json:"type"`
+	Value string `json:"value"`
+}
+
+func (c crontabSchedule) scheduleType() checkInScheduleType {
```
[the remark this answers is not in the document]
aldy505: As a matter of fact, I never call this method anywhere. I implement this just for it to have the same signature with the interface (forgot the name of it). I might change this later to be more idiomatic. 

If on any other thoughts, I didn't change this part of code, I don't think modifying it to have a pointer receiver would make any difference.

check_in.go
aldy505: I'm okay with having it to be `time.Duration`.

check_in.go
aldy505: As I read it from the PHP implementation, it's int, not float. Otherwise we might ping @cleptric for clarification.

cleptric: https://github.com/getsentry/sentry-php/blob/master/src/CheckIn.php#L36-L39

It's `int|float|null` in PHP, so for Go, we should use `float64` as @tonyo recommended.

check_in.go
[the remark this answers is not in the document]
tonyo: So to clarify, you've basically added this method to disallow implementing`MonitorSchedule` interface outside the SDK, right?

aldy505:
I'll address any errors on CI in a few hours.

tonyo:
Failures on Windows can be ignored for now, seems like it's only a test that has been flaky recently.

check_in.go
```diff
@@ -0,0 +1,115 @@
...
+	Timezone string `json:"timezone,omitempty"`
+}
+
+type CheckIn struct { //nolint: maligned // prefer readability over optimal memory layout
+	// The distinct slug of the monitor.
+	MonitorSlug string `json:"monitor_slug"`
+	// The status of the check-in.
+	Status CheckInStatus `json:"status"`
+	// The duration of the check-in. Will only take effect if the status is ok or error.
+	Duration time.Duration `json:"duration,omitempty"`
```
tonyo, line 98: What I realized is that we don't have an option to pass an existing `CheckInID` via `CheckIn` now.
At the moment we basically create a new CheckInID every time, so there's no way to link the "in-progress" and "ok"/"error" status.
Could you handle this case please?

tonyo:
Another thing that we'll have to do eventually is to update the docs, basically adding Go to this list: https://docs.sentry.io/product/crons/getting-started/
@aldy505 let us know if you're ready to contribute to the docs too; that'd speed up the release/adoption of the feature.

check_in.go
aldy505: Done. I might want to add a test case for this one, or an example.

tonyo (approved):

PR #670, opened 2023-07-20
feat: Attachments support

### LINKED ISSUE #574: Support for attachments in go sdk
### Problem Statement

It will be great if you could add the support for attachments for go SDK. 

### Solution Brainstorm

_No response_

edigaryev:
This is similar to how [`sentry-rust` supports attachments](https://docs.rs/sentry/0.31.5/sentry/struct.Scope.html#method.add_attachment).

Resolves #574.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -94,6 +94,38 @@ func getRequestBodyFromEvent(event *Event) []byte {
 	return nil
 }
 
+func encodeAttachment(enc *json.Encoder, b io.Writer, attachment *Attachment) error {
+	// Attachment header
+	err := enc.Encode(struct {
+		Type        string `json:"type"`
+		Length      int    `json:"length"`
+		Filename    string `json:"filename"`
+		ContentType string `json:"content_type,omitempty"`
+	}{
+		Type:        "attachment",
+		Length:      len(attachment.Payload),
+		Filename:    attachment.Filename,
+		ContentType: attachment.ContentType,
+	})
+	if err != nil {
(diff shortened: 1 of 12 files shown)


transport.go
```diff
@@ -94,6 +94,38 @@ func getRequestBodyFromEvent(event *Event) []byte {
 	return nil
 }
 
+func encodeAttachment(enc *json.Encoder, b io.Writer, attachment *Attachment) error {
```
tonyo, line 97: Please add a test in `transport_test.go` that tests this function.

tonyo (commented):
Looking good, just one ask about test coverage 👍

transport.go
edigaryev: PTAL at https://github.com/getsentry/sentry-go/pull/670/commits/52caef25b90d3055ea9ee5066536c2493a4d4fed.

Not sure if testing `encodeAttachment()` makes sense in the presence of tests like `TestEnvelopeFromErrorBody`, which already will call that function under the hood if we set `event.attachments` to non-`nil`.

tonyo: Yep, what you added is actually what I had in mind 👍

tonyo (approved):

tonyo:
Thanks @edigaryev 🎉

tonyo:
The attachments support is now released as part of https://github.com/getsentry/sentry-go/releases/tag/v0.23.0 👍

PR #672, opened 2023-07-24
Expose SpanFromContext

zhezhel:
I'm working on integration [pgx](https://github.com/jackc/pgx) with Sentry. 

The new version of pgx provides an [interface](https://github.com/jackc/pgx/blob/master/tracer.go#L10) for tracing, which work in the way when you need to implement two different methods: TraceQueryStart & TraceQueryEnd.   

To implement `TraceQueryEnd`, I need to get access to span in context and call `Finish` on it.   

It would be great to provide a method to get span from context.   
Current workaround: 
```go
func SpanFromContext(ctx interface{}) (*sentry.Span, bool) {
	values := reflect.ValueOf(ctx).Elem()

	for i := 0; i < values.NumField(); i++ {
		reflectValue := values.Field(i)
		reflectValue = reflect.NewAt(reflectValue.Type(), unsafe.Pointer(reflectValue.UnsafeAddr())).Elem()
		span, ok := reflectValue.Interface().(*sentry.Span)
		if !ok {
			continue
		}
		return span, true
	}
	return nil, false
}
```

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -956,27 +956,9 @@ func TransactionFromContext(ctx context.Context) *Span {
 	return nil
 }
 
-// spanFromContext returns the last span stored in the context or a dummy
+// SpanFromContext returns the last span stored in the context or a dummy
 // non-nil span.
-//
-// TODO(tracing): consider exporting this. Without this, users cannot retrieve a
-// span from a context since spanContextKey is not exported.
-//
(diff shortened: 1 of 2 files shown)


cleptric (approved):
Works for me, @tonyo ?

zhezhel:
Hey @cleptric, @tonyo, do you have plans to merge this PR?

PR #674, opened 2023-07-25
fix(crons): Pass correct event types, fix checkin_margin field, add example

tonyo:
Follow-up to #661 that fixes a few issues and also adds an E2E example.

* `check_in_margin` -> `checkin_margin` (https://develop.sentry.dev/sdk/check-ins/#monitor-upsert-support)
* When creating the event, set the type to `checkinType`
* Do not force event and check-in IDs to be the same

### DIFF
--- a/_examples/crons/main.go
+++ b/_examples/crons/main.go
@@ -1,0 +1,72 @@
+package main
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/getsentry/sentry-go"
+)
+
+func runTask(monitorSlug string, duration time.Duration, shouldFail bool) {
+	checkinId := sentry.CaptureCheckIn(
+		&sentry.CheckIn{
+			MonitorSlug: monitorSlug,
+			Status:      sentry.CheckInStatusInProgress,
+		},
+		&sentry.MonitorConfig{
+			Schedule:      sentry.CrontabSchedule("* * * * *"),
+			MaxRuntime:    2,
+			CheckInMargin: 1,
+		},
+	)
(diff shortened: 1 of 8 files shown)


check_in.go
```diff
@@ -80,7 +80,7 @@ type MonitorConfig struct { //nolint: maligned // prefer readability over optima
 	Schedule MonitorSchedule `json:"schedule,omitempty"`
 	// The allowed margin of minutes after the expected check-in time that
 	// the monitor will not be considered missed for.
-	CheckInMargin int64 `json:"check_in_margin,omitempty"`
+	CheckInMargin int64 `json:"checkin_margin,omitempty"`
```
cleptric, line 83: Hmm, this almost looks like a typo to me. Will raise with the Crons team.

_examples/crons/main.go
```diff
@@ -0,0 +1,72 @@
...
+		status = sentry.CheckInStatusOK
+	}
+
+	sentry.CaptureCheckIn(
+		&sentry.CheckIn{
+			ID:          *checkinId,
+			MonitorSlug: monitorSlug,
+			Status:      status,
+		},
+		nil,
```
cleptric, line 40: I don't know if this causes issues. In Laravel, we sent the monitor config along for all check-ins.

cleptric (approved):

_examples/crons/main.go
tonyo: It sorta feels natural to me: when the monitor already exists (guaranteed by the first check-in), we don't have to pass in the monitor config anymore.

PR #675, opened 2023-07-27
Sentry gin performance transaction improvement

antonsacred:
This improvement should improve usability of sentry performance

This PR consist of two changes:
1. Removed http method from transaction name (this is unnecessary because sentry performance has http method info already)
2. Changed transaction name to `context.FullPath()` witch is more readable for endpoint owners

Changes are shown in screenshot. First one is before, second one is after
Screenshot 2023-07-27 at 2 54 17 PM

### DIFF
--- a/gin/sentrygin.go
+++ b/gin/sentrygin.go
@@ -60,7 +60,7 @@ func (h *handler) handle(c *gin.Context) {
 	}
 
 	transaction := sentry.StartTransaction(ctx,
-		c.Request.URL.Path,
+		c.FullPath(),
 		options...,
 	)
 	defer func() {
--- a/gin/sentrygin.go
+++ b/gin/sentrygin.go
@@ -2,7 +2,6 @@ package sentrygin
 
 import (
 	"context"
-	"fmt"
 	"net"
 	"net/http"
 	"os"
@@ -61,7 +60,7 @@ func (h *handler) handle(c *gin.Context) {
 	}
 
 	transaction := sentry.StartTransaction(ctx,
-		fmt.Sprintf("%s %s", c.Request.Method, c.Request.URL.Path),
+		c.Request.URL.Path,
 		options...,
 	)
 	defer func() {
--- a/gin/sentrygin_test.go
+++ b/gin/sentrygin_test.go
@@ -31,7 +31,7 @@ func TestIntegration(t *testing.T) {
 		WantTransaction *sentry.Event
 	}{
 		{
-			Path:       "/panic",
+			Path:       "/panic/:id",
 			Method:     "GET",
 			WantStatus: 200,
 			Handler: func(c *gin.Context) {
@@ -40,9 +40,9 @@ func TestIntegration(t *testing.T) {
 			WantTransaction: &sentry.Event{
 				Level:       sentry.LevelInfo,
 				Type:        "transaction",
-				Transaction: "GET /panic",
+				Transaction: "/panic/:id",
 				Request: &sentry.Request{
-					URL:    "/panic",
+					URL:    "/panic/:id",
 					Method: "GET",
 					Headers: map[string]string{
 						"Accept-Encoding": "gzip",
@@ -55,7 +55,7 @@ func TestIntegration(t *testing.T) {
 				Level:   sentry.LevelFatal,
 				Message: "test",
 				Request: &sentry.Request{
-					URL:    "/panic",
+					URL:    "/panic/:id",
 					Method: "GET",
 					Headers: map[string]string{
 						"Accept-Encoding": "gzip",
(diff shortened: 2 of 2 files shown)


gin/sentrygin.go
```diff
@@ -61,7 +60,7 @@ func (h *handler) handle(c *gin.Context) {
 	}
 
 	transaction := sentry.StartTransaction(ctx,
-		fmt.Sprintf("%s %s", c.Request.Method, c.Request.URL.Path),
+		c.FullPath(),
```
cleptric, line 63: Any specific reason you removed the HTTP method verb?

antonsacred: Yes, cause it's unnecessary for HTTP Method in transaction name. 
There is two reasons for that: 
1) HTTP Method is already present in translation and no need to duplicate this info 
2) According to [screenshots from sentry docs](https://docs.sentry.io/product/performance/transaction-summary/?original_referrer=https%3A%2F%2Fwww.google.com%2F) transaction name should consist of endpoint without HTTP Method

antonsacred:
> Can you confirm all routes will now be fully parameterized?
> 
> If yes, you can update `sentry.WithTransactionSource(sentry.SourceURL)` to `sentry.WithTransactionSource(sentry.SourceRoute)`.

thanks, done

gin/sentrygin.go
cleptric: I see, I would still keep it, because:

- Not all places in the UI show the `HTTP method`; Hence it is hard to distinguish between transactions with different HTTP methods.
- This is a breaking change, as it will change the behaviour of a configured `traces_sampler` as well as dashboards or alerts created in the product. For the sake of better bucketing, I'm ok doing this, but I wanted to inform you of the implications.

antonsacred: What do you think about giving developers the ability to turn on/off those transactions?
This way I can implement transaction as I want it, but to do that I need the way of disabling built-in transaction.

antonsacred: I think I can do it in separate PR

cleptric: We have certain expectations in the product for how a transaction should be structured. In the case of Go, this is a big problem, as seen in your screenshot above. The `/*` is applied during ingestion, where we perform server-side bucketing to group similar transactions together. This is expensive.
Hence we absolutely do not want to make this process a bigger foot gun as it already is 🙂 

For now, please keep the HTTP verbs. The change is already a massive improvement from before.

antonsacred: Ok, no problem, I can live with HTTP method. Will return it

antonsacred: Done

antonsacred:
Any action required from my side?

cleptric:
@antonsacred No, all good. CI failed because our Windows tests started to be a bit flaky. We'll get this merged in the next few days.

gin/sentrygin.go
```diff
@@ -57,11 +57,11 @@ func (h *handler) handle(c *gin.Context) {
 	options := []sentry.SpanOption{
 		sentry.WithOpName("http.server"),
 		sentry.ContinueFromRequest(c.Request),
-		sentry.WithTransactionSource(sentry.SourceURL),
+		sentry.WithTransactionSource(sentry.SourceRoute),
 	}
 
 	transaction := sentry.StartTransaction(ctx,
-		fmt.Sprintf("%s %s", c.Request.Method, c.Request.URL.Path),
+		fmt.Sprintf("%s %s", c.Request.Method, c.FullPath()),
```
cleptric, line 64:
```
For not found routes returns an empty string.
```

This causes issues for 404 routes, as the transaction name is now always something like `GET `.
We should check for the return of `c.FullPath()`. If it's empty, use `c.Request.URL.Path` as a fallback and set `sentry.WithTransactionSource(sentry.SourceURL)`.

gin/sentrygin_test.go
cleptric, line 1: It would be great to test different paths, like `/panic`, `/panic/:id` etc.

cleptric (changes requested):

gin/sentrygin.go
antonsacred: That's cool that you found it! Done
https://github.com/getsentry/sentry-go/pull/675/commits/3b50c09995a5a8f149545f5e9ea4f52d9e77d061

gin/sentrygin_test.go
antonsacred: https://github.com/getsentry/sentry-go/pull/675/commits/c1edc111039f7b7866b7b795aa5e4c4fe212d569
This way I tried to show where request and router paths are in used

antonsacred:
@cleptric pls review again

gin/sentrygin.go
tonyo: @antonsacred I think what @cleptric suggests is more like, do the check and fallback here in `handle` directly (not in `defer`), and set TransactionSource depending on the value of `c.FullPath()`.

tonyo (approved):

cleptric (approved):

PR #679, opened 2023-07-28
ref(otel): Add `ClientOptions.Instrumenter`

### LINKED ISSUE #678: [otel] integration should sync sampling decision to/from otel in the "parentless" case
## Summary

Currently, using something like the following seems to lead to unexpected results:
```go
	trace.NewTracerProvider(
		trace.WithSampler(
			trace.ParentBased(
				trace.TraceIDRatioBased(cfg.SampleRate),
			),
		),
		trace.WithSpanProcessor(sentryotel.NewSentrySpanProcessor()),
	)
```

The decision from the otel sampler is ignored by sentry, which then samples again based on its own logic, leaving the otel and sentry samplig decisions disconnected.

Using _only_ `sentry.ClientOptions.TracesSampler` is not an option for similar reasons: its decision does not get passed back to otel, so otel's `Span.IsSampled()` returns an unrelated value to sentry's decision (so that you end up - e.g. - using a traceID in exemplars and logs that never actually got sent to sentry, or the other way around).

The otel and sentry sides seem to agree in the case where a parent span could be found.

Am I holding it wrong? :sweat_smile: 

## Steps To Reproduce

Initialize otel with:
```go
trace.NewTracerProvider(
	trace.WithSampler(trace.AlwaysSample()),
	trace.WithSpanProcessor(sentryotel.NewSentrySpanProcessor()),
)
```

Initialize sentry with:
```go
sentry.ClientOptions{
	//...
	EnableTracing:    true,
	TracesSampleRate: 0.0,
	//...
}
```

All traces created with `otel.Tracer("").Start()` will have `span.IsSampled() == true`, but none of them will be sent to sentry.

Switching the to `NeverSample()`/`TracesSampleRate: 1.0` causes the opposite problem.

## Expected Behavior

Sentry should respect OTel's `span.IsSampled()` and only sample trances if and only if it is true.

(The other way around seems a bit more difficult to pull off?)

### SDK
 - `sentry-go` 0.22.0: 
 - Go version: 1.20
 - Using Go Modules? yes

### Sentry

costela:
This small PR is a first attempt to fix the perceived issue in #678.

It extends `sentry.ClientOptions` with the `Instrumenter` parameter, analogous to other language SDKs. This can be optionally set to "otel" or "sentry" (the default).  

This new option is then respected by middlewares to optionally skip creating a sentry transaction, allowing the consumer to only use "pure" otel for traces.

### Considered alternatives:

Another option would be making the transaction created by the otel [span processor](https://pkg.go.dev/github.com/getsentry/sentry-go/otel#NewSentrySpanProcessor) visible to the middlewares. However, since the span processor cannot easily modify the request context, it becomes necessary to add additional helpers (e.g. a new "adapter" middleware) with the single task of injecting the otel-created transaction into the request context.

Fixes: #678 #788

### DIFF
--- a/client.go
+++ b/client.go
@@ -133,6 +133,9 @@ type ClientOptions struct {
 	TracesSampleRate float64
 	// Used to customize the sampling of traces, overrides TracesSampleRate.
 	TracesSampler TracesSampler
+	// Which instrumentation to use for tracing. Either "sentry" (default) or "otel" are supported.
+	// Setting this to "otel" will ignore TracesSampleRate and TracesSampler and assume sampling is performed by otel.
+	Instrumenter string
 	// The sample rate for profiling traces in the range [0.0, 1.0].
 	// This is relative to TracesSampleRate - it is a ratio of profiled traces out of all sampled traces.
 	ProfilesSampleRate float64
@@ -299,6 +302,19 @@ func NewClient(options ClientOptions) (*Client, error) {
 
 	if options.MaxSpans == 0 {
 		options.MaxSpans = defaultMaxSpans
+	}
+
+	switch options.Instrumenter {
+	case "":
+		options.Instrumenter = "sentry"
+	case "sentry":
+		// noop
+	case "otel":
+		// sampling is performed by the OpenTelemetry SDK
+		options.TracesSampleRate = 1.0
+		options.TracesSampler = nil
+	default:
+		return nil, fmt.Errorf("invalid value for TracesInstrumenter (supported are 'sentry' and 'otel'): %q", options.Instrumenter)
 	}
 
 	// SENTRYGODEBUG is a comma-separated list of key=value pairs (similar
--- a/client.go
+++ b/client.go
@@ -309,7 +309,7 @@ func NewClient(options ClientOptions) (*Client, error) {
 		options.Instrumenter = "sentry"
 	case "sentry", "otel": // noop
 	default:
-		return nil, fmt.Errorf("invalid value for TracesInstrumenter (supported are 'sentry' and 'otel'): %q", options.Instrumenter)
+		return nil, fmt.Errorf("invalid value for Instrumenter (supported are 'sentry' and 'otel'): %q", options.Instrumenter)
 	}
 
 	// SENTRYGODEBUG is a comma-separated list of key=value pairs (similar
--- a/client.go
+++ b/client.go
@@ -307,12 +307,7 @@ func NewClient(options ClientOptions) (*Client, error) {
 	switch options.Instrumenter {
 	case "":
 		options.Instrumenter = "sentry"
-	case "sentry":
-		// noop
-	case "otel":
-		// sampling is performed by the OpenTelemetry SDK
-		options.TracesSampleRate = 1.0
-		options.TracesSampler = nil
+	case "sentry", "otel": // noop
 	default:
 		return nil, fmt.Errorf("invalid value for TracesInstrumenter (supported are 'sentry' and 'otel'): %q", options.Instrumenter)
 	}
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ TODO.md
 # IDE system files
 .idea
 .vscode
+
+/go.work*
--- a/client_test.go
+++ b/client_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	pkgErrors "github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
@@ -870,4 +871,17 @@ func TestClientSetsUpTransport(t *testing.T) {
 
 	client, _ = NewClient(ClientOptions{})
 	require.IsType(t, &noopTransport{}, client.Transport)
+}
+
+func TestClientSetupInstrumenter(t *testing.T) {
+	client, err := NewClient(ClientOptions{Dsn: ""})
+	require.NoError(t, err)
+	assert.Equal(t, "sentry", client.Options().Instrumenter)
+
+	_, err = NewClient(ClientOptions{Dsn: "", Instrumenter: "foo"})
+	require.Error(t, err)
+
+	client, err = NewClient(ClientOptions{Dsn: "", Instrumenter: "otel"})
+	require.NoError(t, err)
+	assert.Equal(t, "otel", client.Options().Instrumenter)
 }
--- a/fiber/sentryfiber.go
+++ b/fiber/sentryfiber.go
@@ -74,26 +74,28 @@ func (h *handler) handle(ctx *fiber.Ctx) error {
 		sentry.WithSpanOrigin(sentry.SpanOriginFiber),
 	}
 
-	transaction := sentry.StartTransaction(
-		sentry.SetHubOnContext(ctx.Context(), hub),
-		fmt.Sprintf("%s %s", method, transactionName),
-		options...,
-	)
+	if hub.Client().Options().Instrumenter == "sentry" {
+		transaction := sentry.StartTransaction(
+			sentry.SetHubOnContext(ctx.Context(), hub),
+			fmt.Sprintf("%s %s", method, transactionName),
+			options...,
+		)
 
-	defer func() {
-		status := ctx.Response().StatusCode()
-		transaction.Status = sentry.HTTPtoSpanStatus(status)
-		transaction.SetData("http.response.status_code", status)
-		transaction.Finish()
-	}()
+		defer func() {
+			status := ctx.Response().StatusCode()
+			transaction.Status = sentry.HTTPtoSpanStatus(status)
+			transaction.SetData("http.response.status_code", status)
+			transaction.Finish()
+		}()
 
-	transaction.SetData("http.request.method", method)
+		transaction.SetData("http.request.method", method)
+		ctx.Locals(transactionKey, transaction)
+	}
 
 	scope := hub.Scope()
 	scope.SetRequest(convertedHTTPRequest)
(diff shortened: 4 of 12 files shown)


costela:
The windows test failure seems unrelated? Maybe flaky?

As for the coverage decrease, I could add some tests for this case if the approach seems reasonable.

costela:
Sorry for the delay; vacation got in the way :wink: 

My javascript is pretty rusty (no :crab: pun intended), but AFAICT the js integration works slightly differently. It [always uses an existing span context](https://github.com/getsentry/sentry-javascript/blob/dc653d0adb95198daace6975b2343fd1d4443847/packages/opentelemetry-node/src/spanprocessor.ts#L34) (similarly to the proposed change in this PR), but it also has an extra [`instrumenter` option](https://github.com/getsentry/sentry-javascript/blob/dc653d0adb95198daace6975b2343fd1d4443847/packages/node/src/sdk.ts#L159), unlike the go integration (and this PR, for that matter).

The extra option seems a bit error-prone, compared to the sample-func approach in this PR: the code has to remember to check against known values, instead of explicitly passing exactly what we want it to do. But that's just my subjective (and superficial!) take on it.

costela:
Ok, third time is a charm!

The `sentry.SpanOption` doesn't quite work, because it can't modify it's received context. So we have to go with yet another middleware, which sits as an adapter between otel and `sentryhttp.Handler`. 
It sets the transaction created via the span processor in the context passed to `sentryhttp.Handler`, so that it detects its existence and can react accordingly.

PTAL

costela:
@cleptric can you please allow the tests to run? :pray:

costela:
FYI: :repeat: rebased on master, please unlock tests :pray:

cleptric:
Sorry that it took us a while to get back to you.

When using Otel, you shouldn't use any Sentry performance-related features. This is why we added the `instrumenter` option on other SDKs, which would disable all Sentry instrumentation. In Go, we opted not to add this option, given the manual nature of performance tracing.

So instead of using Sentry's `net/http` instrumentation, you would use https://github.com/open-telemetry/opentelemetry-go-contrib.

Our propagator considers incoming trace headers, so distributed tracing also works.

costela:
> So instead of using Sentry's net/http instrumentation, you would use https://github.com/open-telemetry/opentelemetry-go-contrib.

So you mean skipping the [http middleware](https://pkg.go.dev/github.com/getsentry/sentry-go/http)? But then we'd also throw out error/panic tracking?

If that's the intended way to work with OTEL, then maybe I should change this PR to instead add an option to `sentryhttp.Options` to skip traces?

cleptric:
This is a good point about panic tracking I haven't thought about.

IMO, we could wrap this code https://github.com/getsentry/sentry-go/blob/master/http/sentryhttp.go#L97-L112 into an if condition that checks for `ClientOptions.EnableTracing`, so no transactions are created.

This might make it a bit easier than adding a new integration option.

costela:
@cleptric you mean disabling `ClientOptions.EnableTracing`? I think that alone won't quite work either: then *all* transactions get dropped (even if the OTEL propagator decides to sample them).

We could maybe refactor the sampling logic to respect sample decisions even if `EnableTracing` is `false`, but I fear that might be too surprising to users.

In the end we want `sentryhttp.Handler` to only start a transaction if none has been started already (e.g. by the otel propagator), but unfortunately I haven't found a nice way to pass that informatino along. The propagator can't modify the request context and their otherwise unaware of each other. That's why I ended up with the current suggestion in this PR. :confused:

greywolve:
I think adding the `instrumenter` option also seems to be consistent with the other sdks.

costela:
@cleptric @greywolve now this PR has an attempted implementation of the new `TracesInstrumenter` option. PTAL :pray:

client.go
```diff
@@ -133,6 +133,9 @@ type ClientOptions struct {
 	TracesSampleRate float64
 	// Used to customize the sampling of traces, overrides TracesSampleRate.
 	TracesSampler TracesSampler
+	// Which instrumentation to use for tracing. Either "sentry" or "otel" are supported.
+	// Setting this to "otel" will ignore TracesSampleRate and TracesSampler and assume sampling is performed by otel.
+	TracesInstrumenter string
```
cleptric, line 138: We should call it `Instrumenter` as in other SDKs.

client.go
```diff
@@ -301,6 +304,19 @@ func NewClient(options ClientOptions) (*Client, error) {
...
+	case "sentry":
+		// noop
+	case "otel":
+		// sampling is performed by the OpenTelemetry SDK
+		options.TracesSampleRate = 1.0
+		options.TracesSampler = nil
+	default:
+		return nil, fmt.Errorf("invalid value for TracesInstrumenter (supported are 'sentry' and 'otel'): %q", options.TracesInstrumenter)
+	}
+
```
cleptric, line 319: I'm not sure if we should do this. I would rather have people define this explicitly in their options.

cleptric (commented):
Added some comments. This also needs tests.

client.go
costela: you mean the `TracesSampleRate` and `TracesSampler`? Should we then error if the values are set to anything else? Doing neither seems a bit error-prone: `Instrumenter: "otel"` + `TracesSampleRate: 0` (default) would ignore otel sampling decisions and never send traces.

client.go
costela: Sure thing! Done!

client.go
cleptric: Are you sure it does this? The span processor actually takes the Otel sampling decision into account, taking priority over the traces sample rate and traces sampler.

costela: I believe so. That might be a slightly different issue, though. One I had addressed in a [previous commit](https://github.com/getsentry/sentry-go/commit/9b18d57b992a06b384986d5e6faffef1871a293c#diff-27f9ebae19f3d3f9ca89ecaccdc8d8472ba81f86392efcb74e386532d26ca18a).
The current span processor considers only the sampled decision of the _parent_, so it only actually respects it if the span is part of an incoming trace. If it's a new trace, without a parent, the otel span might be sampled, but the sentry span remains at `SampleUndefined`, which then means we fall back to the client-wide sample rate.

[The current test](https://github.com/getsentry/sentry-go/blob/4c8f48a520c411e05569b35b0ade96310cdfb7af/otel/span_processor_test.go#L121) doesn't catch it because both the sentry client and the otel stack are initialized with "always trace", so they both arrive at the same decision, even though the processor isn't doing anything.

I just added a commit with the fixed test _and_ the fixed code. PTAL.

costela: friendly ping :innocent:

cleptric: I'll take a look later this week.

We made similar changes for node.js in https://github.com/getsentry/sentry-javascript/pull/9203, but I need to dive into this whole topic a bit deeper to give you actionable feedback.

costela:
I want to add another perspective to this MR, that may mean the `ClientOptions.Instrumenter` is not enough: some client SDKs (e.g. sentry-cocoa) do not seem to support opentelemetry at all. These  transactions would be lost when speaking to a sentry+otel service. The service only reads otel headers, the client only sends sentry headers.

The suggestion with the additional middleware would not have this problem: incoming requests could use either sentry or otel propagation, the service code can leverage existing otel integrations (e.g. for [pgx](https://github.com/exaring/otelpgx/); shameless plug :see_no_evil:)

Unless sentry plans to support otel propagation on client sdks, it may be necessary to make the service side accept both. Or am I overlooking something?

cleptric:
We actually extract the Sentry trace header, see

https://github.com/getsentry/sentry-go/blob/master/otel/propagator.go#L77-L95

costela:
@cleptric ah, you're right of course. Appologies for the noise; disregard the last comment.

client.go
costela: friendly ping :innocent: (I know these things get lower prio; just wanna avoid it getting forgotten)

cleptric: Yeah, sorry for that! I had to get some stuff shipped over on the PHP side of things, but the next few weeks will be again Go focus time for me.

client.go
```diff
@@ -301,6 +304,14 @@ func NewClient(options ClientOptions) (*Client, error) {
 		options.MaxSpans = defaultMaxSpans
 	}
 
+	switch options.Instrumenter {
+	case "":
+		options.Instrumenter = "sentry"
+	case "sentry", "otel": // noop
+	default:
+		return nil, fmt.Errorf("invalid value for TracesInstrumenter (supported are 'sentry' and 'otel'): %q", options.Instrumenter)
```
greywolve, line 312:
```suggestion
		return nil, fmt.Errorf("invalid value for Instrumenter (supported are 'sentry' and 'otel'): %q", options.Instrumenter)
```

costela: oh, good catch! Thanks! :pray:

greywolve:
@costela, thinking more about the otel changes specifically. I think you're right and your fix seems correct, but wouldn't that be better as a separate PR since it's addressing an additional concern?

greywolve:
LGTM, but I'll let @cleptric give the final thumbs up :)

costela:
hi!
just a little bi-monthly rebase + reminder about this PR :innocent:

costela:
maybe I can get this PR as a christmas present? :innocent: :santa: :christmas_tree: 

(just rebased :repeat:)

costela:
Rebased again.

Any chance a paying customer can get some feedback? Even if this isn't the right solution?

cleptric:
It's still on our radar!

rodolfoBee:
@cleptric [customer has requested](https://sentry.zendesk.com/agent/tickets/123251) an update on this PR. Is there any?

aldy505:
@cleptric any updates for this?

giortzisg:
Closing this, since the span processor is [removed](https://github.com/getsentry/sentry-go/pull/1307) from 0.47.0 and on.

PR #682, opened 2023-08-01
fix(tests): Bump timeout to 15 minutes, update cache behavior

tonyo:
Windows builds are now sometimes taking more than 10 minutes :/


.github/workflows/test.yml
```diff
@@ -40,7 +40,6 @@ jobs:
           key: ${{ runner.os }}-go-${{ matrix.go }}-${{ hashFiles('**/go.sum') }}
           restore-keys: |
             ${{ runner.os }}-go-${{ matrix.go }}-
-            ${{ runner.os }}-go-
```
tonyo, line 43: Removing this because it looks like when caches are partially cleared, there could be some weird failures when e.g. a Windows build with go1.20 takes the cache from go1.18

cleptric (approved):

PR #684, opened 2023-08-01
fix: OTel incorrect usage of attributes.Value.AsString resulting in an empty string for non-string values

### LINKED ISSUE #683: [OTel] Non-string attributes are being sent as an empty string
## Summary
Non-string attributes are being sent as an empty string.

## Steps To Reproduce
* Configure an Open Telemetry TraceProvider using sentry's SpanProcessor
* Start a Span
* Set a non-string Attribute (such as `semconv.HTTPStatusCode`)
* Verify that in the dashboard this Attribute is an empty string, despite it's value being non-zero

## Expected Behavior
* All attributes should reach Sentry's Dashboard with their respective values.

### SDK
 - `sentry-go` version: v0.22.0
 - Go version: 1.19
 - Using Go Modules? yes

### Sentry
 - Anything particular to your environment that could be related to this issue? No.

## Additional context

The problem seems to be due to these two calls:
https://github.com/getsentry/sentry-go/blob/feb34956af657e570c5fbf629191ee6cd62a2b34/otel/span_processor.go#L125
https://github.com/getsentry/sentry-go/blob/feb34956af657e570c5fbf629191ee6cd62a2b34/otel/span_processor.go#L128

As stated by `Value.AsString` method's doc:

```
// AsString returns the string value. Make sure that the Value's type
// is STRING.
```

Probably the code's intention was to use `Value.Emit()`, instead. This method returns the string representation of the Value's data.

vtfr:
Fixes #683

### DIFF
--- a/otel/internal/utils/mapotelstatus.go
+++ b/otel/internal/utils/mapotelstatus.go
@@ -51,13 +51,13 @@ func MapOtelStatus(s trace.ReadOnlySpan) sentry.SpanStatus {
 
 	for _, attribute := range s.Attributes() {
 		if attribute.Key == semconv.HTTPStatusCodeKey {
-			if status, ok := canonicalCodesHTTPMap[attribute.Value.AsString()]; ok {
+			if status, ok := canonicalCodesHTTPMap[attribute.Value.Emit()]; ok {
 				return status
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


tonyo (approved):

tonyo:
Thanks @vtfr, this is great 👍

PR #694, opened 2023-08-09
Set SDK Name According Framework being used

### LINKED ISSUE #685: Set SDK Name According Framework being used
### Background
Sentry provides a number of framework integration with its Go SDK which allow for more specialized instrumentation in Go applications. However this is not reflected in the sdk name in the event meta data sent to Sentry. Most other SDKs do conform to this, even in python where auto-detection is in place for framework integration.

[Sentry Go Docs](https://docs.sentry.io/platforms/go/)

image

### Proposal
Sentry's Golang SDK will change the sdk name in the event (all events) meta data to reflect the framework integration being used in the application, to conform and be better situated to take advantage of any backend processing of events. This helps to know if users are getting the most out of sentry possible and which integrations are the most popular.

image

cleptric:
Closes #685

### DIFF
--- a/client.go
+++ b/client.go
@@ -407,7 +407,7 @@ func (client *Client) AddEventProcessor(processor EventProcessor) {
 }
 
 // Options return ClientOptions for the current Client.
-func (client Client) Options() ClientOptions {
+func (client *Client) Options() ClientOptions {
 	// Note: internally, consider using `client.options` instead of `client.Options()` to avoid copying the object each time.
 	return client.options
 }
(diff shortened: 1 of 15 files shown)


client.go
```diff
@@ -561,6 +564,14 @@ func (client *Client) EventFromCheckIn(checkIn *CheckIn, monitorConfig *MonitorC
 	return event
 }
 
+func (client *Client) SetSDKIdentifier(identifier string) {
+	sdkIdentifier = identifier
```
cleptric, line 568: Looking at https://github.com/getsentry/sentry-go/actions/runs/5807554425/job/15742607813, we might need a mutex here 🤔

tonyo (approved):

tonyo (approved):

PR #698, opened 2023-08-11
test: use mock transport to avoid many hanging worker routines

vaind:
I've noticed there are many HTTP transport worker goroutines running during tests (34 currently), and it was causing me some trouble when debugging profiling tests. 
This PR reduces the use of actual HTTP transport to tests that need to test it. Replaces with mock transport elsewhere.


tonyo (approved):
Let's do it 👍

PR #702, opened 2023-08-15
test: deduplicate Flush() timeout settings

vaind:
Follow up for #700 - I didn't notice there are multiple places we test the flush...


tonyo (approved):
Thanks 👍

PR #704, opened 2023-08-23
feat: Read release info from build info

DAcodedBEAT:
_Builds from #652_

For information about this change, please read the referenced PR above.  This PR is building off of the previous one and adds tests for coverage.

### DIFF
--- a/util_test.go
+++ b/util_test.go
@@ -1,6 +1,7 @@
 package sentry
 
 import (
+	"runtime/debug"
 	"testing"
 )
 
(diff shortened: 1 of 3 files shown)


util_test.go
```diff
@@ -27,3 +28,80 @@ func TestFileExistsReturnsFalseForNonExistingFiles(t *testing.T) {
 	assertEqual(t, fileExists(("util_nope.go")), false)
 	assertEqual(t, fileExists(("util_nope_test.go")), false)
 }
+
+func TestDefaultReleaseSentryReleaseEnvvar(t *testing.T) {
+	releaseVersion := "1.2.3"
+	cleanup := envSetter(map[string]string{
+		"SENTRY_RELEASE": releaseVersion,
+	})
```
aldy505, line 36: Why don't you use https://pkg.go.dev/testing#T.Setenv ?

util_test.go
```diff
@@ -27,3 +28,80 @@ func TestFileExistsReturnsFalseForNonExistingFiles(t *testing.T) {
...
+			},
+		},
+		Settings: []debug.BuildSetting{
+			{
+				Key:   "vcs.revision",
+				Value: releaseVersion,
+			},
+		},
+	}
+	cleanup := envSetter(map[string]string{})
```
aldy505, line 79: This line won't be necessary, right?

util_test.go
DAcodedBEAT: I liked the ability to set multiple environment variables in a single call but it's not strictly necessary - I can shift to using  `T.Setenv` instead :)

util_test.go
DAcodedBEAT: This line was added to ensure that no environment variables were accidentally set (locally or in CI) which could cause the release value to be different than the info provided in `debug.BuildInfo`

util_test.go
aldy505: It's a non-issue for me. One thing that would probably be an issue is flakiness on CI as doing set/un-set environment variable within a test scope won't enable the test suite to run in parallel. See the notes on https://pkg.go.dev/testing#T.Setenv

DAcodedBEAT: in the end both implementations are similar enough, I just moved this to `T.Setenv`

DAcodedBEAT:
@cleptric / @aldy505 Can I get a re-review?

tonyo (approved):

aldy505 (approved):

PR #707, opened 2023-08-31
Update package name parsing to match Go 1.20+ behavior

### LINKED ISSUE #180: Stacktrace for functions from "go.*" will have incorrect package name
https://github.com/getsentry/sentry-go/blob/4dddaaad5cc53f38de4928e98126dfbabd315423/stacktrace.go#L257-L262

This code as exptected will ignore parsing for functions which name starts from "go" however this is sometimes a valid package name. For example [zap](https://pkg.go.dev/go.uber.org/zap) library: `go.uber.org/zap/zapcore.(*CheckedEntry).Write` will not properly extract `go.uber.org/zap/zapcore` as a package name.

Since this code is copied from `debug/gosym` I filled same issue for original code https://github.com/golang/go/issues/37762

GregBrimble:
Go 1.20+ changed the compiler-generated symbols to use a colon rather than a period after `go` and `type` as the reserved prefixes (https://github.com/golang/go/commit/0f8dffd0aa71ed996d32e77701ac5ec0bc7cde01).

This PR inspects the current runtime version and opt-in to the relevant behavior when trying to split a qualified function at the package name.

Fixes #180

### DIFF
--- a/stacktrace.go
+++ b/stacktrace.go
@@ -1,6 +1,8 @@
 package sentry
 
 import (
+	"errors"
+	"fmt"
 	"go/build"
 	"reflect"
 	"runtime"
@@ -350,14 +352,44 @@ func callerFunctionName() string {
 	return baseName(callerFrame.Function)
 }
 
+func goMajorMinorVersion() (major, minor int, err error) {
+	releaseTags := build.Default.ReleaseTags
+	if len(releaseTags) == 0 {
+		return -1, -1, errors.New("no release tags")
+	}
+
+	// The last release tag is presumed to be the current release. The current
+	// release will be in the form 'go1.2.3'.
+	//
+	// See https://pkg.go.dev/go/build#Context
+	if _, err := fmt.Sscanf(releaseTags[len(releaseTags)-1], "go%d.%d", &major, &minor); err != nil {
+		return -1, -1, fmt.Errorf("sscanf: %w", err)
+	}
+
+	return major, minor, nil
+}
+
 // packageName returns the package part of the symbol name, or the empty string
 // if there is none.
 // It replicates https://golang.org/pkg/debug/gosym/#Sym.PackageName, avoiding a
 // dependency on debug/gosym.
 func packageName(name string) string {
-	// A prefix of "type." and "go." is a compiler-generated symbol that doesn't belong to any package.
-	// See variable reservedimports in cmd/compile/internal/gc/subr.go
-	if strings.HasPrefix(name, "go.") || strings.HasPrefix(name, "type.") {
+	// Since go1.20, a prefix of "type:" and "go:" is a compiler-generated symbol,
+	// they do not belong to any package.
+	//
+	// See cmd/compile/internal/base/link.go:ReservedImports variable.
+	major, minor, err := goMajorMinorVersion()
+	if err != nil {
+		return ""
+	}
+	goVersionGreaterThanOrEqualTo120 := major > 1 || (major == 1 && minor >= 20)
+
+	if goVersionGreaterThanOrEqualTo120 && (strings.HasPrefix(name, "go:") || strings.HasPrefix(name, "type:")) {
+		return ""
+	}
+
+	// For go1.18 and below, the prefix are "type." and "go." instead.
+	if !goVersionGreaterThanOrEqualTo120 && (strings.HasPrefix(name, "go.") || strings.HasPrefix(name, "type.")) {
 		return ""
 	}
 


stacktrace.go
```diff
@@ -350,14 +353,51 @@ func callerFunctionName() string {
...
+	if minor, err = strconv.Atoi(match[2]); err != nil {
+		return 0, 0, couldNotParseError
+	}
+
+	return major, minor, nil
+}
+
 // packageName returns the package part of the symbol name, or the empty string
 // if there is none.
 // It replicates https://golang.org/pkg/debug/gosym/#Sym.PackageName, avoiding a
```
jahands, line 382: This change brings this function in line with the `pkg/debug/gosym` package that this was imported from: https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/debug/gosym/symtab.go;l=64

stacktrace.go
```diff
@@ -350,14 +353,51 @@ func callerFunctionName() string {
 	return baseName(callerFrame.Function)
 }
 
+func goMajorMinorVersion() (int, int, error) {
```
uhthomas, line 356: fwiw I wrote this for something internally.

```go
// version gets the current version of Go from the runtime. The version may be
// in the forms:
//  - go1.17
//  - vdevel go1.17-9a8788c99d Tue Nov 16 10:53:28 2021 +0100
//
// Due to the stark difference in values, the version needs to be located and
// extracted.
func version() (string, error) {
	v := runtime.Version()
	for _, f := range strings.Fields(v) {
		const prefix = "go"
		if !strings.HasPrefix(f, prefix) {
			continue
		}
		f = f[len(prefix):]
		if i := strings.IndexRune(f, '-'); i != -1 {
			f = f[:i]
		}
		return f, nil
	}
	return "", fmt.Errorf("valid go version not found in %q", v)
}
```

It might be nice to avoid regular expressions where possible

uhthomas: I actually rewrote it as

```go
func version() (string, error) {
	releaseTags := build.Default.ReleaseTags
	if len(releaseTags) == 0 {
		return "", errors.New("no release tags")
	}
	// The last release tag is presumed to be the current release. The current
	// release will be in the form 'go1.2.3'.
	//
	// See https://pkg.go.dev/go/build#Context
	return strings.TrimPrefix(releaseTags[len(releaseTags)-1], "go"), nil
}
```

uhthomas: Then I guess you could extract the version like

https://go.dev/play/p/qTzf10VVF8k

```go
package main

import "fmt"

func main() {
	var major, minor int
	if _, err := fmt.Sscanf("go1.2.3", "go%d.%d", &major, &minor); err != nil {
		panic(err)
	}
	fmt.Println(major, minor)
}
```

So, altogether:

https://go.dev/play/p/U_DwYsnV9Nv

```go
package main

import (
	"errors"
	"fmt"
	"go/build"
)

func version() (major, minor int, err error) {
	releaseTags := build.Default.ReleaseTags
	if len(releaseTags) == 0 {
		return -1, -1, errors.New("no release tags")
	}
	// The last release tag is presumed to be the current release. The current
	// release will be in the form 'go1.2.3'.
	//
	// See https://pkg.go.dev/go/build#Context
	if _, err := fmt.Sscanf(releaseTags[len(releaseTags)-1], "go%d.%d", &major, &minor); err != nil {
		return -1, -1, fmt.Errorf("sscanf: %w", err)
	}
	return major, minor, nil
}

func main() {
	fmt.Println(version())
}
```

GregBrimble: Thanks!

tonyo (commented):
Hey,

Thanks for the fix!
The change looks reasonable and we'd like to move forward with it, but there are two things that we'd like to see updated:

1. `goMajorMinorVersion` memoization.
2. Could you add some tests for the change? [This test can be extended](https://github.com/getsentry/sentry-go/blob/8bf9a1582defccf747644e0155deac65a855ce21/stacktrace_test.go#L29) to test this new logic.

stacktrace.go
```diff
@@ -350,14 +352,44 @@ func callerFunctionName() string {
...
 // dependency on debug/gosym.
 func packageName(name string) string {
-	// A prefix of "type." and "go." is a compiler-generated symbol that doesn't belong to any package.
-	// See variable reservedimports in cmd/compile/internal/gc/subr.go
-	if strings.HasPrefix(name, "go.") || strings.HasPrefix(name, "type.") {
+	// Since go1.20, a prefix of "type:" and "go:" is a compiler-generated symbol,
+	// they do not belong to any package.
+	//
+	// See cmd/compile/internal/base/link.go:ReservedImports variable.
+	major, minor, err := goMajorMinorVersion()
```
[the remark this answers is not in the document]
greywolve: Hmm, i wonder if the complexity introduced by trying to memoize this (at least how things are currently written) is worth it at the moment?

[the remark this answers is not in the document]
greywolve: Seems like this does increase the overhead a fair but. Some quick benches seem to show this is around 3x slower.

greywolve:
I'm going to pick this one up.

GregBrimble:
Moved to #730

PR #709, opened 2023-09-04
Attach the `X-Sentry-Auth` header to /envelope requests

cleptric:
After internal discussions, we're adding back the `X-Sentry-Auth` header to requests to the `/envelope` endpoint as it's more efficient than the envelope header auth.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -211,6 +211,15 @@ func getRequestFromEvent(event *Event, dsn *Dsn) (r *http.Request, err error) {
 		if r != nil {
 			r.Header.Set("User-Agent", fmt.Sprintf("%s/%s", event.Sdk.Name, event.Sdk.Version))
 			r.Header.Set("Content-Type", "application/x-sentry-envelope")
+
+			auth := fmt.Sprintf("Sentry sentry_version=%s, "+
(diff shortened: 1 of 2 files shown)


transport.go
```diff
@@ -211,6 +211,15 @@ func getRequestFromEvent(event *Event, dsn *Dsn) (r *http.Request, err error) {
 		if r != nil {
 			r.Header.Set("User-Agent", fmt.Sprintf("%s/%s", event.Sdk.Name, event.Sdk.Version))
 			r.Header.Set("Content-Type", "application/x-sentry-envelope")
+
+			auth := fmt.Sprintf("Sentry sentry_version=%s, "+
+				"sentry_client=%s/%s, sentry_key=%s", apiVersion, event.Sdk.Name, event.Sdk.Version, dsn.publicKey)
```
cleptric, line 216: I removed `sentry_timestamp`, according to https://develop.sentry.dev/sdk/overview/#authentication

> This key is effectively deprecated, and it is ignored on the receiving side. Use the [sent_at envelope header](https://develop.sentry.dev/sdk/envelopes/#envelope-headers) instead.

tonyo (approved):

PR #711, opened 2023-09-06
fix(otel): prevent panic in shutdown

### LINKED ISSUE #692: [Otel] SpanProcessor panic on shutdown with background context
## Summary
I setup sentry with Opentelemetry, and on shutdown, the code panic with stack trace
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xa700f7]

goroutine 7211 [running]:
github.com/getsentry/sentry-go.(*Hub).stackTop(0xc000126000?)
	/go/pkg/mod/github.com/getsentry/sentry-go@v0.22.0/hub.go:109 +0x37
github.com/getsentry/sentry-go.(*Hub).Client(0x4a139c?)
	/go/pkg/mod/github.com/getsentry/sentry-go@v0.22.0/hub.go:136 +0x19
github.com/getsentry/sentry-go.(*Hub).Flush(0x47ab01?, 0xc0008cad50?)
	/go/pkg/mod/github.com/getsentry/sentry-go@v0.22.0/hub.go:347 +0x1e
github.com/getsentry/sentry-go/otel.flushSpanProcessor({0x146b420?, 0xc000d04b10?})
	/go/pkg/mod/github.com/getsentry/sentry-go/otel@v0.22.0/span_processor.go:108 +0x9a
github.com/getsentry/sentry-go/otel.(*sentrySpanProcessor).ForceFlush(...)
	/go/pkg/mod/github.com/getsentry/sentry-go/otel@v0.22.0/span_processor.go:101
github.com/getsentry/sentry-go/otel.(*sentrySpanProcessor).Shutdown(0x101eee0?, {0x146b420, 0xc000d04b10})
	/go/pkg/mod/github.com/getsentry/sentry-go/otel@v0.22.0/span_processor.go:96 +0x39
go.opentelemetry.io/otel/sdk/trace.(*TracerProvider).Shutdown.func1()
	/go/pkg/mod/go.opentelemetry.io/otel/sdk@v1.13.0/trace/provider.go:259 +0x39
sync.(*Once).doSlow(0xc00058ae80?, 0x4064f1?)
	/usr/local/go/src/sync/once.go:74 +0xc2
sync.(*Once).Do(...)
	/usr/local/go/src/sync/once.go:65
go.opentelemetry.io/otel/sdk/trace.(*TracerProvider).Shutdown(0xc000192600, {0x146b420, 0xc000d04b10})
	/go/pkg/mod/go.opentelemetry.io/otel/sdk@v1.13.0/trace/provider.go:258 +0x285

```
How i init client

```go

	batchProcessor := sdktrace.NewBatchSpanProcessor(exporter)
	sampler := sdktrace.ParentBased(sdktrace.TraceIDRatioBased(opts.SampleRate))
	propagator := propagation.NewCompositeTextMapPropagator(
		propagation.TraceContext{},
		b3.New(b3.WithInjectEncoding(b3.B3SingleHeader|b3.B3MultipleHeader)), // For Istio compatibility
	)
	tp := sdktrace.NewTracerProvider(
		sdktrace.WithSampler(sampler),
		sdktrace.WithSpanProcessor(batchProcessor),
		sdktrace.WithResource(resource.NewWithAttributes(
			semconv.SchemaURL,
			semconv.ServiceNameKey.String(opts.TracerName),
			semconv.ServiceVersionKey.String(opts.TracerVersion),
			attribute.String(environmentLabel, opts.Environment),
		)),
	)
	
        tp.RegisterSpanProcessor(sentryotel.NewSentrySpanProcessor())
	
	otel.SetTracerProvider(tp)
	otel.SetTextMapPropagator(propagator)
```
How i shutdown (ctx is background)
```

			if tp, ok := tp.(*sdktrace.TracerProvider); ok && tp != nil {
				return tp.Shutdown(ctx)
			}
```
This code panic because in the context there is no hub assigned, and the spanprocessor try to call stackTop on a nil hub, do i have to explicitly assign a hub into the main context?

### SDK
 - `sentry-go` version: 0.22.0
 - Go version: 1.20
 - Using Go Modules? [yes]

greywolve:
Shutdown panics when passed a background context, or any context which doesn't contain a hub. Returning the current hub if no hub is found on the context prevents this.

Fixes #692

### DIFF
--- a/otel/span_processor.go
+++ b/otel/span_processor.go
@@ -103,6 +103,10 @@ func (ssp *sentrySpanProcessor) ForceFlush(ctx context.Context) error {
 
 func flushSpanProcessor(ctx context.Context) error {
 	hub := sentry.GetHubFromContext(ctx)
+	if hub == nil {
+		hub = sentry.CurrentHub()
+	}
+
 	// TODO(michi) should we make this configurable?
 	defer hub.Flush(2 * time.Second)
 	return nil
--- a/otel/span_processor_test.go
+++ b/otel/span_processor_test.go
@@ -72,7 +72,7 @@ func TestNewSentrySpanProcessor(t *testing.T) {
 
 func TestSpanProcessorShutdown(t *testing.T) {
 	spanProcessor, _, tracer := setupSpanProcessorTest()
-	ctx := emptyContextWithSentry()
+	ctx := context.Background()
 	tracer.Start(emptyContextWithSentry(), "spanName")
 
 	assertEqual(t, sentrySpanMap.Len(), 1)


otel/span_processor.go
```diff
@@ -103,6 +103,10 @@ func (ssp *sentrySpanProcessor) ForceFlush(ctx context.Context) error {
 
 func flushSpanProcessor(ctx context.Context) error {
```
greywolve, line 104: I guess one question is if anyone is actually going to pass a context in here that contain a hub?

Seems more like this will almost always be something like `context.Background()' with no hub. I think the original purpose here, at least looking at the otel project, is to allow the user to pass in a context with a timeout. The flush method though already has a timeout, so context passed here isn't really useful to us.

tonyo: You're right, it's not ideal, and relying on people passing the right context is not a good approach. 

It might be still good to let people pass sentry.Hub via context (for more flexibility, and maybe in the future allowing multiple span processor instances?), so your fix should be good for now 👍

tonyo (approved):

PR #717, opened 2023-09-14
feat: support IgnoreTransactions client option

### LINKED ISSUE #628: Add ignoreTransactions and ignoreErrors
### Background:
To further simplfiy ignoring certain events from being sent to sentry, it is also suggested to provide `ignoreTransactions` and `ignoreErrors`

More info here develop docs updated: https://develop.sentry.dev/sdk/features/#event-sampling

- [ ] ignoreErrors
- [ ] ignoreTransactions
- [ ] SDK docs

greywolve:
Adds support for ignoring transactions by using the `IgnoreTransactions` client option. This requires a slice of regex strings that are matched against a transaction's name. If there's a match the transaction is dropped.

Fixes #628

### DIFF
--- a/client.go
+++ b/client.go
@@ -9,7 +9,6 @@ import (
 	"math/rand"
 	"net/http"
 	"os"
-	"regexp"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


tonyo (approved):
LGTM, thanks 👍 

Could you just add a bit of PR description and link to the issue it's fixing/implementing?

cleptric (changes requested):
We want to make sure this continues to work with the `ignoreErrorsIntegration` that is currently controlled via the `IgnoreErrors` option.

I would opt to make an `ignoreTransactionsIntegration`, as we can't remove the `ignoreErrorsIntegration`, as this would cause people issues for people using it today.

greywolve:
> We want to make sure this continues to work with the [`ignoreErrorsIntegration`](https://github.com/getsentry/sentry-go/blob/master/integrations.go#L120-L178) that is currently controlled via the `IgnoreErrors` option.
> 
> I suggest adding an `ignoreTransactionsIntegration`, as we can't remove the `ignoreErrorsIntegration`, which would cause issues for people using it today.

Hmm, my bad. I missed that `IgnoreErrors` was already a working integration. I'll rework this to just add an `IgnoreTransaction` integration, and not do any of the logic in `client.go`.

greywolve:
@cleptric rework done. :)

cleptric (approved):

Issue #331, opened 2021-02-19
SENTRY_TAGS_* environment variables


zerok:
## Summary

The sentry-logback implementation supports the injection of tags through the `SENTRY_TAGS_<key>=<value>` environment variables. It would be great if the Go SDK would support this too 🙂 

## Motivation

We would like to inject some additional context as tags into each event, especially things like the Kubernetes pod and Kubernetes namespace that event got triggered in. In Java services we have for that set `SENTRY_TAGS_K8S_NAMESPACE` and `SENTRY_TAGS_K8S_POD` and would like to do that also for our Go applications.

## Sample

I've added a test-case + sample implementation here: https://github.com/zerok/sentry-go/commit/36dc6dbb540ea558e0d1225f2f4493b584f0244c

Right now we can work around this using a custom eventprocessor but it would be nice if the Java and Go SDKs would behave in a similar way here out of the box 🙂

dennwc:
Is there anything missing in the sample implementation which prevents merging it upstream?

rhcarvalho:
@dennwc would you mind sharing what is your use case?

A quick look onto https://github.com/zerok/sentry-go/commit/36dc6dbb540ea558e0d1225f2f4493b584f0244c and I see some things would need to be better defined, for example, as implemented, passing `ClientOptions.Tags` when initializing the SDK would ignore the environment.

Also comes to mind that some SDKs have a concept of `InitialScope`, in that case tags could naturally be there.

This is really so simple to add with `EventProcessor`, and with total flexibility on behavior (e.g. add to all events, add to errors only, add to transactions only, etc) that I'm not seeing the value of maintaining a limited version which makes the API surface larger (we could read tags from env without adding new fields to ClientOptions).

dennwc:
It's actually very similar to what is described above: mostly passing K8S or other deployment-related variables.

So I agree that it might be okay to keep `ClientOptions` as-is. At least for our case we will pass everything via environment.

cleptric:
Closing, as the route via an `EventProcessor` seems to be feasible.

seanhoughton:
[...]

PR #718, opened 2023-09-19
feat: global tags

greywolve:
Add global tags support.

Tags can be passed via client options or via environment variables by prefixing them with "SENTRY_TAGS_".

Setting scope tags overrides global tags.

Client option tags override environment variable tags.

Fixes: #331

### DIFF
--- a/client.go
+++ b/client.go
@@ -224,7 +224,7 @@ type ClientOptions struct {
 	// is not optimized for long chains either. The top-level error together with a
 	// stack trace is often the most useful information.
 	MaxErrorDepth int
-	// Optional global tags
+	// Default event tags. These are overriden by tags set on a scope.
 	Tags map[string]string
 }
 
--- a/client.go
+++ b/client.go
@@ -224,7 +224,7 @@ type ClientOptions struct {
 	// is not optimized for long chains either. The top-level error together with a
 	// stack trace is often the most useful information.
 	MaxErrorDepth int
-	// Default event tags. These are overriden by tags set on a scope.
+	// Default event tags. These are overridden by tags set on a scope.
 	Tags map[string]string
 }
 
--- a/client.go
+++ b/client.go
@@ -224,6 +224,8 @@ type ClientOptions struct {
 	// is not optimized for long chains either. The top-level error together with a
 	// stack trace is often the most useful information.
 	MaxErrorDepth int
+	// Optional global tags
+	Tags map[string]string
 }
 
 // Client is the underlying processor that is used by the main API and Hub
@@ -375,6 +377,7 @@ func (client *Client) setupIntegrations() {
 		new(modulesIntegration),
 		new(ignoreErrorsIntegration),
 		new(ignoreTransactionsIntegration),
+		new(globalTagsIntegration),
 	}
 
 	if client.options.Integrations != nil {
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -224,6 +224,8 @@ type ClientOptions struct {
 	// is not optimized for long chains either. The top-level error together with a
 	// stack trace is often the most useful information.
 	MaxErrorDepth int
+	// Optional global tags
```
greywolve, line 227: Can probably give a better description here.

integrations.go
```diff
@@ -325,3 +326,62 @@ func (cfi *contextifyFramesIntegration) addContextLinesToFrame(frame Frame, line
...
+	tags    map[string]string
+	envTags map[string]string
+}
+
+func (ti *globalTagsIntegration) Name() string {
+	return "GlobalTags"
+}
+
+func (ti *globalTagsIntegration) SetupOnce(client *Client) {
+	ti.tags = client.options.Tags
```
tonyo, line 346: I'd suggest we make a deep copy here to avoid any surprises and potential concurrent access.

integrations.go
```diff
@@ -325,3 +326,62 @@ func (cfi *contextifyFramesIntegration) addContextLinesToFrame(frame Frame, line
...
+		return event
+	}
+
+	if event.Tags == nil {
+		event.Tags = make(map[string]string, len(ti.tags)+len(ti.envTags))
+	}
+
+	for k, v := range ti.tags {
+		if _, ok := event.Tags[k]; !ok {
+			event.Tags[k] = v
```
tonyo, line 362: Since we also update `event.Tags` in `scope.ApplyToEvent` (and potentially somewhere else?), please double check that we won't cause any "concurrent writes" panics around it.

integrations.go
greywolve: Good point :)

integrations.go
greywolve: This should be fine and in line with other integrations from what I can tell. 

They all tend to modify the event in some way. `event.Tags` doesn't seem special in this regard. `scope.ApplyToEvent` is called when the event is prepared as part of running each event processor. (the scope seems to have it's own slice of tags which is protected for concurrent access)

greywolve:
@tonyo comments addressed. :)

integrations.go
```diff
@@ -325,3 +326,67 @@ func (cfi *contextifyFramesIntegration) addContextLinesToFrame(frame Frame, line
...
+
+func loadEnvTags() map[string]string {
+	tags := map[string]string{}
+	for _, pair := range os.Environ() {
+		parts := strings.Split(pair, "=")
+		if !strings.HasPrefix(parts[0], envTagsPrefix) {
+			continue
+		}
+		tag := strings.TrimPrefix(parts[0], envTagsPrefix)
+		tag = strings.ToLower(tag)
```
tonyo, line 388: Let's not apply `ToLower` to the tag key, we should support keys with mixed casing.

tonyo (commented):
Noticed another small thing, otherwise LGTM 👍

integrations.go
greywolve: Fixed :)

integrations.go
tonyo: Sgtm, thanks for checking 👍

tonyo (approved):

cleptric (approved):
Nice one!

PR #723, opened 2023-10-01
feat(performance): start transaction for fasthttp integration

aldy505:
I realized that only a few integration starts performance span for their HTTP handlers (such as net/http and Gin), the others are not. Here is my effort on fixing it for other integrations/frameworks.

This also introduces the `GetTransactionFromContext` function that takes the `fasthttp.RequestCtx` as input and returns `*sentry.Span` or `nil`. It handles that `fasthttp.RequestCtx`'s context is not compatible with our `sentry.GetHubFromContext` due to different format of context key.

Discord link: https://discord.com/channels/621778831602221064/621786587939864586/1157114103156969513


ribice (approved):

fasthttp/sentryfasthttp.go
```diff
@@ -66,10 +68,33 @@ func (h *Handler) Handle(handler fasthttp.RequestHandler) fasthttp.RequestHandle
...
+		options := []sentry.SpanOption{
+			sentry.WithOpName("http.server"),
+			sentry.ContinueFromRequest(convertedHTTPRequest),
+			sentry.WithTransactionSource(sentry.SourceRoute),
+		}
+
+		method := string(ctx.Method())
+
+		transaction := sentry.StartTransaction(
+			sentry.SetHubOnContext(ctx, hub),
```
vaind, line 82: Maybe I'm missing something but this line doesn't seem to match the comment above:

```
		// Unlike for other integrations, we don't support getting an existing
		// hub from the current request context because fasthttp doesn't use the
		// standard net/http.Request and because fasthttp.RequestCtx implements
		// context.Context but requires string keys.
		hub := sentry.CurrentHub().Clone()
```

vaind (changes requested):

fasthttp/sentryfasthttp.go
aldy505: Yeah we shall remove this

vaind (approved):

PR #725, opened 2023-10-02
fix: profiler infinite loop after a long transaction

### LINKED ISSUE #724: Infinite Loop in the Transaction Profiler
## Summary

The transaction profiler `profiler.go` loops infinite if the transaction duration exceeds the `profilerRuntimeLimit`, causing the application to terminate due to OOM.

## Steps To Reproduce

```go
func main() {
	err := sentry.Init(sentry.ClientOptions{EnableTracing: true, ProfilesSampleRate: 1})
	if err != nil {
		fmt.Printf("sentry.Init: %s", err.Error())
		return
	}

	span := sentry.StartSpan(context.Background(), "main", sentry.WithSpanSampled(sentry.SampledTrue))
	time.Sleep(35 * time.Second)
	span.Finish()
}
```

## Expected Behavior

Either:
- The Sentry SDK is able to profile also transactions lasting longer than 30 seconds without crashing the application.
- The Sentry SDK discards the profiling of transactions exceeding the `profilerRuntimeLimit`.

### SDK
 - `sentry-go` version: v0.24.1
 - Go version: 1.21.1
 - Using Go Modules? yes

### Sentry
 - No Sentry Backend required to reproduce the behavior.

## Additional context

The following loop runs infinite:
https://github.com/getsentry/sentry-go/blob/7e37edeb9589c5e8966c5c80822aa4ddaa0fbfd2/profiler.go#L243-L251

The `profilerRuntimeLimit` defines a limit of 30 seconds. If the transaction lasts longer, the `start`-30-second-ring-structure is filled completely which causes both break conditions to not apply.

vaind:
Fixes #724

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Bug fixes
+
+- Profiler infinite loop after a transaction longer than the limit of 30 seconds ([#725](https://github.com/getsentry/sentry-go/pull/725))
+
 ## 0.24.1
 
 The Sentry SDK team is happy to announce the immediate availability of Sentry Go SDK v0.24.1.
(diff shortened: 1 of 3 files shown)


cleptric (approved):
Thanks! 🙂

Issue #686, opened 2023-08-03
add a per event sample rate.


james-lawrence:
## Summary
the global sample rate is wonderful for a default setting but misses the mark for actual usage. Its a fact of life that events are more critical than others and you'll want to sample those are higher frequencies.

## Motivation
getting actionable alerts faster without blowing up overall usage.

## Additional Context
this is actually applicable to most global options (like attaching a stack trace)

cleptric:
There are quite a few mechanisms in place you can use to downsample specific errors.
Have a look at https://docs.sentry.io/platforms/go/configuration/filtering/.

james-lawrence:
that's a lot of work to offload to end users using a globally scoped functionality for something that should be straight forward to mark in a apm and will lead to spooky action at a distance behaviors in the code base.

james-lawrence:
@cleptric as I mentioned SampleRate/TracesSampleRate are global; not all errors/traces are equal. having the ability to set the sample rate per event/trace would resolve the issue here. the 'solutions' you mentioned are all much larger maintenance burdens for end users of sentry. we're likely talking < 10 lines of code to handle this gracefully (assuming reasonable internals in the client sdks) which is to add the sample rate to the events themselves and default the value from the global hub.

PR #726, opened 2023-10-02
RFC: implement per event sample rates.

james-lawrence:
opening this to move discussion forward on #686 which i felt was closed without adequately addressing the problem being brought forth. if the idea is acceptable i'll gladly add tests etc.

pros:
- allows setting sample rates per event. meaning we don't need complicated interactions between sample rates and beforesend.
- no need for a global filter to do basic sampling.
- maintains default behavior.

### DIFF
--- a/client.go
+++ b/client.go
@@ -607,7 +607,7 @@ func (client *Client) processEvent(event *Event, hint *EventHint, scope EventMod
 	// Transactions are sampled by options.TracesSampleRate or
 	// options.TracesSampler when they are started. All other events
 	// (errors, messages) are sampled here.
-	if event.Type != transactionType && !sample(client.options.SampleRate) {
+	if event.Type != transactionType && !sample(firstNonNegativeRate(event.sampleRate, client.options.SampleRate)) {
 		Logger.Println("Event dropped due to SampleRate hit.")
 		return nil
 	}
@@ -737,4 +737,14 @@ func (client *Client) integrationAlreadyInstalled(name string) bool {
 // [0.0, 1.0].
 func sample(probability float64) bool {
 	return rng.Float64() < probability
+}
+
+func firstNonNegativeRate(rates ...float64) float64 {
+	for _, r := range rates {
+		if r >= 0.0 {
+			return r
+		}
+	}
+
+	return 0.0
 }
--- a/interfaces.go
+++ b/interfaces.go
@@ -335,6 +335,13 @@ type Event struct {
 
 	sdkMetaData SDKMetaData
 	attachments []*Attachment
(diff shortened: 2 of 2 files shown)


cleptric:
Instead of this, we would be open to adding an `ErrorsSampler` with similar behaviour as the [`TracesSampler`](https://docs.sentry.io/platforms/go/configuration/filtering/#using-platformidentifier-nametraces-sampler-).

james-lawrence:
that wouldn't doesn't resolve the issue i'm putting forth. TracesSampler has the exact same problem as SampleRate (its global and doesn't let the decision happen at the time of creation of the event).

cleptric:
I don't see a problem with this, and I'm still trying to figure out precisely what you are trying to solve.
As a Sentry user, I'm likely aware of which transactions are less interesting and can return different sample rates based on the transaction name. A similar approach would work fine with errors as well.

james-lawrence:
lets assume I have 10k types of events I send through to sentry. for the majority of them the sample rate is whatever. but there is a subset say 200 or so that I *really* need to see always. I don't want to have to implement a global filter that detects those 200 when I literally just need to set the sample rate to the value I need it to be for those 200 events.

it also ensures that the setting of the sample rate happens *where* the event is generated vs disconnected in this global client settings option that now needs to know about everything usually through some convention that everyone has to define in their codebase.

cleptric:
Are you creating the events manually? I fail to see how your approach works with `sentry.CaptureMessage` etc. Can you post a snippet of what the usage of your proposed API looks like?

james-lawrence:
yes I would be creating events manually; but if i wanted to expect this to CaptureMessage I'd just add functional options to the method call.

```golang
func CaptureMessage(m string, options ...func(*Event))
```

```golang
CaptureMessage("foo", sentry.EventOptionSampleRate(1.0))
```

cleptric:
The outcome is the same. You either need to define the sample rate on each event or centralized in the `ErrorsSampler`. I want us to stay somewhat unified with other SDKs here, so I'm open to adding the `ErrorsSamler` but not a unique sample rate as proposed here.

james-lawrence:
i have this same problem in the other SDKs too. ;)

cleptric:
> i have this same problem in the other SDKs too. ;) and this isn't an either or. this implemention supports a global default and a per event override.

Yes, and you either need to change all the places in your codebase to add the event-based sample rate or do this centralized in `ErrorsSampler`.

I disagree with the argument that this adds a larger maintenance burden.

We're open to merging a PR that adds an `ErrorsSampler` but won't move forward with the suggested sample rate per event.

PR #730, opened 2023-10-05
fix(stacktrace): support package name parsing for Go 1.20+

### LINKED ISSUE #180: Stacktrace for functions from "go.*" will have incorrect package name

greywolve:
Go 1.20+ changed reserved prefixes for compiler-generated symbols from "go." and "type." to "go:" and "type:" respectively.

See: https://github.com/golang/go/commit/0f8dffd0aa71ed996d32e77701ac5ec0bc7cde01a

This adds support for both kinds of compiler-generated symbols depending on the Go version by taking advantage of build tags.

Fixes #180 

Inspired by and continued on from: https://github.com/getsentry/sentry-go/pull/707

### DIFF
--- a/stacktrace.go
+++ b/stacktrace.go
@@ -355,9 +355,7 @@ func callerFunctionName() string {
 // It replicates https://golang.org/pkg/debug/gosym/#Sym.PackageName, avoiding a
 // dependency on debug/gosym.
 func packageName(name string) string {
-	// A prefix of "type." and "go." is a compiler-generated symbol that doesn't belong to any package.
-	// See variable reservedimports in cmd/compile/internal/gc/subr.go
-	if strings.HasPrefix(name, "go.") || strings.HasPrefix(name, "type.") {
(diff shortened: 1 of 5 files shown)


GregBrimble:
Thanks! Appreciate you picking that up for me. I'll close mine out.

cleptric (approved):
:shipit:

PR #733, opened 2023-10-13
build(deps): manually bump golang.org/x/net from 0.8.0 to 0.17.0

greywolve:
Tests were failing due to the otel module also needing a `go mod tidy`, which dependabot didn't handle.

Replaces: https://github.com/getsentry/sentry-go/pull/732


cleptric:
We normally don't bump the baseline, see https://github.com/getsentry/sentry-go/security/policy

greywolve:
Closing , not important enough for an exception to the policy.

PR #736, opened 2023-10-17
ref: remove Span.isTransaction attribute

### LINKED ISSUE #581: Remove Span.isTransaction attribute
We already have `Span.IsTransaction()` method that can be rewritten by checking [`Span.recorder`](https://github.com/getsentry/sentry-go/blob/f03b31a4a314fda36229c131a20306ffd686f2e9/tracing.go#L54) in constant time, which basically makes [`Span.isTransaction`](https://github.com/getsentry/sentry-go/blob/f03b31a4a314fda36229c131a20306ffd686f2e9/tracing.go#L52)) redundant. Thus, `Span.isTransaction` can be eventually removed, reducing the number of internal Span states.

greywolve:
Instead we rely on a new implementation of IsTransaction() that checks if the span's parent is nil.

Fixes #581

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -167,14 +167,14 @@ func StartSpan(ctx context.Context, operation string, options ...SpanOption) *Sp
 		option(&span)
 	}
 
+	span.Sampled = span.sample()
+
 	if hasParent {
 		span.recorder = parent.spanRecorder()
 	} else {
 		span.recorder = &spanRecorder{}
 	}
 	span.recorder.record(&span)
-
-	span.Sampled = span.sample()
 
 	hub := hubFromContext(ctx)
 
--- a/tracing.go
+++ b/tracing.go
@@ -253,9 +253,8 @@ func (s *Span) SetContext(key string, value Context) {
 func (s *Span) IsTransaction() bool {
 	if s.recorder != nil {
 		return s == s.recorder.root()
-	} else {
-		return false
 	}
+	return false
 }
 
 // GetTransaction returns the transaction that contains this span.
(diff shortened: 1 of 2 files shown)


tracing_test.go
```diff
@@ -546,12 +542,16 @@ func TestContinueTransactionFromHeaders(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		s := &Span{isTransaction: true}
+		s := &Span{}
+		s.recorder = &spanRecorder{}
```
greywolve, line 546: This does make testing more awkward. Could probably have a test helper to create a span that's a transaction.

tracing.go
```diff
@@ -173,15 +167,15 @@ func StartSpan(ctx context.Context, operation string, options ...SpanOption) *Sp
...
-	span.Sampled = span.sample()
-
 	if hasParent {
 		span.recorder = parent.spanRecorder()
 	} else {
 		span.recorder = &spanRecorder{}
 	}
 	span.recorder.record(&span)
 
+	span.Sampled = span.sample()
```
greywolve, line 177: This has to happen after we record the span, since the sample method checks IsTransaction()

tracing.go
```diff
@@ -257,7 +251,10 @@ func (s *Span) SetContext(key string, value Context) {
 
 // IsTransaction checks if the given span is a transaction.
 func (s *Span) IsTransaction() bool {
-	return s.isTransaction
+	if s.recorder != nil {
+		return s == s.recorder.root()
+	}
+	return false
```
cleptric, line 257: Can't we perform the same check as before?

```go
isTransaction: !hasParent
```

greywolve: Ah silly me. That's much simpler. I'll rework this all. :)

cleptric (approved):

Issue #731, opened 2023-10-09
goroutine leak in HTTPTransport


kucherenkovova:
## Summary
You [iterate over a channel](https://github.com/getsentry/sentry-go/blob/master/transport.go#L462) in a [separate goroutine](https://github.com/getsentry/sentry-go/blob/master/transport.go#L343). This channel is never closed which leads to a goroutine leak. 

## Steps To Reproduce

Install `"go.uber.org/goleak"` and add the following code to `transport_test.go` file
```go
package sentry

import (
        ...
	"go.uber.org/goleak"
)

...

func TestHTTPTransportDoesntLeakGoroutines(t *testing.T) {
	// verify that we're not leaking any goroutine
	defer goleak.VerifyNone(t)

	// init and configure transport
	transport := NewHTTPTransport()
	transport.Configure(ClientOptions{
		Dsn:        "https://test@foobar/1",
		HTTPClient: http.DefaultClient,
	})
	// use transport if needed

	// the transport is not needed anymore, get rid of it
	transport = nil

	// do some other work here
}
```
output:
```
=== RUN   TestHTTPTransportDoesntLeakGoroutines
    transport_test.go:701: found unexpected goroutines:
        [Goroutine 20 in state chan receive, with github.com/getsentry/sentry-go.(*HTTPTransport).worker on top of the stack:
        goroutine 20 [chan receive]:
        github.com/getsentry/sentry-go.(*HTTPTransport).worker(0x140002478f0)
        	/path/to/sentry-go/transport.go:471 +0xb8
        created by github.com/getsentry/sentry-go.(*HTTPTransport).Configure.func1 in goroutine 19
        	/path/to/sentry-go/transport.go:343 +0x60
        ]
```

## Expected Behavior

As a user, I want to be able to tear down all the sentry-allocated resources when it's no longer needed.

### SDK
 - `sentry-go` version: v0.25.0
 - Go version: go version go1.20.9
 - Using Go Modules? yes

cleptric:
Could you explain your use case in a bit more detail?
By default, the SDK is supposed to run all the time so it can capture errors and send performance data.

kucherenkovova:
Hi @cleptric! We don't explicitly instantiate `HTTPTransport` in our app.
Our initialization code looks like this:
```
hub := sentry.NewHub(nil, sentry.NewScope())
client, _ := sentry.NewClient(opts)
hub.BindClient(client)
...
hub.CaptureException(err)
```

[...]

PR #744, opened 2023-11-02
ci: improve flaky tests

vaind:
Some tests are flaky if the machine is overloaded - some goroutines may not get captured consistently on all ticks. And since we cannot control the number of goroutines running (HTTP workers from other tests hang around indefinitely, see https://github.com/getsentry/sentry-go/issues/731). Disabling the flaky test in CI when race detection is on. We can enable it once #731 is fixed.

The other flaky test I'm not sure about so just adding logs and maybe it would hep figure out what's wrong:

```
--- FAIL: TestProfilerPanicOnTick (0.00s)
    profiler_test.go:339: 
        	Error Trace:	/home/runner/work/sentry-go/sentry-go/profiler_test.go:339
        	            				/home/runner/work/sentry-go/sentry-go/profiler_test.go:263
        	Error:      	Should NOT be empty, but was 
        	Test:       	TestProfilerPanicOnTick
```

#skip-changelog

### DIFF
--- a/profiler_test.go
+++ b/profiler_test.go
@@ -338,7 +338,7 @@ func validateProfile(t *testing.T, trace *profileTrace, duration time.Duration) 
 	}
 
 	for i, frame := range trace.Frames {
-		if jsonData, e := json.Marshal(frame); e == nil {
+		if jsonData, err := json.Marshal(frame); err == nil {
 			t.Logf("Frame %d: %v", i, string(jsonData))
 		}
 
--- a/profiler_test.go
+++ b/profiler_test.go
@@ -1,6 +1,7 @@
 package sentry
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"math/rand"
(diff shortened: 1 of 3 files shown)


profiler_test.go
```diff
@@ -317,29 +318,34 @@ func findPrimeNumber(n int) int {
...
+		assert.NotEmpty(thread.Name)
 	}
 
-	for _, frame := range trace.Frames {
-		require.NotEmpty(frame.Function)
-		require.NotContains(frame.Function, " ") // Space in the function name is likely a parsing error
-		require.Greater(len(frame.AbsPath)+len(frame.Filename), 0)
-		require.Greater(frame.Lineno, 0)
+	for i, frame := range trace.Frames {
+		if jsonData, e := json.Marshal(frame); e == nil {
```
greywolve, line 341:
```suggestion
		if jsonData, e := json.Marshal(frame); err == nil {
```

greywolve (approved):
LGTM

PR #747, opened 2023-11-17
Allowing platform and SDK overrides

RiotHaxwagon:
This change will allow overrides to the reported platform and SDK.  This is useful if you're using the go library to convert and push other kinds of crashes (i.e. not go program crashes).  Additionally, this allows you to add Event-specific attachments beyond the shared scope's.

### DIFF
--- a/client.go
+++ b/client.go
@@ -673,15 +673,19 @@ func (client *Client) prepareEvent(event *Event, hint *EventHint, scope EventMod
 		event.Environment = client.options.Environment
 	}
 
-	event.Platform = "go"
-	event.Sdk = SdkInfo{
-		Name:         client.GetSDKIdentifier(),
-		Version:      SDKVersion,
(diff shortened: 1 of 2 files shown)


cleptric:
All of this can be achieved using `BeforeSend`. https://docs.sentry.io/platforms/go/configuration/filtering/#using-platformidentifier-namebefore-send-

RiotHaxwagon:
This is not the same behavior.  I shouldn't have to modify the event with a global handler.  In this case I'd have to hold onto the attachments instead of just adding when constructing the event.

RiotHaxwagon:
The first part of the change, I, indeed, can achieve with BeforeSend.  I will make that change to just be attachment modification

RiotHaxwagon:
https://github.com/getsentry/sentry-go/pull/768

PR #752, opened 2023-12-11
feat(tracing without performance): add ContinueTrace

greywolve:
Add ContinueTrace as a method on the hub. This is how other SDKs that have to deal with concurrency tend to do it, ie Java.


scope.go
```diff
@@ -400,6 +401,31 @@ func (scope *Scope) ApplyToEvent(event *Event, hint *EventHint) *Event {
 		}
 	}
 
+	// Apply the trace context to errors if there is a Span on the scope. If
+	// there isn't then fall back to the propagation context.
+	if event.Type != transactionType {
```
cleptric, line 406: Could we also use this for transactions, so it's centralized?

dynamic_sampling_context.go
```diff
@@ -121,3 +121,47 @@ func (d DynamicSamplingContext) String() string {
...
+	if release := client.options.Release; release != "" {
+		entries["release"] = release
+	}
+	if environment := client.options.Environment; environment != "" {
+		entries["environment"] = environment
+	}
+
+	if userSegment := scope.user.Segment; userSegment != "" {
+		entries["user_segment"] = userSegment
+	}
```
cleptric, line 161: We can remove this, we decided to remove this feature from the product/SDKs.

scope.go
greywolve: Yeah, looking at the code now that makes sense. I'll make the change. :)

PR #754, opened 2023-12-18
Only apply `SampleRate ` to errors/messages

cleptric:
We should only sample errors/messages. This fixes the `SampleRate` being applied to `check_in` as well.


sl0thentr0py (approved):

greywolve (approved):

greywolve:
@cleptric some tests need to be fixed it seems

PR #766, opened 2024-01-22
Version update due to retracted versions.

UnaffiliatedCode:
Retracted versions are causing issues in gorelease (go release tool from Google).


cleptric (changes requested):
Only bump the dependencies required for the fix you have in mind.

cleptric:
I'm not sure I can follow what you try to achieve here. As you now also bumped the minimum required Go version, I'm closing this.

UnaffiliatedCode:
> I'm not sure I can follow what you try to achieve here. As you now also bumped the minimum required Go version, I'm closing this.

Happy to answer.

We utilize this library in our application, due to bluemonday's published retraction of the current version (0.23), we need the bluemonday dependency updated to 0.26.

From the commit message:
```
Retracted versions are causing issues in gorelease (go release tool from Google).
This is the minimum update to apply for updating bluemonday from 0.23 to 0.26.
"github.com/kataras/iris/v12" required bluemonday in order to function.

This can be replicated locally by doing the following:

go get -u "github.com/kataras/iris/v12"
go mod tidy
```

If updating the minimum Go version is an option, all for it. But it doesn't appear so (but happy to learn of an alternative).

You can replicate our error by using any "hello world" application which has this library as a dependency.


Iris has been maintaining their dependencies by using dependabot, and upgraded this dependency here:
858136eafb4291b912b114dda76907ae32be8154



Additional documentation of this issue can be found here:
https://github.com/go-pkgz/auth/issues/183
https://github.com/microcosm-cc/bluemonday/discussions/194

PR #768, opened 2024-02-05
Allow custom attachments to be added to Events

RiotHaxwagon:
This change allows you to add specific attachments to the Event (without solely relying on scope-based attachments)


cleptric:
I don't think we need these methods, so I just exported the struct field instead in https://github.com/getsentry/sentry-go/pull/771.

Issue #643, opened 2023-05-27
Refactor Envelope build & sending


vaind:
## Summary
Currently, events are handed over to transport for sending. In order to fully switch to Envelopes, transport should not know about events and instead accept envelopes for sending. Transport should also not do any event-type specific processing - all the information transport needs is contained in the Envelope & its Items. See https://develop.sentry.dev/sdk/envelopes/

## Motivation
Besides finishing the move to Envelopes, the separation of concerns would allow us to build envelopes with multiple items, without hacking around events, as I've had to do in [here](https://github.com/getsentry/sentry-go/pull/626/files/8417a7bdf3462217a959ea9e1f4e2e0876dc235f#diff-5982a4ca350449c8a5deba675bc4bc4bb0f1823c45e6c7ba0076cc6b159fef8fR156) - adding a `transactionProfile` field to `Event` and handling that in `transport.go`.

## Additional Context
Ideally, I'd also suggest something to think about in the future - changing the APIs to not accept `Event` struct but rather an interface, sth. like which just knows how to serialize and deserialize itself. This would also make it more flexible and avoid adding fields specific to different event types, e.g. Transactions, as there already are.

PR #789, opened 2024-02-26
feat(metrics): add basic metrics type support

viglia:
this PR aims at adding metrics type definition and the related serialization logic.

This does not include high-level APIs that will be used by the users (with proper aggregation/bucketing).

Example of how to send metrics through the `Transport` struct: 
```go
package main

import (
	"time"

	"strings"

	"github.com/google/uuid"
)

func main() {
	id := strings.Replace(uuid.New().String(), "-", "", -1)
	e := sentry.NewEvent()
	e.EventID = sentry.EventID(id)

	e.Type = "statsd"
	t := time.Now().Unix()
	release := "1.0"
	env := "dev"
	e.Metrics = []sentry.Metric{
		sentry.NewCounterMetric("test-go-metric", sentry.Second(), map[string]string{"transaction": "go-metrics", "release": release, "env": env, "foo": "bar"}, t, 1.0),
		sentry.NewCounterMetric("test-go-metric", sentry.Second(), map[string]string{"transaction": "go-metrics", "release": release, "env": env, "route": " /foo"}, t, 1.0),
	}
	tr := sentry.NewHTTPSyncTransport()
	tr.Configure(sentry.ClientOptions{
		Dsn:   "<YOUR_DSN_HERE>",
		Debug: true,
	})

	tr.SendEvent(e)

}
```


metrics.go
```diff
@@ -0,0 +1,354 @@
...
+	"regexp"
+	"sort"
+	"strings"
+)
+
+type (
+	MetricUnit int
+
+	NumberOrString interface {
+		int | string
```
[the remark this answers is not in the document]
viglia: @cleptric this is only for the set metric. Do we want to use `float64` even in that case?

I thought it was only `int & string` for set.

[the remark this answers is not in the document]
cleptric: ok, this is fine then

metrics.go
```diff
@@ -0,0 +1,416 @@
...
+		re.ReplaceAllString(unit, ""),
+	}
+}
+
+type Metric interface {
+	GetType() string
+	GetTags() map[string]string
+	GetKey() string
+	GetUnit() string
+	GetTimestamp() int
```
aldy505, line 182: Why does timestamp is in `int` instead of `int64`? In Go, `int` does not always refers to `int64`, on certain machines it's `int32`. Should also add more declaration documentation of what this is: unix second? unix millisecond? unix microsecond?

metrics.go
```diff
@@ -0,0 +1,416 @@
...
+
+func (c CounterMetric) GetType() string {
+	return "c"
+}
+
+func (c CounterMetric) SerializeValue() string {
+	return fmt.Sprintf(":%v", c.value)
+}
+
+func NewCounterMetric(key string, unit MetricUnit, tags map[string]string, timestamp int, value float64) CounterMetric {
```
aldy505, line 249: I'd prefer `timestamp` to be `time.Time` instead of `int` or any derivative of integers. The SDK should serialize it, instead of throwing it to users.

metrics.go
```diff
@@ -0,0 +1,416 @@
...
+}
+
+func (d DistributionMetric) GetType() string {
+	return "d"
+}
+
+func (d DistributionMetric) SerializeValue() string {
+	var sb strings.Builder
+	for _, el := range d.values {
+		sb.WriteString(fmt.Sprintf(":%v", el))
```
aldy505, line 280: Why uses `%v` when we already know that `d.values` is a float64 slice?

metrics.go
```diff
@@ -0,0 +1,416 @@
...
+	g.sum += value
+	g.count++
+}
+
+func (g GaugeMetric) GetType() string {
+	return "g"
+}
+
+func (g GaugeMetric) SerializeValue() string {
+	return fmt.Sprintf(":%v:%v:%v:%v:%v", g.last, g.min, g.max, g.sum, g.count)
```
aldy505, line 322: The same goes with this, regarding the usage of `%v`

metrics.go
```diff
@@ -0,0 +1,416 @@
...
+	return SetMetric[T]{
+		map[T]void{
+			value: member,
+		},
+		am,
+	}
+}
+
+func sanitizeKey(s string) string {
+	re := regexp.MustCompile(`[^a-zA-Z0-9_/.-]+`)
```
aldy505, line 399: This regex will be compiled everytime the `sanitizeKey` is called, why not put it outside as a private global variable?

metrics.go
```diff
@@ -0,0 +1,416 @@
...
+	return re.ReplaceAllString(s, "_")
+}
+
+func sanitizeValue(s string) string {
+	re := regexp.MustCompile(`[^\w\d\s_:/@\.{}\[\]$-]+`)
```
aldy505, line 404: This too (regarding the regex).

aldy505 (changes requested):
To be honest, this metrics implementation would be so easy if we did https://github.com/getsentry/sentry-go/issues/643 first.

metrics.go
viglia: because this way, when serializing to a string, a float of `1` get serialized to  `1` instead of `1.000000`

aldy505: @viglia Mind to put a comment on the code about it? That would be good for people reading the code in the future

metrics.go
viglia: This is a fair suggestion, but this structs and functions that we're adding in this PR are just the foundation for metrics support and won't be used by users mostly. 

Users will use the high level APIs that will be added later in a different PR. So that's where we could accept `Time` and make the appropriate conversion.

aldy505:
Any updates for this?

cleptric (approved):

Issue #693, opened 2023-08-08
Support Go 1.20 Unwrap() with a slice of errors


ainar-g:
## Summary
Go 1.20 added another way to unwrap errors, `interface { Unwrap() []error }`.  As far as I can see, as of now `Event.SetException` doesn't support it.

https://github.com/getsentry/sentry-go/blob/cda691d09673fe847351b41adb4c65012cb6a5a9/interfaces.go#L356-L363

## Motivation
Developers are updating to Go 1.20, as Go 1.19 is EOL now, and they will expect sentry-go to support the new interface and include all errors into the event data.

## Additional Context
* https://pkg.go.dev/errors@go1.20
* golang/go#53435

tonyo:
Thanks, makes sense, added to the backlog.

ainar-g:
The Python API seems to already have a handler for a similar mechanism called `ExceptionGroup`:

https://github.com/getsentry/sentry-python/blob/641822dcf3cc90ee0c3e9726d4a5a979d4755c10/sentry_sdk/utils.py#L815-L832

Perhaps the same flatting algorithm could be used?

pierrre:
I'm not very familiar with Python, sorry :sweat_smile: 
Is it adding recursively all errors to the array of exception ?

ainar-g:
Neither do I really, heh.  My point was more that there are similar idioms in other languages, and when the Sentry team starts working on the issue, they should probably follow the same principles as the APIs in those languages do.

And yes, it seems to add them one after the other, at least based on what I can read.

pierrre:
This ?
```go
// SetException appends the unwrapped errors to the event's exception list.
//
// maxErrorDepth is the maximum depth of the error chain we will look
// into while unwrapping the errors.
func (e *Event) SetException(exception error, maxErrorDepth int) {
	err := exception
	if err == nil {
		return
	}

	e.setException(err, maxErrorDepth)

	// Add a trace of the current stack to the most recent error in a chain if
	// it doesn't have a stack trace yet.
	// We only add to the most recent error to avoid duplication and because the
	// current stack is most likely unrelated to errors deeper in the chain.
	if e.Exception[0].Stacktrace == nil {
		e.Exception[0].Stacktrace = NewStacktrace()
	}

	// event.Exception should be sorted such that the most recent error is last.
	reverse(e.Exception)
}

func (e *Event) setException(err error, maxErrorDepth int) {
[...]

PR #792, opened 2024-03-08
Add support for errors.Unwrap() in SetException

ribice:
Add support for handling errors using errors.Unwrap() in SetException.

Add ParentID, ExceptionID and Source to Exception.

Related issue: #693 

[RFC](https://github.com/getsentry/rfcs/blob/main/text/0079-exception-groups.md)


cleptric (dismissed):

interfaces.go
```diff
@@ -238,12 +239,15 @@ func (m *Mechanism) SetUnhandled() {
...
-	Mechanism  *Mechanism  `json:"mechanism,omitempty"`
+	Type             string      `json:"type,omitempty"`  // used as the main issue title
+	Value            string      `json:"value,omitempty"` // used as the main issue subtitle
+	Module           string      `json:"module,omitempty"`
+	ThreadID         uint64      `json:"thread_id,omitempty"`
+	IsExceptionGroup bool        `json:"is_exception_group,omitempty"`
+	ExceptionID      int         `json:"exception_id"`
+	ParentID         *int        `json:"parent_id,omitempty"`
+	Stacktrace       *Stacktrace `json:"stacktrace,omitempty"`
+	Mechanism        *Mechanism  `json:"mechanism,omitempty"`
```
cleptric, line 250: Adding these fields is a potential breaking change, just something we have to point out in the changelog.

interfaces.go
```diff
@@ -238,12 +239,15 @@ func (m *Mechanism) SetUnhandled() {
...
-	Value      string      `json:"value,omitempty"` // used as the main issue subtitle
-	Module     string      `json:"module,omitempty"`
-	ThreadID   uint64      `json:"thread_id,omitempty"`
-	Stacktrace *Stacktrace `json:"stacktrace,omitempty"`
-	Mechanism  *Mechanism  `json:"mechanism,omitempty"`
+	Type             string      `json:"type,omitempty"`  // used as the main issue title
+	Value            string      `json:"value,omitempty"` // used as the main issue subtitle
+	Module           string      `json:"module,omitempty"`
+	ThreadID         uint64      `json:"thread_id,omitempty"`
+	IsExceptionGroup bool        `json:"is_exception_group,omitempty"`
```
cleptric, line 246: We never set this. I think we should do so, if we unwrap more than one error.

interfaces.go
```diff
@@ -238,12 +239,15 @@ func (m *Mechanism) SetUnhandled() {
...
+	ExceptionID      int         `json:"exception_id"`
+	ParentID         *int        `json:"parent_id,omitempty"`
```
cleptric, line 248: According to the RFC, these fields need to go on the [`Mechansim`](https://github.com/getsentry/sentry-go/blob/master/interfaces.go#L223-L230).

client_test.go
```diff
@@ -169,9 +169,11 @@ func TestCaptureException(t *testing.T) {
...
 				},
 				{
-					Type:       "*errors.withStack",
-					Value:      "wat",
-					Stacktrace: &Stacktrace{Frames: []Frame{}},
+					Type:        "*errors.withStack",
+					Value:       "wat",
+					Stacktrace:  &Stacktrace{Frames: []Frame{}},
+					ExceptionID: 1,
+					ParentID:    pointerToInt(0),
```
cleptric, line 176:
```suggestion
					Type:        "*errors.withStack",
					Value:       "wat",
					Stacktrace:  &Stacktrace{Frames: []Frame{}},
                                        Mechanism:   {
                                            Type:        'generic',
                                            ExceptionID: 1,
					    ParentID:    pointerToInt(0),
                                        }
```

interfaces_test.go
```diff
@@ -215,6 +216,129 @@ func TestEventWithDebugMetaMarshalJSON(t *testing.T) {
...
+					ParentID:    pointerToInt(0),
+				},
+				{
+					Value:       "level 2: level 1: base error",
+					Type:        "*fmt.wrapError",
+					Stacktrace:  &Stacktrace{Frames: []Frame{}},
+					ExceptionID: 2,
+					ParentID:    pointerToInt(1),
+				},
+			},
```
cleptric, line 279: In this case, we should use an ExceptionGroup, so the Mechanism should include `IsExceptionGroup: true`.

util.go
```diff
@@ -112,3 +112,7 @@ func revisionFromBuildInfo(info *debug.BuildInfo) string {
 
 	return ""
 }
+
+func Pointer[T any](v T) *T {
+	return &v
+}
```
aldy505, line 118: What's this used for?

ribice: Seems I forgot to remove this. It was a helper function to create pointers.

interfaces.go
```diff
@@ -341,25 +343,40 @@ type Event struct {
...
+
+		causer, ok := err.(interface{ Cause() error })
+		if !ok {
+			// We cannot unwrap the error further.
+			break
 		}
+
+		// The error implements the Cause method, indicating it may have been wrapped
+		// using the github.com/pkg/errors package.
+		err = causer.Cause()
```
vaind, line 379: Is the grammatically correct word reserved or can we use this instead?
```suggestion
		cause, ok := err.(interface{ Cause() error })
		if !ok {
			// We cannot unwrap the error further.
			break
		}

		// The error implements the Cause method, indicating it may have been wrapped
		// using the github.com/pkg/errors package.
		err = cause.Cause()
```

vaind (approved):
LGTM after a changelog entry is added

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Add `http.request.method` attribute for performance span data ([#786](https://github.com/getsentry/sentry-go/pull/786))
 - Automatic transactions for Fasthttp integration ([#732](https://github.com/getsentry/sentry-go/pull/723))
 - Add `Fiber` integration ([#795](https://github.com/getsentry/sentry-go/pull/795))
+- Add support for errors.Unwrap() when unwrapping exceptions and new exception mechanism fields from [RFC-0079](https://github.com/getsentry/rfcs/blob/main/text/0079-exception-groups.md#new-mechanism-fields) ([#792](https://github.com/getsentry/sentry-go/pull/792))
```
vaind, line 12: "new exception mechanism fields" is an implementation detail IMO. How about simplifying this a little, maybe:
```suggestion
- Use `errors.Unwrap()` to create exception groups ([#792](https://github.com/getsentry/sentry-go/pull/792))
```

vaind (approved):

PR #273, opened 2020-08-23
Add Fiber integration

arsmn:
[Fiber](https://github.com/gofiber/fiber) integration

PR #795, opened 2024-03-18
Fiber integration

### LINKED ISSUE #441: Support for Fiber
Could you please make sure that sentry-go works with [Fiber](https://github.com/gofiber/fiber)

ribice:
[Fiber](https://github.com/gofiber/fiber) integration

These are just fixes and master merged on the #273. Since the fork was deleted, I was unable to push to it without some shenanigans thus opted for creating a new PR. 

Resolves #441.

Might also reach to Fiber as they provide the same integration at https://github.com/gofiber/contrib/tree/main/fibersentry.


fiber/README.md
```diff
@@ -0,0 +1,127 @@
+<p align="center">
+  <a href="https://sentry.io" target="_blank" align="center">
+    <img src="https://sentry-brand.storage.googleapis.com/sentry-logo-black.png" width="280">
+  </a>
+  <br />
+</p>
+
+# Official Sentry fiber Handler for Sentry-go SDK
+
+**Godoc:** https://godoc.org/github.com/getsentry/sentry-go/fiber
```
aldy505, line 10: We should change the URL to `https://pkg.go.dev/github.com/getsentry/sentry-go/fiber` instead

fiber/sentryfiber.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (h *handler) handle(ctx *fiber.Ctx) error {
+	hub := sentry.CurrentHub().Clone()
+	scope := hub.Scope()
+	scope.SetRequest(convert(ctx))
+	scope.SetRequestBody(ctx.Request().Body())
+	ctx.Locals(valuesKey, hub)
+	defer h.recoverWithSentry(hub, ctx)
+	return ctx.Next()
+}
```
aldy505, line 60: Should we support automatic HTTP server spans creation right now? Or do you plan to have it on a separate PR?

fiber/README.md
ribice: Yes, but all the URLs in the repository link to GoDoc. I'll create a separate PR to address that.

fiber/sentryfiber.go
ribice: I can address that in a separate PR (and add for all remaining integrations as well).

fiber/sentryfiber.go
```diff
@@ -0,0 +1,120 @@
...
+			panic(err)
+		}
+	}
+}
+
+func GetHubFromContext(ctx *fiber.Ctx) *sentry.Hub {
+	hub := ctx.Locals(valuesKey)
+	if hub, ok := hub.(*sentry.Hub); ok {
+		return hub
+	}
```
vaind, line 81: The first call result is not a hub so this may be a bit confusing. Can we do this instead? 

```suggestion
	if hub, ok := ctx.Locals(valuesKey).(*sentry.Hub); ok {
		return hub
	}
```

vaind (changes requested):
looks pretty good to me. Can you please add the missing tests for wait+timeout and cookies?  See https://app.codecov.io/gh/getsentry/sentry-go/pull/795?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=getsentry

fiber/sentryfiber_test.go
```diff
@@ -0,0 +1,247 @@
...
+			t.Fatalf("Request %q failed: %s", tt.Path, err)
+		}
+		if resp.StatusCode != http.StatusOK {
+			t.Errorf("Status code = %d", resp.StatusCode)
+		}
+	}
+
+	if ok := sentry.Flush(time.Second); !ok {
+		t.Fatal("sentry.Flush timed out")
+	}
```
vaind, line 222: this should be unnecessary with `WaitForDelivery: true`
```suggestion
```

vaind (approved):
LGTM except for the missing changelog entry

PR #800, opened 2024-03-26
Move exception group fields to mechanism

ribice:
While updating `develop` for [#1204](https://github.com/getsentry/develop/issues/1204), noticed that the new mechanism fields were at the wrong place.

[RFC 0079](https://github.com/getsentry/rfcs/blob/main/text/0079-exception-groups.md#new-mechanism-fields) specifies these fields (is_exception_group, parent_id, exception_id) to be in the mechanism.

This is how I initially implemented it, but for some reason (UI being broken?) I moved them to Data which is wrong according to RFC and other SDKs.

This is how it looks like on the UI.

CleanShot 2024-03-26 at 19 56 42@2x

#skip-changelog

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -165,10 +165,8 @@ func TestCaptureException(t *testing.T) {
 					Type:  "*sentry.customErr",
 					Value: "wat",
 					Mechanism: &Mechanism{
-						Data: map[string]any{
-							"is_exception_group": true,
-							"exception_id":       0,
-						},
+						ExceptionID:      0,
+						IsExceptionGroup: true,
 					},
 				},
 				{
@@ -176,11 +174,9 @@ func TestCaptureException(t *testing.T) {
 					Value:      "wat",
 					Stacktrace: &Stacktrace{Frames: []Frame{}},
 					Mechanism: &Mechanism{
-						Data: map[string]any{
-							"exception_id":       1,
-							"is_exception_group": true,
-							"parent_id":          0,
-						},
+						ExceptionID:      1,
+						ParentID:         Pointer(0),
+						IsExceptionGroup: true,
 					},
 				},
 			},
@@ -204,10 +200,8 @@ func TestCaptureException(t *testing.T) {
 					Type:  "*sentry.customErr",
 					Value: "wat",
 					Mechanism: &Mechanism{
-						Data: map[string]any{
-							"is_exception_group": true,
-							"exception_id":       0,
-						},
+						ExceptionID:      0,
+						IsExceptionGroup: true,
 					},
 				},
 				{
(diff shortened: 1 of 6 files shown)


ribice:
Shall we add `omitempty` to exception_id? If so, the 0 is not sent (unless it's a pointer), but otherwise it will always be included.

Also other SDK-s have special logic for IsExceptionGroup, while for Go the only logic is if it's wrapped - it's an exception group.

client_test.go
```diff
@@ -165,22 +165,18 @@ func TestCaptureException(t *testing.T) {
...
 					Value:      "wat",
 					Stacktrace: &Stacktrace{Frames: []Frame{}},
 					Mechanism: &Mechanism{
-						Data: map[string]any{
-							"exception_id":       1,
-							"is_exception_group": true,
-							"parent_id":          0,
-						},
+						ExceptionID:      1,
+						ParentID:         Pointer(0),
```
vaind, line 178: not a fan of the pointer type here. It's not referring to a memory address

vaind (commented):
I'm not familiar with the grouping mechanism so to clarify the discrepancy between RFC and implementations in other SDKs and the UI, I suggest you reach out to the developers that landed this in SDKs and maybe find a related PR in https://github.com/getsentry/sentry as well

client_test.go
ribice: The reason for using pointer is that for first exception, it expects `0` for exception_id and nothing for parent_id. The second exception is `{"exception_id":1,"parent_id":0}`. Since go defaults to 0 for zero int values, it wouldn't work for first exception.

It's either using pointers (simpler) or a nullable type with custom marshal function (something like sql.NullInt16).

ribice:
> I'm not familiar with the grouping mechanism so to clarify the discrepancy between RFC and implementations in other SDKs and the UI, I suggest you reach out to the developers that landed this in SDKs and maybe find a related PR in https://github.com/getsentry/sentry as well

They are definitely on Mechanism level, not data. This PR should fix that.

Example: https://github.com/getsentry/sentry/commit/f4fa54e861c3b6182121c1716799982217c555ed

cleptric:
The source of truth is always the protocol implemented in Relay.
In case of `Mechansim`, see https://github.com/getsentry/relay/blob/master/relay-event-schema/src/protocol/mechanism.rs#L170-L191.

Changes look good to me, besides this `Pointer()` thing.

ribice:
The reason for using the pointer type for `parent_id` (and that should be used for `exception_id` as well for similar reasons) is that we shouldn't send anything in case it's not set / first in the exception group, and for the second item we should send `0`.

So it's either using a pointer to achieve this (what I did) or using a nullable value (something like sql.NullInt16).

cleptric (approved):

PR #806, opened 2024-04-02
Fix setting of tracing variables

ribice:
Perhaps we should also do type checks?


vaind (commented):
Can you also add a test case that fails before the code changes are applied.

> Perhaps we should also do type checks?

yes, I agree. Maybe move the type cast to the condition.

cleptric (approved):

PR #807, opened 2024-04-02
Add transactions to fiber

ribice:


ribice:
@cleptric For failing tests, not sure where the `POS` method is coming from.

aldy505 (approved):

fiber/sentryfiber.go
```diff
@@ -51,11 +55,39 @@ func New(options Options) fiber.Handler {
...
+		fmt.Sprintf("%s %s", method, ctx.Path()),
+		options...,
+	)
+
+	defer func() {
+		transaction.Status = sentry.HTTPtoSpanStatus(ctx.Response().StatusCode())
+		transaction.Finish()
+	}()
+
+	transaction.SetData("http.request.method", method)
```
cleptric, line 84: Any chance we can also intercept the response and set `http.response.status_code`?
https://github.com/getsentry/sentry-go/issues/667

fiber/sentryfiber.go
```diff
@@ -51,11 +55,39 @@ func New(options Options) fiber.Handler {
...
+		sentry.WithOpName("http.server"),
+		sentry.ContinueFromRequest(convertedHTTPRequest),
+		sentry.WithTransactionSource(sentry.SourceRoute),
+	}
+
+	method := ctx.Method()
+
+	transaction := sentry.StartTransaction(
+		sentry.SetHubOnContext(ctx.Context(), hub),
+		fmt.Sprintf("%s %s", method, ctx.Path()),
```
cleptric, line 75: Could you verify that parameters are properly parameterized?
https://docs.gofiber.io/guide/routing/#parameters

So for e.g. `app.Get("/user/:id", func(c *fiber.Ctx)` with a request like `/user/1` we would want a transaction name of `GET /user/:id`.
If this is the case, we should add some tests. If not, and there is not always a ways to get a parameterized route, we need to change the transaction source to `SourceURL`.

ribice: They are not, I changed it. The way to get it in fiber is by using ctx.Route().Name, but that's only filled after .Next which wouldn't work in this case.

Issue #496, opened 2022-11-09
Update integrations & examples around performance monitoring


cleptric:
## Summary
With the addition of `StartTransaction`, we should update our integrations to support performance monitoring out of the box.

- [x] net/http
- [x] Echo
- [x] FastHTTP
- [x] Gin
- [x] Fiber
- [x] Iris
- [x] Negroni

We will also need to update or docs, I'll use this issue to track this work as well.

uded:
Any progress with Gin? We are using custom code now, not a great way...

aldy505:
Some updates:
- Echo: https://github.com/getsentry/sentry-go/pull/722
- Fasthttp: https://github.com/getsentry/sentry-go/pull/723
- Fiber: https://github.com/getsentry/sentry-go/pull/795

I'll do the rest next (Iris, Martini, Negroni), I haven't use them before but I'll give it a try.

cleptric:
You can skip Martini. Will remove it, no updates for 8 years. https://github.com/go-martini/martini

aldy505:
Great. Assign this to me, maybe? 😝

cleptric:
PRs are welcome.

PR #809, opened 2024-04-03
feat(iris): performance tracing

aldy505:
Addresses https://github.com/getsentry/sentry-go/issues/496

Tests will come at a later time (on this PR), probably next weekend.


aldy505:
cc @ribice @cleptric

iris/sentryiris.go
```diff
@@ -59,8 +62,31 @@ func (h *handler) handle(ctx iris.Context) {
...
+		options...,
+	)
+
+	defer func() {
+		hub.Scope().SetExtra("http.response.status_code", ctx.GetStatusCode())
+		transaction.Status = sentry.HTTPtoSpanStatus(ctx.GetStatusCode())
+		transaction.Finish()
+	}()
+
+	hub.Scope().SetExtra("http.request.method", ctx.Request().Method)
```
aldy505, line 85: Wait, why did I use `hub.Scope().SetExtra()` instead of `transaction.SetData()`, sigh...

iris/sentryiris_test.go
```diff
@@ -0,0 +1,479 @@
...
+					Method: "GET",
+					Headers: map[string]string{
+						"Accept-Encoding": "gzip",
+						"User-Agent":      "Go-http-client/1.1",
+					},
+				},
+				TransactionInfo: &sentry.TransactionInfo{Source: "route"},
+				Extra: map[string]interface{}{
+					"http.request.method":       string("GET"),
+					"http.response.status_code": int(200),
```
ribice, line 174: Nitpick, but what's with all these conversions to int? And for all other integrations I used constants from net/http, perhaps we could use them here as well (i.e. http.StatusOK for this one).

ribice: In fact you used them in your PR for Negroni, so should put them here too.

ribice (approved):

cleptric:
Will merge this later this week.

PR #824, opened 2024-05-03
add SendEventWithContext method

viglia:
This will let us pass a context.Context when creating a new request with getRequestFromEvent.

This will let us have a more granular control over how/when we cancel a request instead of relying only on the client timeout.

For example, if such API is invoked while serving another HTTP request, we might want to take advantage of such request context to bail out earlier than the client timeout if necessary.

As per [documentation](https://pkg.go.dev/net/http#Request.Context):

> For outgoing client requests, the context controls cancellation.
> 
> For incoming server requests, the context is canceled when the client's connection closes, the request is canceled (with HTTP/2), or when the ServeHTTP method returns.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -285,15 +285,9 @@ func getRequestFromEvent(ctx context.Context, event *Event, dsn *Dsn) (r *http.R
 	if err != nil {
 		return nil, err
 	}
-	if ctx != nil {
-		return http.NewRequestWithContext(
-			ctx,
-			http.MethodPost,
-			dsn.GetAPIURL().String(),
-			envelope,
-		)
-	}
-	return http.NewRequest(
+
+	return http.NewRequestWithContext(
+		ctx,
 		http.MethodPost,
 		dsn.GetAPIURL().String(),
 		envelope,
--- a/transport.go
+++ b/transport.go
@@ -258,7 +258,7 @@ func envelopeFromBody(event *Event, dsn *Dsn, sentAt time.Time, body json.RawMes
 	return &b, nil
 }
 
-func getRequestFromEvent(event *Event, dsn *Dsn, ctx context.Context) (r *http.Request, err error) {
+func getRequestFromEvent(ctx context.Context, event *Event, dsn *Dsn) (r *http.Request, err error) {
 	defer func() {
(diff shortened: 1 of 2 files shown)


transport.go
```diff
@@ -284,6 +285,14 @@ func getRequestFromEvent(event *Event, dsn *Dsn) (r *http.Request, err error) {
...
 	}
+	if ctx != nil {
+		return http.NewRequestWithContext(
+			ctx,
+			http.MethodPost,
+			dsn.GetAPIURL().String(),
+			envelope,
+		)
+	}
 	return http.NewRequest(
```
phacops, line 296: You don't need to leave this here, just set `ctx = context.Background()` if it's `nil` and call `NewRequestWithContext`.

phacops (changes requested):

transport.go
```diff
@@ -397,6 +404,11 @@ func (t *HTTPTransport) Configure(options ClientOptions) {
 
 // SendEvent assembles a new packet out of Event and sends it to remote server.
 func (t *HTTPTransport) SendEvent(event *Event) {
+	t.SendEventWithContext(context.Background(), event)
+}
+
+// SendEvent assembles a new packet out of Event and sends it to remote server.
+func (t *HTTPTransport) SendEventWithContext(ctx context.Context, event *Event) {
```
cleptric, line 411: Can we fix the grammar here please, even if it's just a copy of the old mistakes?

transport.go
```diff
@@ -627,6 +639,11 @@ func (t *HTTPSyncTransport) Configure(options ClientOptions) {
 
 // SendEvent assembles a new packet out of Event and sends it to remote server.
 func (t *HTTPSyncTransport) SendEvent(event *Event) {
+	t.SendEventWithContext(context.Background(), event)
+}
+
+// SendEvent assembles a new packet out of Event and sends it to remote server.
+func (t *HTTPSyncTransport) SendEventWithContext(ctx context.Context, event *Event) {
```
cleptric, line 646: Same here.

cleptric (commented):
Could you add a bit more about the why?

phacops (approved):

PR #833, opened 2024-05-15
Tracing without performance

ribice:
Continuation from #752


scope.go
```diff
@@ -379,6 +405,31 @@ func (scope *Scope) ApplyToEvent(event *Event, hint *EventHint) *Event {
 		}
 	}
 
+	// Apply the trace context to errors if there is a Span on the scope. If
+	// there isn't then fall back to the propagation context.
+	if event.Type != transactionType {
```
ribice, line 410: @cleptric There was a comment on old type to use this for transactions as well?

tracing.go
```diff
@@ -870,6 +870,22 @@ func WithSpanSampled(sampled Sampled) SpanOption {
 	}
 }
 
+func GetTraceHeader(s *Scope) string {
+	if s.span != nil {
+		return s.span.ToSentryTrace()
+	}
+
+	return fmt.Sprintf("%s-%s", s.propagationContext.TraceID, s.propagationContext.SpanID)
```
ribice, line 878: @cleptric This may be wrong?

cleptric: wdym? The lack of a sample flag at the end?

scope.go
cleptric: We have to throw this on all events, so check.ins, errors, transactios, etc.

tracing.go
ribice: Yes. Is that mandatory?

cleptric: Nope, omitting it signals the upstream sdk to make its own sampling decision. In case of TwP, this is the desired behavior.
https://develop.sentry.dev/sdk/performance/#header-sentry-trace

hub.go
```diff
@@ -365,6 +366,31 @@ func (hub *Hub) Flush(timeout time.Duration) bool {
 	return client.Flush(timeout)
 }
 
+// Continue a trace based on HTTP header values. If performance is enabled this
+// returns a SpanOption that can be used to start a transaction, otherwise nil.
+func (hub *Hub) ContinueTrace(trace, baggage string) (SpanOption, error) {
```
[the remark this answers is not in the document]
ribice: Yes, added that.

ribice:
The thing I feel wrong about is that I don't get the traces nested. Am I doing something wrong here?

CleanShot 2024-05-19 at 07 06 25@2x

aldy505:
> The thing I feel wrong about is that I don't get the traces nested. Am I doing something wrong here?
> 
> The parent span ID is set properly.
> 
> CleanShot 2024-05-19 at 07 06 25@2x

@ribice can I see the code to repro this one?

ribice:
> > The thing I feel wrong about is that I don't get the traces nested. Am I doing something wrong here?
> > The parent span ID is set properly.
> > CleanShot 2024-05-19 at 07 06 25@2x
> 
> @ribice can I see the code to repro this one?

https://gist.github.com/ribice/2c583f9c1edc3084d3572e2ea5065419

aldy505:
> > > The thing I feel wrong about is that I don't get the traces nested. Am I doing something wrong here?
> > > The parent span ID is set properly.
> > > CleanShot 2024-05-19 at 07 06 25@2x
> > 
> > 
> > @ribice can I see the code to repro this one?
> 
> https://gist.github.com/ribice/2c583f9c1edc3084d3572e2ea5065419

I think your `sentry.ContinueTrace` is conflicting with the ones from here on the sentryhttp middleware

https://github.com/getsentry/sentry-go/blob/9f80641c4ca2b23ee254170e164556cf91e4ef07/http/sentryhttp.go#L113-L122

The `sentry.ContinueFromRequest` should acquire the Sentry-related headers and append it to the span options. Try to remove that part from the middleware and see if everything works, maybe?

aldy505:
Got time to finally come back to this, made a repro on Nodejs, nothing wrong about it not being nested. But I wonder why it got the same parent span id though. It should only got 2-3 captured events

image

Repro https://gist.github.com/aldy505/90354abc6877071dec601c8ddabca092

ribice:
Hmm, seems you're right. Capturing it first time works properly (same as in node snippet above), but every subsequent request appends to that trace which is weird. I'll investigate now.

First:

CleanShot 2024-05-26 at 18 07 52

Second:

CleanShot 2024-05-26 at 18 09 27

ribice:
I was able to resolve it by creating a new Scope/Hub in every request. This way it generates new traceID/spanID.

https://gist.github.com/ribice/2c583f9c1edc3084d3572e2ea5065419 (service1-updated.go).

Not sure if this is the intended way.

aldy505:
On the JS SDK, the `ContinueTrace` is a function that accepts a callback. I think on the callback, it creates a new scope internally, Not sure if this should also be done on Go though.

cleptric (approved):

hub.go
```diff
@@ -365,6 +365,34 @@ func (hub *Hub) Flush(timeout time.Duration) bool {
 	return client.Flush(timeout)
 }
 
+// Continue a trace based on HTTP header values. If performance is enabled this
+// returns a SpanOption that can be used to start a transaction, otherwise nil.
+// Uses the Global Hub.
+func ContinueTrace(trace, baggage string) (SpanOption, error) {
```
cleptric, line 371: We've seen a lot of issues in the past with methods operating on the global hub, hence I would not add this.

cleptric (approved):

PR #834, opened 2024-05-16
feat(metrics): add metrics support

viglia:


cleptric:
As we're fundamentally shifting how Metrics will work at Sentry, we'll put this PR on hold for now. At least until we know how the next steps look like.

PR #837, opened 2024-05-28
Implement HTTP Hijacker interface

### LINKED ISSUE #836: regression: after upgrade to 0.28 my websocket server no longer works 🐛
## Summary

I have a service service mixed REST and Websocket requests over a single http.Server.

The sentry-go is setup the following way:

```go
        srv.Server = &http.Server{
                Handler: sentryhttp.New(sentryhttp.Options{
                        Repanic:         true,
                        WaitForDelivery: false,
                        Timeout:         30 * time.Second,
                }).Handle(m),
                // ...
```

Websocket endpoint is using [nhooyr/websocket](https://github.com/nhooyr/websocket) to accept websocket connection, and it is pretty straightforward:

```go
func (s *Server) createSubscription(w http.ResponseWriter, r *http.Request) {
        // ...

        conn, err := websocket.Accept(w, r, &websocket.AcceptOptions{
                CompressionMode: websocket.CompressionContextTakeover,
        })
        if err != nil {
                shared.Errorf(w, r, http.StatusInternalServerError, "failed to accept websocket connection: %w", err)
                return
        }
```

After upgrading from 0.27.0 to 0.28.0 the above endpoint no longer works, every request fails with:

```
failed to accept websocket connection: failed to accept WebSocket connection: http.ResponseWriter does not implement http.Hijacker
```

The reason is that go-sentry started overriding `http.ResponseWriter` which is now of the following type:

```
struct { httpsnoop.Unwrapper; http.ResponseWriter }{Unwrapper:(*httpsnoop.rw)( 0xc00039e6e0), ResponseWriter:(*httpsnoop.rw)( 0xc00039e6e0)}
```

Which clearly does not implement http.Hijacker interface.

## Steps To Reproduce

1. have a working websocket server
2. upgrade sentry-go to 0.28.0
3. ???
4. no profit

## Expected Behavior

4. profit

## Screenshots

n/a

## Sentry.io Event

n/a

## Environment

important

### SDK
 - `sentry-go` version: 0.28.0
 - Go version: go1.22.3
 - Using Go Modules? yes

### Sentry

 - Anything particular to your environment that could be related to this issue? n/a

## Additional context

n/a

ribice:
Resolves #836 

Was able to reproduce it with 0.28.0 (`Error accepting websocket connection:  failed to accept WebSocket connection: http.ResponseWriter does not implement http.Hijacker`) and see it resolved with this fix.


ribice:
The only difference between this (and previous incomplete Wrapper) and Chi's is that this defaults to 200 instead of 0 if WriteHeader has not yet been called.

http/wrap_writer.go
```diff
@@ -0,0 +1,178 @@
+package sentryhttp
+
+import (
+	"bufio"
+	"io"
+	"net"
+	"net/http"
+)
+
+// This wrapper is derived from https://github.com/go-chi/chi/blob/master/middleware/wrap_writer.go
```
cleptric, line 10: Please include the entire content of their license when copying/deriving code.
https://github.com/go-chi/chi/blob/master/LICENSE

cleptric (approved):

kydemy-fran:
Is this now working transparently? or requires some extra initialization?
I am still getting the error using the negronisentry package.

ribice:
I just checked and it seems you're right, I've updated the implementation for stdlib but not for negroni. I'll submit a fix today.

PR #847, opened 2024-06-28
fix: input validation to NewRequest

athyk:
Fixes `panic: runtime error: invalid memory address or nil pointer dereference` when passing `http.Request` that has not defined `r.Host` or `r.URL.Path` or even just `nil` in general.

```go
req := &http.Request{}
sentry.NewRequest(req)
```

- This adds an error signature onto [`NewRequest`](https://github.com/athyk/sentry-go/blob/fff8a3c49815e31a65c13a0c6ca4e37e0267da0d/interfaces.go#L179) and [`entryToEvent`](https://github.com/athyk/sentry-go/blob/fff8a3c49815e31a65c13a0c6ca4e37e0267da0d/logrus/logrusentry.go#L151)
- Tests have been added to cover error handling

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -176,7 +176,11 @@ var sensitiveHeaders = map[string]struct{}{
 //
 // NewRequest avoids operations that depend on network access. In particular, it
 // does not read r.Body.
-func NewRequest(r *http.Request) *Request {
+func NewRequest(r *http.Request) (*Request, error) {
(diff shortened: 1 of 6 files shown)


ribice:
This is a breaking change, as it changes the signature of NewRequest. I assume that's used externally (as it's public/exposed). The added validation don't seem worthy of a breaking change.

PR #852, opened 2024-07-03
Add ability to skip frames

### LINKED ISSUE #741: Stack trace has confusing top frame when using sentry.CurrentHub().Recover(err)
## Summary

When using `sentry.CurrentHub().Recover(err)` to report an exception the stack trace contains that stack frame at the top.

## Steps To Reproduce

Create recover block:

```go
		defer func() {
			if err := recover(); err != nil {
				log.Printf("[ERR] Error: %v", err)
				sentry.CurrentHub().Recover(err)
			}
		}()
```

Have panic after that in a nested function.

## Expected Behavior

I expect the top stack frame to be the frame that caused the panic. Instead, it is the stack frame that has the recover. This makes some logical sense, because it is the thing that is calling `sentry.CurrentHub().Recover(err)` but it is unintuitive.

### SDK
 - `sentry-go` version: 0.25.0
 - Go version: go1.20.10
 - Using Go Modules? yes

### Sentry

ribice:
I don't find this solution elegant in particular, but it adds ability to skipFrames in StackTrace (which resolves #741 for example).

Is EventFromMessage invoked directly by external users? If so, this would be a breaking change and instead a variadic options/parameter could be added.

I'll add tests once we agree on the implementation.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Test
         run: make test-coverage
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@125fc84a9a348dbcf27191600683ec096ec9021c # pin@v4.4.1
+        uses: codecov/codecov-action@e28ff129e5465c2c0dcc6f003fc735cb6ae0c673 # pin@v4.5.0
         with:
           directory: .coverage
           token: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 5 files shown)


cleptric:
I fear this is something we would need to expose on the top level methods too :/
So `CaptureMessage`, `CaptureException` and most likely `Recover` too, to be truly useful.
An variadic options argument might be the better way forward.

ribice:
> I fear this is something we would need to expose on the top level methods too :/ So `CaptureMessage`, `CaptureException` and most likely `Recover` too, to be truly useful. An variadic options argument might be the better way forward.

I've added it to CaptureMessage, CaptureEvent, Recover/RecoverWithContext.

cleptric (approved):

PR #857, opened 2024-07-10
Add Zerolog integration

### LINKED ISSUE #853: Add support for zerolog
## Summary

Support for [zerolog](https://github.com/rs/zerolog) would be great. It has similar Hook approach to logrus which is already supported.

## Motivation

It's a widely used logging package and would help a lot with tools that are using it.

ribice:
Resolves #853


ribice:
Instead of implementing the zerolog.Writer, alternative would be to use Hooks (similar to logrus implementation).

ribice:
For the ./otel `go mod tidy` issue, I locally cd into otel and do `go mod tidy` - nothing happens.

antisocrates:
Will this be available in the next minor release?

ribice:
> Will this be available in the next minor release?

Yes, we plan to include this in next release.

cleptric (dismissed):
Let's make this it's own package. You can update https://github.com/getsentry/sentry-go/blob/master/.craft.yml and I'll add it to our release registry.

cleptric (commented):
Meant to comment

Makefile
```diff
@@ -79,4 +79,4 @@ lint: ## Lint (using "golangci-lint")
 
 fmt: ## Format all Go files
 	gofmt -l -w -s .
-.PHONY: fmt
+.PHONY: fmt
```
cleptric, line 82:
```suggestion
.PHONY: fmt

```

cleptric (approved):

PR #859, opened 2024-07-18
feat(request): Add Proxy-Authorization as a sensitive header

jonathroth:

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -166,10 +166,11 @@ type Request struct {
 }
 
 var sensitiveHeaders = map[string]struct{}{
-	"Authorization":   {},
-	"Cookie":          {},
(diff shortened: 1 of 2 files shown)


cleptric (approved):
Thanks! Did these values show up in the Sentry UI, or did server-side scrubbing redact them?

jonathroth:
> Thanks! Did these values show up in the Sentry UI, or did server-side scrubbing redact them?

I was debugging the SDK locally and noticed this value would be passed to Sentry. I haven't checked anything against the UI or server side yet, but the SDK shouldn't send this sensitive value to the server when PII is disabled

cleptric:
Yes, I just wanted to also make sure our scrubbers work correctly.

PR #862, opened 2024-07-22
Tracing without performance improvements

cleptric:


tracing.go
```diff
@@ -324,17 +324,22 @@ func (s *Span) ToSentryTrace() string {
...
+	if t == nil {
+		return ""
+	}
 
-		return containingTransaction.dynamicSamplingContext.String()
+	// In case there is currently no frozen DynamicSamplingContext attached to the transaction,
+	// create one from the properties of the transaction.
+	if !s.dynamicSamplingContext.IsFrozen() {
+		// This will return a frozen DynamicSamplingContext.
+		if dsc := DynamicSamplingContextFromTransaction(t); dsc.HasEntries() {
```
ribice, line 336: @cleptric Not sure if this will cause unwanted changes, but it did fix the failing test because it was setting the dynamicSamplingContext to "" due to nil client.

ribice:
I fixed the gin panic issue. But with these changes, the request body is not properly set on the event. I'll keep debugging this.

ribice:
> I fixed the gin panic issue. But with these changes, in `gin`, the request body is not properly set on the event. I'll keep debugging this.

Resolved.

hub.go
```diff
@@ -366,28 +366,6 @@ func (hub *Hub) Flush(timeout time.Duration) bool {
 	return client.Flush(timeout)
 }
 
-// Continue a trace based on HTTP header values. If performance is enabled this
-// returns a SpanOption that can be used to start a transaction, otherwise nil.
-// TODO: remove - moved to tracing.go.
-func (hub *Hub) ContinueTrace(trace, baggage string) (SpanOption, error) {
```
ribice, line 370: Should we call this change in changelog?

cleptric: This was never released, was part of https://github.com/getsentry/sentry-go/pull/833

ribice (approved):

PR #865, opened 2024-07-28
Add Slog integration

### LINKED ISSUE #527: Create an exp/slog.Handler
The Go team have released [an experimental structured logger](https://golang.org/x/exp/slog). It is very likely to become a part of the standard library in the future. As a Sentry user, I would like any errors that I log with slog.Error to be sent to Sentry.

The simplest way for this to happen that I can think of is for Sentry to release some kind of slog.Handler that intercepts certain log calls and sends some Sentry before writing them to an underlying wrapped slog.Handler for output to stderr, etc.

ribice:
Resolves #527


ribice:
@cleptric Just realised we can't use 1.18 with exp/slog (requires at least 1.19): https://github.com/golang/go/issues/57113

assembly-winston:
Will this be merged?

cleptric:
Yes

cleptric (approved):
Let's make this it's own package. You can update https://github.com/getsentry/sentry-go/blob/master/.craft.yml and I'll add it to our release registry.

ribice:
> Let's make this it's own package. You can update https://github.com/getsentry/sentry-go/blob/master/.craft.yml and I'll add it to our release registry.

I updated .craft.yml and made `slog` integration a module. Is there anything else that needs to be updated?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### Features
+
+- Add `sentryzerolog` integration ([#865](https://github.com/getsentry/sentry-go/pull/865))
```
cleptric, line 5:
```suggestion
- Add `sentryslog` integration ([#865](https://github.com/getsentry/sentry-go/pull/865))
```

cleptric:
Just CI needs to pass and then we can get this merged 🙂

ribice:
Shall I update the `vet` script in Makefile to skip if no go files are found (which is the case for slog in Go <1.21 due to build constraints)?

ribice:
Waiting for #898 to be merged before I merge this.

slog/common.go
```diff
@@ -0,0 +1,260 @@
...
+
+	for _, attr := range attrs {
+		k, v := attr.Key, attr.Value
+		output[k] = valueToString(v)
+	}
+
+	return output
+}
+
+func valueToString(v slog.Value) string {
```
pierrre, line 114: I'm just wondering: why not use https://pkg.go.dev/log/slog#Value.String ?

PR #866, opened 2024-07-28
Add logger name to logrus integration

ribice:


cleptric:
Any reason for this change?

ribice:
> Any reason for this change?

I assumed this `Logger` field is used for this purpose (saw it in other integrations).

cleptric (approved):

Issue #868, opened 2024-08-05
fiber dependency has vulnerability


shivamsouravjha:
## Summary

* Currently `github.com/gofiber/fiber/v2 v2.52.2` is being used in Sentry because of this a critical vulnerability is introduced.
* if [RHDA](https://marketplace.visualstudio.com/items?itemName=redhat.fabric8-analytics) is checked this would give a critical score of 10/10.
![Image](https://github.com/user-attachments/assets/f9cd90a9-bd0a-4307-a52d-5fd129b15e14)
* This led to the check around the fibre releases and in the latest release they themselves have mentioned solving [Vulnerability](https://github.com/gofiber/fiber/releases/tag/v2.52.5)

* Would be great if this could be solved with a minor version bump as this tool is used by many OSS projects.

shivamsouravjha:
I've added a minor fix around this and have linked the PR here https://github.com/getsentry/sentry-go/pull/869

Best Regards.

cleptric:
https://github.com/getsentry/sentry-go/pull/869#issuecomment-2269636064

PR #869, opened 2024-08-05
update: fiber version

shivamsouravjha:
Bumped the version of fibre as this is a minor version bump nothing seems to be refactored.
1. Solves https://github.com/getsentry/sentry-go/issues/868


cleptric:
> Integration dependencies: If a security issue is discovered in one of our integration dependencies (for example, gin, echo, or negroni), it is the responsibility of the end user to make sure that those modules are updated in a timely manner in their applications.

https://github.com/getsentry/sentry-go/security/policy

PR #871, opened 2024-08-12
Implement hijacker interface for Negroni integration

ribice:
Resolves https://github.com/getsentry/sentry-go/pull/837#issuecomment-2283914037


cleptric (approved):
There are some breaking changes in here we need to call out on the changelog. Besides this, LGTM!

ribice:
> There are some breaking changes in here we need to call out on the changelog. Besides this, LGTM!

Are you referring to moving the WrapResponseWriter from `sentryhttp` to `sentry`? 

That wasn't supposed to be exposed externally, so doubt anyone used it.  To prevent a breaking change, we can keep it at sentryhttp as well (and just call the one at sentry internally)?

cleptric:
It’s ok, we just need to call it out 🙂

ribice:
@cleptric Perhaps we should move the Hijacker interface to `/internal`?

aldy505:
> Perhaps we should move the Hijacker interface to `/internal`?

@ribice I'd say you should move it to `internal` package, but keep the one at `sentryhttp` to avoid breaking change, then add `DEPRECATION` notice (the one that's [according to godoc format](https://go.dev/blog/godoc)). 

The one on `sentryhttp` should be a passthrough to `internal` package. Then after 2-3 minor version release, you can safely remove it.

cleptric (approved):

kydemy-fran:
Hi there,
I reported the originar issue when negroni, the response writter does not implement `http.Hijacker` interface.
Today we have upgraded from `v0.27.0` to `v0.29.0` and the error persists.

I've reviewed the code and I've found the issue:
https://github.com/getsentry/sentry-go/blob/a6acd05f8d419d0445475b9018bd59fc0503b322/internal/httputils/wrap_writer.go#L35-L57

The w is of type `*negroni.responseWriter` in negroni v1, which is the one referenced by the `sentry-go` package.
This type implements `http.Flusher`, `http.Hijacker`, but not `io.ReaderFrom`.
So `fl` and `hj` are true, but `rf` is not. So in line 48 does not return the `httpFancyWriter`.
And it returns the basic `flushWriter` in line 53. (which does not support hijacker / websockets.

We have had to rollback our releases today as this was failing.

As mentioned before, negroni v1 uses a limited writer. This is solved in negroni v3, which is easy to migrate to, but it would be nice if go-sentry was upgraded to v3 too, as v1 was released in 2018. Having our project requiring both versions is a nightmare.

From my point of view, if `go-sentry` imports negroni v1, then the library should work with v1 (and this is not happening at the moment). If you expect people to be using v3 (which works with go-sentry v0.29.0) then you should import negroni v3 in your `go.mod`.

Please let me know your thoughts

Thank you!

cleptric:
@ribice can you take a look please?

ribice:
Yes I can. Before I change the interface to support v1, I'm curious if we can update to v3? Is that an option?

cleptric:
Fine with me.

kydemy-fran:
Thank you for looking into this guys :)

ribice:
Upgraded negroni to latest version in #885. This worked just fine: https://gist.github.com/ribice/95934b3c209b9692efc4b50b91d5dd4a

ribice:
@kydemy-fran Can you confirm that it works as expected with 0.29.1?

kydemy-fran:
Hi there @ribice and @cleptric.
I've just tested it and it seems to be working fine using the negroni v3 implementation. 
Thank you for your help!!

PR #872, opened 2024-08-20
ci: add go1.23

mjq:
We advertise support for the latest 3 Go versions in the README, but this puts us at testing the latest 6 - maybe we can consider trimming some of the older versions.


ribice (approved):

ribice:
> We advertise support for the latest 3 Go versions in the README, but this puts us at testing the latest 6 - maybe we can consider trimming some of the older versions.

I discussed this with @cleptric earlier, and the plan is to keep support for older versions for now.

PR #876, opened 2024-08-27
feat: http client integration

aldy505:
An effort to implement this: https://develop.sentry.dev/sdk/telemetry/traces/modules/requests/

Since we already have tracing without performance, this should be good to go.

### DIFF
--- a/httpclient/sentryhttpclient.go
+++ b/httpclient/sentryhttpclient.go
@@ -1,0 +1,106 @@
+// Package sentryhttpclient provides Sentry integration for Requests modules to enable distributed tracing between services.
+// It is compatible with `net/http.RoundTripper`.
+//
+//	import sentryhttpclient "github.com/getsentry/sentry-go/httpclient"
+//
+//	roundTrippper := sentryhttpclient.NewSentryRoundTripper(nil, nil)
+//	client := &http.Client{
+//		Transport: roundTripper,
+//	}
+//
+//	request, err := client.Do(request)
+package sentryhttpclient
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/getsentry/sentry-go"
+)
+
+// SentryRoundTripTracerOption provides a specific type in which defines the option for SentryRoundTripper.
+type SentryRoundTripTracerOption func(*SentryRoundTripper)
+
+// WithTags allows the RoundTripper to includes additional tags.
+func WithTags(tags map[string]string) SentryRoundTripTracerOption {
+	return func(t *SentryRoundTripper) {
+		for k, v := range tags {
+			t.tags[k] = v
+		}
+	}
+}
+
+// WithTag allows the RoundTripper to includes additional tag.
+func WithTag(key, value string) SentryRoundTripTracerOption {
+	return func(t *SentryRoundTripper) {
+		t.tags[key] = value
+	}
+}
+
+// NewSentryRoundTripper provides a wrapper to existing http.RoundTripper to have required span data and trace headers for outgoing HTTP requests.
+//
+//   - If `nil` is passed to `originalRoundTripper`, it will use http.DefaultTransport instead.
+func NewSentryRoundTripper(originalRoundTripper http.RoundTripper, opts ...SentryRoundTripTracerOption) http.RoundTripper {
+	if originalRoundTripper == nil {
+		originalRoundTripper = http.DefaultTransport
+	}
+
+	t := &SentryRoundTripper{
+		originalRoundTripper: originalRoundTripper,
+		tags:                 make(map[string]string),
+	}
+
+	for _, opt := range opts {
+		if opt != nil {
+			opt(t)
+		}
+	}
+
+	return t
+}
+
+// SentryRoundTripper provides a http.RoundTripper implementation for Sentry Requests module.
+type SentryRoundTripper struct {
+	originalRoundTripper http.RoundTripper
+
+	tags map[string]string
+}
+
+func (s *SentryRoundTripper) RoundTrip(request *http.Request) (*http.Response, error) {
+	ctx := request.Context()
+	cleanRequestURL := request.URL.Redacted()
+
+	span := sentry.StartSpan(ctx, "http.client", sentry.WithTransactionName(fmt.Sprintf("%s %s", request.Method, cleanRequestURL)))
+
+	for k, v := range s.tags {
+		span.SetTag(k, v)
+	}
+
+	defer span.Finish()
+
+	span.SetData("http.query", request.URL.Query().Encode())
+	span.SetData("http.fragment", request.URL.Fragment)
+	span.SetData("http.request.method", request.Method)
+	span.SetData("server.address", request.URL.Hostname())
+	span.SetData("server.port", request.URL.Port())
+
+	// Always add `Baggage` and `Sentry-Trace` headers.
+	request.Header.Add("Baggage", span.ToBaggage())
+	request.Header.Add("Sentry-Trace", span.ToSentryTrace())
+
+	response, err := s.originalRoundTripper.RoundTrip(request)
+
+	if response != nil {
+		span.Status = sentry.HTTPtoSpanStatus(response.StatusCode)
+		span.SetData("http.response.status_code", response.StatusCode)
+		span.SetData("http.response_content_length", response.ContentLength)
+	}
+
+	return response, err
+}
+
+// SentryHttpClient provides a default HTTP client with SentryRoundTripper included.
+// This can be used directly to perform HTTP request.
+var SentryHttpClient = &http.Client{
+	Transport: NewSentryRoundTripper(http.DefaultTransport),
+}
--- a/httpclient/sentryhttpclient_test.go
+++ b/httpclient/sentryhttpclient_test.go
@@ -1,0 +1,244 @@
+package sentryhttpclient_test
+
+import (
+	"bytes"
+	"context"
+	"crypto/rand"
+	"crypto/tls"
+	"io"
+	"net/http"
+	"strconv"
+	"testing"
+
+	"github.com/getsentry/sentry-go"
+	sentryhttpclient "github.com/getsentry/sentry-go/httpclient"
+	"github.com/getsentry/sentry-go/internal/testutils"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+)
+
+type noopRoundTripper struct {
+	ExpectResponseStatus int
+	ExpectResponseLength int
+}
+
+func (n *noopRoundTripper) RoundTrip(request *http.Request) (*http.Response, error) {
+	responseBody := make([]byte, n.ExpectResponseLength)
+	rand.Read(responseBody)
+	return &http.Response{
+		Status:     "",
+		StatusCode: n.ExpectResponseStatus,
+		Proto:      "HTTP/1.1",
+		ProtoMajor: 1,
+		ProtoMinor: 1,
+		Header: map[string][]string{
+			"Content-Length": {strconv.Itoa(len(responseBody))},
+		},
+		Body:             io.NopCloser(bytes.NewReader(responseBody)),
+		ContentLength:    int64(len(responseBody)),
+		TransferEncoding: []string{},
+		Close:            false,
+		Uncompressed:     false,
+		Trailer:          map[string][]string{},
+		Request:          request,
+		TLS:              &tls.ConnectionState{},
+	}, nil
+}
+
+func TestIntegration(t *testing.T) {
+	tests := []struct {
+		RequestMethod      string
+		RequestURL         string
+		TracerOptions      []sentryhttpclient.SentryRoundTripTracerOption
+		WantStatus         int
+		WantResponseLength int
+		WantTransaction    *sentry.Event
+	}{
+		{
+			RequestMethod:      "GET",
+			RequestURL:         "https://example.com/foo",
+			WantStatus:         200,
+			WantResponseLength: 0,
+			WantTransaction: &sentry.Event{
+				Extra: map[string]interface{}{
+					"http.fragment":                string(""),
+					"http.query":                   string(""),
+					"http.request.method":          string("GET"),
+					"http.response.status_code":    int(200),
+					"http.response_content_length": int64(0),
+					"server.address":               string("example.com"),
+					"server.port":                  string(""),
+				},
+				Level:           sentry.LevelInfo,
+				Transaction:     "GET https://example.com/foo",
+				Type:            "transaction",
+				TransactionInfo: &sentry.TransactionInfo{Source: "custom"},
+			},
+		},
+		{
+			RequestMethod:      "GET",
+			RequestURL:         "https://example.com:443/foo/bar?baz=123#readme",
+			TracerOptions:      []sentryhttpclient.SentryRoundTripTracerOption{nil, nil, nil},
+			WantStatus:         200,
+			WantResponseLength: 0,
+			WantTransaction: &sentry.Event{
+				Extra: map[string]interface{}{
+					"http.fragment":                string("readme"),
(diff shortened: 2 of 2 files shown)


ribice:
@aldy505 We should document this (in _examples, and perhaps sentry-docs?), and also update changelog accordingly.

httpclient/sentryhttpclient.go
```diff
@@ -0,0 +1,106 @@
...
+	tags map[string]string
+}
+
+func (s *SentryRoundTripper) RoundTrip(request *http.Request) (*http.Response, error) {
+	ctx := request.Context()
+	cleanRequestURL := request.URL.Redacted()
+
+	span := sentry.StartSpan(ctx, "http.client", sentry.WithTransactionName(fmt.Sprintf("%s %s", request.Method, cleanRequestURL)))
+
+	for k, v := range s.tags {
```
ribice, line 75: Could we just set the span.Tags instead of ranging through these, and setting one by one?

This is the 'right' way to do it, but it involves checking for nil map and dealing with mutex.

aldy505: Lol good catch. I was copying directly from this PR https://github.com/aldy505/sentry-integration/pull/1

httpclient/sentryhttpclient.go
```diff
@@ -0,0 +1,106 @@
...
+
+	span.SetData("http.query", request.URL.Query().Encode())
+	span.SetData("http.fragment", request.URL.Fragment)
+	span.SetData("http.request.method", request.Method)
+	span.SetData("server.address", request.URL.Hostname())
+	span.SetData("server.port", request.URL.Port())
+
+	// Always add `Baggage` and `Sentry-Trace` headers.
+	request.Header.Add("Baggage", span.ToBaggage())
+	request.Header.Add("Sentry-Trace", span.ToSentryTrace())
```
cleptric, line 89: This won't work anymore with the recent tracing without performance changes.

httpclient/sentryhttpclient.go
```diff
@@ -0,0 +1,102 @@
...
+	tags map[string]string
+}
+
+func (s *SentryRoundTripper) RoundTrip(request *http.Request) (*http.Response, error) {
+	ctx := request.Context()
+	cleanRequestURL := request.URL.Redacted()
+
+	span := sentry.StartSpan(ctx, "http.client", sentry.WithTransactionName(fmt.Sprintf("%s %s", request.Method, cleanRequestURL)))
+	span.Tags = s.tags
+	defer span.Finish()
```
cleptric, line 75: We do not want to always start transactions here. If one already exists, start a child span.

cleptric (changes requested):

httpclient/sentryhttpclient.go
aldy505: Just to be clear: http.client spans should only created if there is any parent span?

cleptric: Yes, Sentry does not work well with these kind of single span transactions. We can revisit this once span streaming is fully supported in the product, but this will take a few months.

cleptric: https://sentry.io/insights/http/ should also work with these spans fwiw, docs are [here](https://develop.sentry.dev/sdk/telemetry/traces/modules/requests/).

cleptric: *which you already mentioned in the PR description, nvm

aldy505: Another question (unrelated): Does that also affects all the span op within the insights module? Meaning not only `http.client`, but `db.sql` and `messaging.*` also counts?

cleptric: I guess technically we extract all the required fields from a transaction and store it into the span data set, though this is not how these modules where initially designed to work.

httpclient/sentryhttpclient.go
```diff
@@ -0,0 +1,107 @@
...
+
+	tags map[string]string
+}
+
+func (s *SentryRoundTripper) RoundTrip(request *http.Request) (*http.Response, error) {
+	// Only create the `http.client` span only if there is a parent span.
+	parentSpan := sentry.GetSpanFromContext(request.Context())
+	if parentSpan == nil {
+		return s.originalRoundTripper.RoundTrip(request)
+	}
```
aldy505, line 74: A quick note why I prefer this instead of creating a `sentry.WithOnlyIfParentExists()` (a `SpanOption` type):

I used the JS SDK a lot at work lately, and I've been wondering the entire week, on their `startSpan` function, they have this as an option: https://github.com/getsentry/sentry-javascript/blob/216aaeba1ee27cce8a4876e1f9212ba374eb30b3/packages/types/src/startSpanOptions.ts#L14-L15

Should the Go SDK move forward with that, as a `SpanOption` or not?

We could to the `SpanOption` thing, but it feels really weird since the way Go SDK handles scopes/hubs with span is to add the span pointer into the scope. See https://github.com/getsentry/sentry-go/blob/a6acd05f8d419d0445475b9018bd59fc0503b322/tracing.go#L196-L199

Although we could just not call the `scope.SetSpan()` function, I don't know about the behavior if this span that we're creating will ever have a child span being created manually by the user.

httpclient/sentryhttpclient.go
```diff
@@ -0,0 +1,107 @@
...
+
+	span.SetData("http.query", request.URL.Query().Encode())
+	span.SetData("http.fragment", request.URL.Fragment)
+	span.SetData("http.request.method", request.Method)
+	span.SetData("server.address", request.URL.Hostname())
+	span.SetData("server.port", request.URL.Port())
+
+	// Always add `Baggage` and `Sentry-Trace` headers.
+	request.Header.Add("Baggage", span.ToBaggage())
+	request.Header.Add("Sentry-Trace", span.ToSentryTrace())
```
cleptric, line 90: To support the recently added "Tracing without Performance" feature, we should use `hub.GetTraceparent()` and `hub.GetBaggage()`

httpclient/sentryhttpclient.go
```diff
@@ -0,0 +1,107 @@
...
+	// Only create the `http.client` span only if there is a parent span.
+	parentSpan := sentry.SpanFromContext(request.Context())
+	if parentSpan == nil {
+		return s.originalRoundTripper.RoundTrip(request)
+	}
+
+	cleanRequestURL := request.URL.Redacted()
+
+	span := parentSpan.StartChild("http.client", sentry.WithTransactionName(fmt.Sprintf("%s %s", request.Method, cleanRequestURL)))
+	span.Tags = s.tags
```
cleptric, line 79: Why are we setting span tags here?

httpclient/sentryhttpclient.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// SentryRoundTripper provides a http.RoundTripper implementation for Sentry Requests module.
+type SentryRoundTripper struct {
+	originalRoundTripper http.RoundTripper
+
+	tags map[string]string
+}
+
+func (s *SentryRoundTripper) RoundTrip(request *http.Request) (*http.Response, error) {
```
cleptric, line 69: We need to add a new `ClientOption` for https://develop.sentry.dev/sdk/telemetry/traces/#tracepropagationtargets

httpclient/sentryhttpclient.go
```diff
@@ -0,0 +1,107 @@
...
+	span.SetData("http.fragment", request.URL.Fragment)
+	span.SetData("http.request.method", request.Method)
+	span.SetData("server.address", request.URL.Hostname())
+	span.SetData("server.port", request.URL.Port())
+
+	// Always add `Baggage` and `Sentry-Trace` headers.
+	request.Header.Add("Baggage", span.ToBaggage())
+	request.Header.Add("Sentry-Trace", span.ToSentryTrace())
+
+	response, err := s.originalRoundTripper.RoundTrip(request)
```
cleptric, line 92: We also want to handle the err case

httpclient/sentryhttpclient.go
aldy505: When I wrote this, the docs said trace propagation targets is only for avoiding CORS-related issue, and since the dotnet SDK does not have that option too, I don't implement it here.

But is my assumption wrong here? Should all SDK implements trace propagation targets then?

httpclient/sentryhttpclient.go
aldy505: Is it not ideal? Should I remove it?

httpclient/sentryhttpclient.go
cleptric: Yes

httpclient/sentryhttpclient.go
cleptric: What info is in there?

aldy505: Decided to just remove it.

client.go
```diff
@@ -133,6 +133,10 @@ type ClientOptions struct {
 	TracesSampleRate float64
 	// Used to customize the sampling of traces, overrides TracesSampleRate.
 	TracesSampler TracesSampler
+	// Control with URLs trace propagation should be enabled. Does not support regex patterns.
+	TracePropagationTargets []string
+	// When set to true, the SDK will start a span for outgoing HTTP OPTIONS requests.
+	TraceOptionsRequests bool
```
cleptric, line 139: Why did you add this option?

aldy505: Saw it on the dev docs. Should I not implement that?

cleptric: Link?

aldy505: Here: https://develop.sentry.dev/sdk/telemetry/traces/#traceoptionsrequests

cleptric: > When set to true **transactions should be created** for HTTP OPTIONS requests.

This option does not apply to outgoing HTTP requests.

aldy505: Okay. Will update that later.

giortzisg (changes requested):
Looks good, only the description is missing from the span.

httpclient/sentryhttpclient.go
```diff
@@ -0,0 +1,140 @@
...
+		span.SetData("http.response.status_code", response.StatusCode)
+		span.SetData("http.response_content_length", response.ContentLength)
+	}
+
+	return response, err
+}
+
+// SentryHTTPClient provides a default HTTP client with SentryRoundTripper included.
+// This can be used directly to perform HTTP request.
+var SentryHTTPClient = &http.Client{
```
giortzisg, line 138: nit: can we change the default name here? it's already too verbose and the package already specifies that we are referring to the sentryhttpclient.
```suggestion
var Client = &http.Client{
```

_examples/httpclient/main.go
```diff
@@ -0,0 +1,75 @@
...
+		Transport: sentryhttpclient.NewSentryRoundTripper(nil),
+	}
+
+	err := getExamplePage(ctx, httpClient)
+	if err != nil {
+		panic(err)
+	}
+
+	// With Sentry's HTTP client
+	err = getExamplePage(ctx, sentryhttpclient.SentryHTTPClient)
```
giortzisg, line 41: fix the name here.
```suggestion
	err = getExamplePage(ctx, sentryhttpclient.Client)
```

giortzisg (commented):
Looks good, let's fix the lint errors and the name in the examples and then it's ready to be merged.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add net/http client integration ([#876](https://github.com/getsentry/sentry-go/pull/876))
+
```
cleptric, line 8:
```suggestion

```

cleptric:
Not sure we actually want to go the route with the round tripper. But we take another look after logs.

httpclient/sentryhttpclient.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	return response, err
+}
+
+// SentryHTTPClient provides a default HTTP client with SentryRoundTripper included.
+// This can be used directly to perform HTTP request.
+var Client = &http.Client{
+	Transport: NewSentryRoundTripper(http.DefaultTransport),
+}
```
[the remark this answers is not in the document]
giortzisg: cursor is correct here.

httpclient/sentryhttpclient.go
```diff
@@ -0,0 +1,139 @@
...
+	response, err := s.originalRoundTripper.RoundTrip(request)
+	if err != nil {
+		span.Status = sentry.SpanStatusInternalError
+	}
+
+	if response != nil {
+		span.Status = sentry.HTTPtoSpanStatus(response.StatusCode)
+		span.SetData("http.response.status_code", response.StatusCode)
+		span.SetData("http.response_content_length", response.ContentLength)
+	}
```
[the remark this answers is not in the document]
giortzisg: we can early return on error here.

httpclient/sentryhttpclient.go
```diff
@@ -0,0 +1,134 @@
...
+			request.Header.Add("Sentry-Trace", hub.GetTraceparent())
+		}
+
+		return s.originalRoundTripper.RoundTrip(request)
+	}
+
+	cleanRequestURL := request.URL.Redacted()
+
+	span := parentSpan.StartChild("http.client", sentry.WithDescription(fmt.Sprintf("%s %s", request.Method, cleanRequestURL)))
+	defer span.Finish()
```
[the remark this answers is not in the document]
aldy505: @giortzisg do you want to leave it like this, or do you want to not send spans when the http method is OPTIONS?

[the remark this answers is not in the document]
giortzisg: No we need this. Cursor is getting confused by traceOptionRequests

httpclient/sentryhttpclient_test.go
```diff
@@ -0,0 +1,464 @@
...
+				Origin:      "manual",
+				Sampled:     sentry.SampledTrue,
+				Status:      sentry.SpanStatusInternalError,
+			},
+		},
+		{
+			RequestMethod: "OPTIONS",
+			RequestURL:    "https://example.com",
+			WantError:     false,
+			WantSpan:      nil,
```
giortzisg: also we should get a span here.

_examples/httpclient/main.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	// With Sentry's HTTP client
+	err = getExamplePage(ctx, sentryhttpclient.Client)
+	if err != nil {
+		panic(err)
+	}
+}
```
giortzisg: we can remove this

httpclient/sentryhttpclient_test.go
```diff
@@ -0,0 +1,479 @@
...
+				break
+			}
+		}
+
+		if tt.WantSpan != nil && !foundMatch {
+			t.Errorf("Span mismatch (-want +got):\n%s", strings.Join(diffs, "\n"))
+		} else if tt.WantSpan == nil && foundMatch {
+			t.Errorf("Expected no span, got %+v", gotSpans)
+		}
+	}
```
giortzisg, line 316: Cursor is right for the test. We need to check for empty spans in the beginning 
```suggestion
	for i, tt := range tests {
		gotSpans := got[i]

		if tt.WantSpan == nil {
			if len(gotSpans) > 0 {
				t.Errorf("Expected no span, got %+v", gotSpans)
			}
			continue
		}

		var foundMatch = false
		var diffs []string
		for _, gotSpan := range gotSpans {
			if diff := cmp.Diff(tt.WantSpan, gotSpan, optstrans); diff != "" {
				diffs = append(diffs, diff)
			} else {
				foundMatch = true
				break
			}
		}

		if !foundMatch {
			t.Errorf("Span mismatch (-want +got):\n%s", strings.Join(diffs, "\n"))
		}
	}
```

PR #878, opened 2024-08-30
fix(hub): Do not call AddBreadcrumb if breadcrumb is nil

giovanni-orciuolo:
This PR will allow the function `AddBreadcrumb` to work even if you pass nil as the breadcrumb. I know it's never supposed to be nil, however in Go if you want something to never be nil you don't use a pointer for it. I thought of making the parameter not a pointer but that would have some implication for the codebase so I decided against that, and I added only 2 nil checks in the function.


ribice:
What's `oryxBuildBinary`?

I'm against this, but just a note that it would be much simpler and more readable if you want to achieve this, add a check at beginning (if breadcrumb == nil {return}) instead of checking twice later and doing the same thing.

giovanni-orciuolo:
> What's `oryxBuildBinary`?
> 
> I'm against this, but just a note that it would be much simpler and more readable if you want to achieve this, add a check at beginning (if breadcrumb == nil {return}) instead of checking twice later and doing the same thing.

I don't know, it appeared when I ran `go test`. Regarding the check at the beginning, I thought it would have prevented the BeforeBreadcrumb from being executed so I didn't do it. If we don't want BeforeBreadcrumb to be executed when breadcrumb is nil, then it's ok to put the check at the beginning.

aldy505:
> What's `oryxBuildBinary`?

Found an answer here: https://github.com/orgs/community/discussions/18856#discussioncomment-3142633

It's a build artifacts that's created when you're using Codespaces.

PR #887, opened 2024-10-04
fix(tracing): Correct sentry-trace header parsing for multiple trace IDs

shlyamster:
### Problem

The `sentry-trace` header parsing fails when multiple trace IDs are present due to strict matching with `^` and `$` in the regex.

### Solution

This PR fixes the regex by removing `^` and `$` anchors, allowing to correctly parse headers with multiple trace IDs.

### Tests

Added unit tests for parsing multiple trace IDs in the `sentry-trace` header.

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -383,7 +383,7 @@ func (s *Span) doFinish() {
 //
 //	TRACE_ID - SPAN_ID - SAMPLED
 //	[[:xdigit:]]{32}-[[:xdigit:]]{16}-[01]
-var sentryTracePattern = regexp.MustCompile(`^([[:xdigit:]]{32})-([[:xdigit:]]{16})(?:-([01]))?$`)
+var sentryTracePattern = regexp.MustCompile(`([[:xdigit:]]{32})-([[:xdigit:]]{16})(?:-([01]))?`)
 
 // updateFromSentryTrace parses a sentry-trace HTTP header (as returned by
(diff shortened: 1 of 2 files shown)


cleptric:
This is not supported by our trace header spec, may I ask where you’ve seen such header values?

shlyamster:
I get these headers from my frontend on Next JS. And because of this, the trace of the request to the backend is incorrectly attached to the trace from the frontend. The first span ID is correct, while the second one is not logged into the system at all. There are no proxies and similar services in front of the backend, only Cloudflare.

cleptric:
Which version of the Sentry Next.js SDK are you using? Anything else noteworthy of your setup?
If you’re using Sentry SaaS (sentry.io), what is your org slug?
This might be a bug in the Next SDK, hence we’ll take a look there instead of moving forward with this PR.

lforst:
@shlyamster Maintainer of the Next.js SDK here. This indeed looks like a bug in the Next.js SDK more than the PHP SDK. Would you mind opening an issue over at getsentry/sentry-javascript with details on how you set up the Next.js SDK and maybe some info on the spans generated by Next.js? Thank you!

shlyamster:
@lforst Thank you for reviewing this pull request. I have created an [issue](https://github.com/getsentry/sentry-javascript/issues/13870) in the `sentry-javascript` repository.

@cleptric If the issue is confirmed to originate from Next.js, this pull request can be considered closed.

Issue #884, opened 2024-09-12
"Missing Trace Root" on v0.29


Cyberax:
## Summary
Upgrade to v0.29 results in "Missing Trace Root" traces. Downgrade to 0.28.1 resolves the issue.

## Steps To Reproduce

Use the Go OTEL sample

```go
	sp := sentryotel.NewSentrySpanProcessor()
	tracerProvider := sdktrace.NewTracerProvider(
		sdktrace.WithResource(serviceResource),
		sdktrace.WithSpanProcessor(sp),
	)
```

## Expected Behavior
The trace root should be correct.

## Screenshots
![Image](https://github.com/user-attachments/assets/b033bf69-4ffd-4057-b03a-a31555dc6e45)

### SDK
 - `sentry-go` version: v0.29
 - Go version: 1.22
 - Using Go Modules? yes

### Sentry
 - Using hosted Sentry in sentry.io? yes

cleptric:
We'll take a look.

cleptric:
This is currently to be expected. The reason is adding support to something we call "Tracing without performance", a way to connect errors from different services together without the need to emit traces. As the payload of each event now always contains a `parent_span_id`, the product now displays "Missing Trace Root". We'll make the changes in Sentry itself. I'll get back here once I have a rough timeline.

radazen:
@cleptric I may be wrong, but it feels like there's still a bug here. Note the part of the original screenshot where two different spans in the same trace have the same `Span ID` (2e28b774). This is happening to me as well! My top-level transaction span is what I'd expect to see as the Trace Root, and if I record a trace with a single span and no children, that's exactly what happens: I see a proper trace root, not a `Missing Trace Root`.

But if I record any child spans, the root transaction span ends up having its original `Span ID` overwritten by the span ID of one of the child spans, and I end up with two spans with the same `Span ID` and a `Missing Trace Root`.

cleptric:
Yes, I was able to reproduce this and will look into a fix. Thanks @radazen for calling this out!

bububa:
Same problem, really annoying on self host

ribice:
@Cyberax / @radazen I was able to reproduce this as well using the tracing without performance, but downgrading to 0.28.1 lacks this feature, so I have to pull some commits in.

Could you share a reproducible snippet?

cleptric:
A fix has been shipped with https://github.com/getsentry/sentry-go/releases/tag/v0.29.1

PR #888, opened 2024-10-07
Fix applying the `trace` context on events

cleptric:
Fixes https://github.com/getsentry/sentry-go/issues/884

We mistakenly used the scope's span `trace` context as the `trace` context for the transaction. Instead, we now set the `trace` context on the event when the transaction finishes and only use the scope's span `trace` context in case there is no `trace` context on the event already.

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -383,6 +383,15 @@ func (scope *Scope) ApplyToEvent(event *Event, hint *EventHint, client *Client) 
 		}
 
 		for key, value := range scope.contexts {
+			if key == "trace" && event.Type == transactionType {
+				// Do not override trace context of
+				// transactions, otherwise it breaks the
+				// transaction event representation.
+				// For error events, the trace context is used
+				// to link errors and traces/spans in Sentry.
+				continue
+			}
+
 			// Ensure we are not overwriting event fields
 			if _, ok := event.Contexts[key]; !ok {
 				event.Contexts[key] = cloneContext(value)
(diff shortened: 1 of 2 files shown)


Darevski:
@cleptric  This fix is a partially functional solution that fixes the trace context only for transactions.

It does not resolve the tracing information for events such as messages, exceptions, or any other attributes  related to scope behavior (like breadcrumbs etc) based on the current span settings here: https://github.com/getsentry/sentry-go/blob/master/tracing.go#L197-L198.

Here is an example of how the following code behaves:
```go
        // Start the parent transaction
	ctx := context.Background()
	parentTransaction := sentry.StartTransaction(ctx, "parent-transaction")

	// Start the parent transaction span
	parentSpan := sentry.StartSpan(parentTransaction.Context(), "parent-span")

	parentChildSpan := sentry.StartSpan(parentSpan.Context(), "parent-child-span")

	// Capture message that should be attached to "parent-child-span"
	sentry.CaptureMessage("Parent child span")

	// Finish the  "parent-child-span"
	parentChildSpan.Finish()

	// Capture message that should be attached to "parent-span", but will be attached to the "parent-child-span"
	sentry.CaptureMessage("Parent span")

	// Finish the "parent span"
	parentSpan.Finish()

	// Capture message that should be attached to "parent-transaction", but will be attached to the "parent-child-span"
	sentry.CaptureMessage("Parent transaction completed")

	parentTransaction.Finish()
```

On the right side is the code running with PR #886, and on the left side is the code running with the current PR #888.
In #888 all sentry events are related to the parent-child-span, but its wrong.

https://github.com/user-attachments/assets/6056f7c9-612f-416a-ad52-d904fb53cd5f

szokeasaurusrex (approved):
stamp

PR #893, opened 2024-10-18
feat: database/sql integration

### LINKED ISSUE #1128: Database/SQL Integration
Adds database/sql instrumentation to the Sentry Go SDK via a new sentrysql package. This integration wraps SQL drivers and connections to automatically create spans for queries, transactions, and statements, enabling better visibility into database performance and errors.

aldy505:
(From this [comment](https://github.com/getsentry/sentry-go/pull/893#issuecomment-3690840573)): 

For those wondering why this is taking too long:

We are lacking two missing features for this integration:
1. Client-side scrubbing
2. Parameterized span attributes (need to have some kind of SQL lexer. Easy to do, not implemented yet).

If we ship this integration with or without those 2 missing features, we're a bit concerned about the performance penalty (or performance bottleneck) for users.

Another option would be just ship the integration and mark it as "experimental and might be removed in very near future", gather feedback about the performance & quality. If it's good enough, we might continue implement those missing features.

---

Queries insights tracing for Go.

closes #1128 
closes GO-95


sentrysql/conn.go
```diff
@@ -0,0 +1,247 @@
...
+func (s *sentryConn) Begin() (driver.Tx, error) {
+	tx, err := s.originalConn.Begin() //nolint:staticcheck We must support legacy clients
+	if err != nil {
+		return nil, err
+	}
+
+	return &sentryTx{originalTx: tx, ctx: s.ctx, config: s.config}, nil
+}
+
+func (s *sentryConn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, error) {
```
ribice, line 63: Seems like you need `ConnBeginTx`?

sentrysql/conn.go
```diff
@@ -0,0 +1,247 @@
...
+		return nil, driver.ErrSkip
+	}
+
+	parentSpan := sentry.SpanFromContext(s.ctx)
+	if parentSpan == nil {
+		return queryer.Query(query, args)
+	}
+
+	span := parentSpan.StartChild("db.sql.query", sentry.WithDescription(query))
+	if s.config.databaseSystem != "" {
```
ribice, line 93: Instead of repeating this for times, perhaps implement a method on `sentrySQLConfig` to set the Span Data?

sentrysql/stmt.go
```diff
@@ -0,0 +1,190 @@
...
+		span.SetData("server.address", s.config.serverAddress)
+	}
+	if s.config.serverPort != "" {
+		span.SetData("server.port", s.config.serverPort)
+	}
+
+	result, err := s.originalStmt.Exec(args) //nolint:staticcheck We must support legacy clients
+	if err != nil {
+		span.Status = sentry.SpanStatusInternalError
+		span.Finish()
```
ribice, line 49: Defer this above and you can get rid of 54.

sentrysql/stmt.go
```diff
@@ -0,0 +1,190 @@
...
+
+	return rows, nil
+}
+
+func (s *sentryStmt) ExecContext(ctx context.Context, args []driver.NamedValue) (driver.Result, error) {
+	// should only be executed if the original driver implements StmtExecContext
+	stmtExecContext, ok := s.originalStmt.(driver.StmtExecContext)
+	if !ok {
+		// fallback to the so-called deprecated "Exec" method
+		var values []driver.Value
```
ribice, line 99: You can preallocate this.

sentrysql/stmt.go
```diff
@@ -0,0 +1,190 @@
...
+		span.SetData("server.address", s.config.serverAddress)
+	}
+	if s.config.serverPort != "" {
+		span.SetData("server.port", s.config.serverPort)
+	}
+
+	result, err := stmtExecContext.ExecContext(ctx, args)
+	if err != nil {
+		span.Status = sentry.SpanStatusInternalError
+		span.Finish()
```
ribice, line 128: Same here

sentrysql/stmt.go
aldy505: I'm gonna need to read the driver package docs again, see if I can just throw driver.ErrSkipped (or ErrSkip, I forgot)

aldy505:
I really don't like golangci-lint, it's so pedantic :(

aldy505:
Should be good by now.

sentrysql/operation.go
```diff
@@ -0,0 +1,22 @@
...
+
+func parseDatabaseOperation(query string) string {
+	// The operation is the first word of the query.
+	operation := query
+	if i := strings.Index(query, " "); i >= 0 {
+		operation = strings.ToUpper(query[:i])
+	}
+
+	// Only returns known words.
+	for _, knownOperation := range knownDatabaseOperations {
```
ribice, line 15: You can make `knownDatabaseOperations` a map[string]bool (or map[string]struct{}), and skip the for range.

Something like:

```
	operation := strings.ToUpper(strings.Split(query, " ")[0])
	if _, exists := knownDatabaseOperations[operation]; exists {
		return operation
	}
	return ""
```

sentrysql/operation_test.go
```diff
@@ -0,0 +1,50 @@
+package sentrysql
+
+import "testing"
+
+func TestParseDatabaseOperation(t *testing.T) {
+	tests := []struct {
+		name  string
```
ribice, line 7: Just a suggestion, I prefer tests in format map[string]struct, where key is the name of the test. This makes it easier to read the name of the test (due to indentation), and prevents duplicate names.

sentrysql/operation.go
aldy505: Okay, might do this.

sentrysql/operation_test.go
aldy505: I actually don't like this way of testing (using `tests` array of struct and then iterate over it), I prefer something like `TestParseDatabaseOperation_CaseXXX` kind of thing.

ribice:
I'll test this locally myself this week and report back. We should also update the docs and main repository with this integration.

ribice:
You're missing examples in _examples. These should also include example on how to set the DSN.

aldy505:
> You're missing examples in _examples. These should also include example on how to set the DSN.

@ribice I've been occupied so much with work this week. Will find the time to work on this later.

ribice:
LGTM Overall. Still misses some tests for the coverage, entry in CHANGELOG.MD and documentation in sentry/sentry-docs.

aldy505:
> LGTM Overall. Still misses some tests for the coverage

To be honest, I don't know how to make some of that as covered other than using old Go versions. But I might try creating another database driver implementation that does not implement `XXXContext` method.

> ...documentation in main repo and sentry-docs.

I might be able to help with the sentry-docs one, since I also need to update some of the self-hosted docs.

cleptric:
I haven't reviewed this PR in full yet, but some things I would like to be changed:

- I would suggest we start with https://pkg.go.dev/database only, I'm not keen to add support for 3rd party libraries right away.
- If we end up adding 3rd party dependencies, we should publish this as it's own package. See https://github.com/getsentry/sentry-go/tree/master/otel for prior art.
- We also support Mongo DB in our Insights query module, so the naming of the package could be more generic, such as `sentrydatabase`
- `nolint` is an immediate code smell.

aldy505:
> * I would suggest we start with https://pkg.go.dev/database only, I'm not keen to add support for 3rd party libraries right away.
> * If we end up adding 3rd party dependencies, we should publish this as it's own package. See https://github.com/getsentry/sentry-go/tree/master/otel for prior art.

We are not supporting 3rd party libraries here. We only import the database driver since [Go does not bring built-in database drivers and rely on third party drivers](https://go.dev/wiki/SQLDrivers). The database drivers we're using should not going to be compiled to the resulting binary since it's only used in tests.

Although it seems like we're using it as a library here, it is not, it's still considered a database driver:
https://github.com/getsentry/sentry-go/blob/c5fa49de994f32da77499c632e8e512724556a95/_examples/sql/main.go#L40-L45

> * We also support Mongo DB in our Insights query module, so the naming of the package could be more generic, such as `sentrydatabase`

MongoDB does not have and does not conform to the `database/sql` driver. So we're not going to add it into the enum here. Instrumentation for MongoDB should be specific for the Go MongoDB library. 

> * `nolint` is an immediate code smell.

Since we're wrapping a database driver and thus making us another database driver, we are required to implement multiple methods that are considered _duplicates_ by the linter, yet again, from the [driver documentation](https://pkg.go.dev/database/sql/driver@go1.23.2#pkg-overview), we're still required to implement those.

sentrysql/stmt.go
```diff
@@ -0,0 +1,159 @@
...
+
+	result, err := s.originalStmt.Exec(args) //nolint:staticcheck // We must support legacy clients
+	if err != nil {
+		span.Status = sentry.SpanStatusInternalError
+		return nil, err
+	}
+
+	span.Status = sentry.SpanStatusOK
+
+	return result, nil
```
aldy505, line 48: This section is not covered by tests, yet I don't know how to make it so.

I've added this check beforehand:
```go
if s.ctx == nil && ctx != nil {
  s.ctx = ctx
}
```

But it doesn't seem to help anyway.

jazzido:
Don't let this die! :) I'm sure many of us in the community will be grateful for an officially supported DB integration

go.mod
```diff
@@ -3,8 +3,11 @@ module github.com/getsentry/sentry-go
 go 1.21
 
 require (
+	github.com/glebarez/go-sqlite v1.21.1
```
giortzisg, line 6: should still move driver dependencies to the sentrysql submodule

sentrysql/conn.go
```diff
@@ -0,0 +1,260 @@
...
+		config:       s.config,
+	}, nil
+}
+
+func (s *sentryConn) PrepareContext(ctx context.Context, query string) (driver.Stmt, error) {
+	// should only be executed if the original driver implements ConnPrepareContext
+	connPrepareContext, ok := s.originalConn.(driver.ConnPrepareContext)
+	if !ok {
+		// We can't return driver.ErrSkip here. We should fall back to Prepare without context.
+		return s.Prepare(query)
```
giortzisg, line 57: we should update sentryConn.ctx here right? I know we don't create a span on `Prepare()` but should at least have the correct context.

sentrysql/conn.go
```diff
@@ -0,0 +1,260 @@
...
+	return &sentryTx{originalTx: tx, ctx: s.ctx, config: s.config}, nil
+}
+
+func (s *sentryConn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, error) {
+	// should only be executed if the original driver implements ConnBeginTx
+	connBeginTx, ok := s.originalConn.(driver.ConnBeginTx)
+	if !ok {
+		// We can't return driver.ErrSkip here. We should fall back to Begin without context.
+		return s.Begin()
+	}
```
giortzisg, line 92: same thing here?

sentrysql/conn.go
```diff
@@ -0,0 +1,260 @@
...
+func (s *sentryConn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
+	// should only be executed if the original driver implements QueryerContext
+	queryerContext, ok := s.originalConn.(driver.QueryerContext)
+	if !ok {
+		values, err := namedValueToValue(args)
+		if err != nil {
+			return nil, err
+		}
+
+		s.ctx = ctx
```
giortzisg, line 139: Nice change. I think this was the reason why the legacy functions weren't covered. However that would mean that for legacy drivers the spans were incorrent, can we add a test case for that?

sentrysql/conn.go
```diff
@@ -0,0 +1,260 @@
...
+		return s.Query(query, values)
```
giortzisg, line 140: Also to my understanding since the sentry conn always implements `QueryContext`, we really don't want the Query function? Would this mean that maybe we can handle everything in queryCtx and handle both cases here to avoid duplication?

giortzisg (commented):
Overall looks good with the most recent changes and should be functionally correct. Just wanna be sure that the context is passed correctly and also add some minor improvements

sentrysql/conn.go
aldy505: Can you put a link to the incorrect spans?

sentrysql/sentrysql_legacy_test.go
```diff
@@ -0,0 +1,697 @@
...
+
+			var diffs []string
+			for _, gotSpan := range gotSpans {
+				if diff := cmp.Diff(tt.WantSpan, gotSpan, optstrans); diff != "" {
+					diffs = append(diffs, diff)
+				} else {
+					foundMatch = true
+					break
+				}
+			}
```
giortzisg, line 142: the tests here are incorrent. If we never receive spans then the check for getting the expected span is skipped. Applies to non legacy drivers as well.

giortzisg (changes requested):
some test changes.

sentrysql/conn.go
giortzisg: added some comments.

sentrysql/sentrysql_legacy_test.go
aldy505: I'm not sure what you mean by that. the expected span of each `gotSpan` variable is checked through the `cmp.Diff`. Or I misunderstand you?

giortzisg: if gotSpans is empty you never enter the for loop and don't check the tt.WantSpan. So there's a case where you expect to receive spans, you don't, and the tests pass.

aldy505: Done!

sentrysql/sentrysql_connector_test.go
```diff
@@ -128,6 +128,18 @@ func TestNewSentrySQLConnector_Integration(t *testing.T) {
...
+				t.Errorf("Expecting no spans, but got %d spans: %v", len(gotSpans), gotSpans)
+				continue
+			}
+
+			// if WantSpan is not nil, we should have at least one span
+			if len(gotSpans) == 0 {
+				t.Errorf("Expecting at least one span, but got %d spans: %v", len(gotSpans), gotSpans)
+				continue
+			}
+
```
giortzisg, line 142: Can we not have `tt.WantSpans` as a `[]Span` so that we can compare immediately with `gotSpans`? I'd say that it's not very readable.

sentrysql/sentrysql_test.go
```diff
@@ -0,0 +1,1418 @@
...
+	),
+}
+
+func TestMain(m *testing.M) {
+	sql.Register("sentrysql-sqlite", sentrysql.NewSentrySQL(&sqlite.Driver{}, sentrysql.WithDatabaseName("memory"), sentrysql.WithDatabaseSystem(sentrysql.DatabaseSystem("sqlite")), sentrysql.WithServerAddress("localhost", "5432")))
+	// sentrysql-legacy is used by `sentrysql_legacy_test.go`
+	sql.Register("sentrysql-legacy", sentrysql.NewSentrySQL(ldriver, sentrysql.WithDatabaseSystem(sentrysql.DatabaseSystem("legacydb")), sentrysql.WithDatabaseName("fake")))
+
+	os.Exit(m.Run())
+}
```
giortzisg, line 33: This is obfuscated from the other tests. Should probably be an `init` statement on each file that registers whatever we need only for the tests on the file.

sentrysql/sentrysql_test.go
```diff
@@ -0,0 +1,1418 @@
...
+					diffs = append(diffs, diff)
+				} else {
+					foundMatch = true
+					break
+				}
+			}
+
+			if !foundMatch {
+				t.Errorf("Span mismatch (-want +got):\n%s", strings.Join(diffs, "\n"))
+			}
```
giortzisg, line 169: Same thing applies here. We should probably use a `[]Span` for `WantSpans`

sentrysql/sentrysql_test.go
aldy505: From the docs: If Register is called twice with the same name or if driver is nil, it panics.

I believe setting this here is the "right" way to do so, since it'll be run after any library's `init` invocation.

sentrysql/stmt.go
```diff
@@ -0,0 +1,164 @@
...
+		values, err := namedValueToValue(args)
+		if err != nil {
+			return nil, err
+		}
+
+		s.ctx = ctx
+		return s.Query(values)
+	}
+
+	parentSpan := sentry.SpanFromContext(s.ctx)
```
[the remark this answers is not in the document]
aldy505: This is wrong. `ctx` provided from the function argument does not have sentry span context. SInce what matters is the context in which the statement was started.

sentrysql/stmt.go
```diff
@@ -0,0 +1,164 @@
...
+		values, err := namedValueToValue(args)
+		if err != nil {
+			return nil, err
+		}
+
+		s.ctx = ctx
+		return s.Exec(values)
+	}
+
+	parentSpan := sentry.SpanFromContext(s.ctx)
```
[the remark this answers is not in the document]
aldy505: This is wrong. `ctx` provided from the function argument does not have sentry span context. SInce what matters is the context in which the statement was started.

sentrysql/tx.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"context"
+	"database/sql/driver"
+)
+
+type sentryTx struct {
+	originalTx driver.Tx
+	ctx        context.Context
+	config     *sentrySQLConfig
```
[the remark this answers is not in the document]
aldy505: Tracing on `Commit` / `Rollback` would be useless.

aldy505:
@cleptric @giortzisg any updates?

sentrysql/conn.go
```diff
@@ -0,0 +1,262 @@
...
+		s.ctx = ctx
+		return s.Begin()
+	}
+
+	tx, err := connBeginTx.BeginTx(ctx, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	return &sentryTx{originalTx: tx, ctx: s.ctx, config: s.config}, nil
```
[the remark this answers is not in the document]
giortzisg: the comment is actually correct here, we should derive the `ctx` from the caller and if that is nil then fallback to the parent ctx.

sentrysql/conn.go
```diff
@@ -0,0 +1,262 @@
...
+	parentSpan := sentry.SpanFromContext(ctx)
+	if parentSpan == nil {
+		return queryerContext.QueryContext(ctx, query, args)
+	}
+
+	span := parentSpan.StartChild("db.sql.query", sentry.WithDescription(query))
+	s.config.SetData(span, query)
+	defer span.Finish()
+
+	rows, err := queryerContext.QueryContext(ctx, query, args)
```
giortzisg: We should propagate the span context here.
```suggestion
	rows, err := queryerContext.QueryContext(span.Context(), query, args)
```

sentrysql/conn.go
```diff
@@ -0,0 +1,262 @@
...
+	parentSpan := sentry.SpanFromContext(ctx)
+	if parentSpan == nil {
+		return execerContext.ExecContext(ctx, query, args)
+	}
+
+	span := parentSpan.StartChild("db.sql.exec", sentry.WithDescription(query))
+	s.config.SetData(span, query)
+	defer span.Finish()
+
+	rows, err := execerContext.ExecContext(ctx, query, args)
```
giortzisg:
```suggestion
	rows, err := execerContext.ExecContext(span.Context(), query, args)
```

sentrysql/stmt.go
```diff
@@ -0,0 +1,164 @@
...
+	if parentSpan == nil {
+		return stmtExecContext.ExecContext(ctx, args)
+	}
+
+	span := parentSpan.StartChild("db.sql.exec", sentry.WithDescription(s.query))
+	s.config.SetData(span, s.query)
+	defer span.Finish()
+
+	result, err := stmtExecContext.ExecContext(ctx, args)
```
giortzisg:
```suggestion
	result, err := stmtExecContext.ExecContext(span.Context(), args)
```

sentrysql/stmt.go
```diff
@@ -0,0 +1,164 @@
...
+	parentSpan := sentry.SpanFromContext(s.ctx)
+	if parentSpan == nil {
+		return stmtQueryContext.QueryContext(ctx, args)
+	}
+
+	span := parentSpan.StartChild("db.sql.query", sentry.WithDescription(s.query))
+	s.config.SetData(span, s.query)
+	defer span.Finish()
+
+	rows, err := stmtQueryContext.QueryContext(ctx, args)
```
giortzisg:
```suggestion
	rows, err := stmtQueryContext.QueryContext(span.Context(), args)
```

sentrysql/stmt.go
```diff
@@ -0,0 +1,164 @@
...
+		return s.Query(values)
+	}
+
+	parentSpan := sentry.SpanFromContext(s.ctx)
+	if parentSpan == nil {
+		return stmtQueryContext.QueryContext(ctx, args)
+	}
+
+	span := parentSpan.StartChild("db.sql.query", sentry.WithDescription(s.query))
+	s.config.SetData(span, s.query)
```
giortzisg, line 173: we need to set span data according to this page https://develop.sentry.dev/sdk/telemetry/traces/sql-transactions/

giortzisg (commented):
We can keep the tests as is for now. Added a few comments, let's amend them and bump the go version and we should be good to merge this.

aldy505:
For those wondering why this is taking too long:

We are lacking two missing features for this integration:
1. Client-side scrubbing
2. Parameterized span attributes (need to have some kind of SQL lexer. Easy to do, not implemented yet).

If we ship this integration with or without those 2 missing features, we're a bit concerned about the performance penalty (or performance bottleneck) for users.

Another option would be just ship the integration and mark it as "experimental and might be removed in very near future", gather feedback about the performance & quality. If it's good enough, we might continue implement those missing features.

giortzisg:
Just an update on this, the PR has expanded to the point where it's really hard to review and there are still multiple issues, so will be closing this and trying to properly split and salvage some parts in a way that makes sense. Will be taking over the sql integration.

PR #920, opened 2024-12-09
Add WithContext variants of methods

### LINKED ISSUE #555: [OTel] Expose context-aware methods for capturing events in public API
Extracted from https://github.com/getsentry/sentry-go/pull/537.

### Problem

Associating errors with transactions is generally implemented (see `otel/event_processor.go`), but there's a caveat: the `EventHint` of the event processor has to contain a valid Context with a set OpenTelemetry span. This means that e.g. messages captured with `sentry.CaptureMessage` or `hub.CaptureMessage` won't be linked to Sentry spans at the moment.

So to properly link the raised exception now users have to do things like:

```go
	client := sentry.CurrentHub().Client()
	client.CaptureException(
		errors.New("test sentry exception"),
		&sentry.EventHint{Context: ctx},
		nil,
	)
```

### Proposed Solution

We might want to expose methods like `CaptureMessageWithContext(ctx context.Context, message string)` and others that also take a `ctx` parameter. Under the hood, they would set add the passed `ctx` to `EventHint`.

LesDomen:
Added WithContext variants of captureException, captureMessage, captureCheckIn and captureEvent, that pass the context down into the EventHint

Closes #555

### DIFF
--- a/client.go
+++ b/client.go
@@ -435,6 +435,16 @@ func (client *Client) CaptureCheckIn(checkIn *CheckIn, monitorConfig *MonitorCon
 	return nil
 }
 
+// CaptureCheckInWithContext captures a check in.
+func (client *Client) CaptureCheckInWithContext(ctx context.Context, checkIn *CheckIn, monitorConfig *MonitorConfig, scope EventModifier) *EventID {
+	event := client.EventFromCheckIn(checkIn, monitorConfig)
+	if event != nil && event.CheckIn != nil {
+		client.CaptureEvent(event, &EventHint{Context: ctx}, scope)
+		return &event.CheckIn.ID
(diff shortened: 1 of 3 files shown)


LesDomen:
Any other ideas for linking between captured exceptions and transactions/spans?

cleptric:
If you're using Sentry's tracing APIs, it will just work out of the box. For Otel, there is a workaround provided int he issue you linked.

LesDomen:
We're moving from Elastic APM to OpenTelemetry, and shipping traces to both Sentry and other OTEL Backends. Is there any viable way to write a wrapper around sentry.CaptureException, that will still report the initial error location?

PR #931, opened 2024-12-17
add setHubOnContext functions for integrations that use custom context

ribice:


iris/sentryiris.go
```diff
@@ -114,6 +114,10 @@ func GetHubFromContext(ctx iris.Context) *sentry.Hub {
 	return nil
 }
 
+func SetHubOnContext(ctx iris.Context, hub *sentry.Hub) {
+	ctx.Values().Set(valuesKey, hub)
```
ribice, line 118: @cleptric The warning seems like a false negative, since I added there are tests for this.

cleptric: Which warning?

ribice: CleanShot 2024-12-17 at 10 56 23@2x

PR #935, opened 2024-12-21
Implement flush with context

### LINKED ISSUE #746: Flush to receive go context type
## Summary
Currently `Flush` function receives a timeout, so when it is called someone needs to provide a timeout value. By using a context of the app and propagating to `Flush` the operation will timeout whenever the context expires, so no explicit timeout is needed.

## Motivation
This solution will remove some corner-case in the blocking operations of an application. If a long timeout is specified the application might wait for long. If the application already performs some finalisation and has an overall shutdown timeout, we would need to calculate the duration left to timeout, in order to provide to `Flush`.

## Additional Context
-

ribice:
Resolves #746 
Closes GO-25


ribice:
The only thing failing here is the coverage. Do I have to 'fix' it or we can merge as-is?

cleptric:
Do we have docs for this?

ribice:
> Do we have docs for this?

No. Should I add it in `sentry-docs`?

cleptric:
Yes

ribice:
@cleptric https://github.com/getsentry/sentry-docs/pull/12308

cleptric (approved):

PR #936, opened 2024-12-22
Add HubProvider functionality for logrus integration

### LINKED ISSUE #513: Add goroutine support in logrus implementation
## Summary
The logrus implementation always use the same Sentry hub. Logrus loggers are often shared between goroutines. Your documentation states that using separate hubs for each goroutine is mandatory. 
## Motivation
With the current logrus implementation a new logger is mandatory for each goroutine. It would be nice if the implementation would be able to integrate with a single logger, while still being able to use the logger in different goroutines.

ribice:
Resolves #513

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,26 @@
 
 Add ability to override `hub` in `context` for integrations that use custom context ([#931](https://github.com/getsentry/sentry-go/pull/931))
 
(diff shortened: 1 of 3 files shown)


ribice:
The only thing failing here is the coverage (81% instead of 83%). Do I have to 'fix' it or we can merge as-is?

cleptric:
Do we have docs for this?

cleptric (approved):

PR #919, opened 2024-12-08
Create modules for integrations

ribice:

PR #942, opened 2024-12-30
Revert “Create modules for integrations (#919)”

cleptric:
This is currently causing issues, as even without a new tag on the master branch, these new modules are being picked up as something like `github.com/getsentry/sentry-go/fiber v0.0.0-20241230225403-109fca191e09` which breaks current installations.


ribice:
Are you going to merge these immediately prior to new release (to prevent these issues) or will leave them as they were?

cleptric:
We need to merge this ASAP, and can figure out the problem afterwards.

ribice (approved):

PR #956, opened 2025-01-22
Add date range to licence

philipphofmann:
To align with all other Sentry SDKs.

#skip-changelog


cleptric:
We do not bump the license year. This was discussed a while ago, can't find a link though.

PR #958, opened 2025-01-25
Add zap integration

ribice:
TODO:

- Write docs for sentrydocs(https://github.com/getsentry/sentry-docs)
- Write starting guide for [sentry](https://github.com/getsentry/sentry)
- Add more tests
- Add more examples?


giortzisg:
We made the decision to not support issue event conversion for zap, to avoid confusion and keep things simpler. Furthermore, converting logs to issues in Go behaves weirdly when not attaching stacktraces, so all captured error logs under the same trace would be grouped under the same `errors.String` type in the Sentry UI, so the recommended way to capture issues would be using `CaptureException`. However we will still support logs for zap.

PR #1002, opened 2025-04-25
fix(otel): Prefer http route over http target for span description

### LINKED ISSUE #999: Otel seems to prefer httpTarget over httpRoute
### Problem Statement

I'm using go.opentelemetry.io/contrib/instrumentation/github.com/gin-gonic/gin/otelgin together with sentry + otel setup. otelgin seems to set span name correctly to the route `/api/v1/projects/:id/issues` but sentry otel sets it back to `/api/v1/projects/1234/issues` which causes transactions not to be grouped properly in the UI.

### Solution Brainstorm

When debugging, I see that  the code that does this is in these lines: https://github.com/getsentry/sentry-go/blob/7996759d48371bfcdcee388167c43172aa005ea9/otel/internal/utils/spanattributes.go#L104-L120

`httpTarget` takes precedence over `httpRoute`. I'm not sure what those terms exactly mean in otel context but my goal is to see grouped transactions in the UI

OrkhanAlikhanov:
Fixes #999


OrkhanAlikhanov:
@giortzisg I added more test cases

otel/internal/utils/spanattributes_test.go
```diff
@@ -0,0 +1,174 @@
+package utils_test
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.opentelemetry.io/otel/attribute"
```
giortzisg, line 9: should add otel attribute to the go.mod.

OrkhanAlikhanov: Just ran `go mod tidy` to fix it all

giortzisg (approved):
looks good

OrkhanAlikhanov:
Thank you @giortzisg. Looking forward to seeing it released

Issue #646, opened 2023-06-01
Implement EventScrubber for PII scrubbing 


tonyo:
## Summary

Implement an event scrubber and scrub events before sending them to the server when `SendDefaultPII` is set to `false` (which is the default).

## Motivation

Various event fields (request headers, request body, breadcrumbs, etc.) can be a source of PII. We should be more transparent and flexible about what data is stripped, in what cases, and how. 

The [respective RFC](https://github.com/getsentry/rfcs/blob/main/text/0062-controlling-pii-and-credentials-in-sd-ks.md) has been approved, and Python already [has a reference implementation](https://github.com/getsentry/sentry-python/blob/a48a3bbc7d06fb23444b612e64d56ec6e82f1109/sentry_sdk/scrubber.py). 

## Additional Context

The issue of PII stripping has been already brought up multiple times: https://github.com/getsentry/sentry-go/issues/248, https://github.com/getsentry/sentry-go/pull/485

Related RFC: https://github.com/getsentry/rfcs/blob/main/text/0070-document-sensitive-data-collected.md

Also related to https://github.com/getsentry/sentry-go/issues/169

asfaltboy:
While we wait for this to be implemented, would it be acceptable to expand the list of headers as per [the list of sensitive fields in the python sdk](https://github.com/getsentry/sentry-python/blob/a48a3bbc7d06fb23444b612e64d56ec6e82f1109/sentry_sdk/scrubber.py#L18-L57)? I have a PR pending, but just want to make sure it's going to get merged or not 🙏

giortzisg:
Duplicate of https://github.com/getsentry/sentry-go/issues/1241, not planned, since we're moving towards https://github.com/getsentry/sentry-go/issues/1251

PR #1008, opened 2025-05-06
Add more sensitive headers

asfaltboy:
These headers should cover a few more headers that are commonly used, with the same method.

Note: this PR does relate to https://github.com/getsentry/sentry-go/issues/646 but does not fully "fix it".

The additional headers were mostly copy pasted from [the sentry-python SDK](https://github.com/getsentry/sentry-python/blob/eee4cac8f8ae44e5bee9364a482597680b81b52f/sentry_sdk/scrubber.py#L14), and [canonicalized](https://pkg.go.dev/net/http?utm_source=godoc#CanonicalHeaderKey) just in case.


cleptric:
I'm fine with this, but please remove the Python specific comments.

viglia (approved):

interfaces.go
```diff
@@ -160,6 +160,37 @@ var sensitiveHeaders = map[string]struct{}{
...
+	"X-Csrftoken":         {},
+	"Set-Cookie":          {},
+	"X-Api-Key":           {},
+	"Aiohttp-Session":     {},
+	"Connect.sid":         {},
+	"Csrf-Token":          {},
+	"Csrf":                {},
+	"_csrf":               {},
+	"_csrf_token":         {},
+	"PHPSESSID":           {},
```
cleptric, line 188: Why would I need this in a Go app?

interfaces.go
```diff
@@ -160,6 +160,37 @@ var sensitiveHeaders = map[string]struct{}{
...
+	"_session":            {},
+	"Symfony":             {},
```
cleptric, line 190: Same question.

interfaces.go
```diff
@@ -160,6 +160,37 @@ var sensitiveHeaders = map[string]struct{}{
...
+	"Ip-Address":          {},
+	"Session":             {},
+	"Csrftoken":           {},
+	"Sessionid":           {},
+	"Remote-Addr":         {},
+	"X-Csrftoken":         {},
+	"Set-Cookie":          {},
+	"X-Api-Key":           {},
+	"Aiohttp-Session":     {},
+	"Connect.sid":         {},
```
cleptric, line 183: Isn't this Express specific?

cleptric (changes requested):

interfaces.go
asfaltboy: @cleptric let's keep one conversation here for all these comments.

The headers you're asking about were added by @antonpirker to the python SDK [here](https://github.com/getsentry/sentry-python/pull/1972), and Anton would be best suited to answer this.

If I had to guess the reason, I'd say it is because these keys can appear in any client request/response objects, e.g when we make an http request from the project that uses sentry (the http client) to another app (the http server) that can be running on anything

interfaces.go
asfaltboy: let's keep one conversation here for all these comments: https://github.com/getsentry/sentry-go/pull/1008#discussion_r2092787497

interfaces.go
asfaltboy: let's keep one conversation here for all these comments: https://github.com/getsentry/sentry-go/pull/1008#discussion_r2092787497

PR #1010, opened 2025-05-08
Add initial logger implementation

### LINKED ISSUE #1000: Sentry Structured Logging for Go
https://develop.sentry.dev/sdk/telemetry/logs/

Sentry is adding support for [structured logging](https://github.com/getsentry/sentry/discussions/86804). Let's add it to the Go SDK!

1. Define the [logs protocol](https://develop.sentry.dev/sdk/telemetry/logs/#logs-protocol) and log envelope item in the SDK
1. Add the Public API (SDK options and methods) as per the [docs](https://develop.sentry.dev/sdk/telemetry/logs/#public-api)
1. Add [SDK behavior](https://develop.sentry.dev/sdk/telemetry/logs/#sdk-behavior) as per docs
1. Attach [default attributes](https://develop.sentry.dev/sdk/telemetry/logs/#default-attributes) to the SDK as per docs
1. Instrument popular logging libraries to send logs to Sentry. For Go we should aim to support the libraries we already have docs for:
a) https://pkg.go.dev/log/slog / https://docs.sentry.io/platforms/go/guides/slog/
b) https://pkg.go.dev/github.com/sirupsen/logrus / https://docs.sentry.io/platforms/go/guides/logrus/
c) https://pkg.go.dev/github.com/rs/zerolog / https://docs.sentry.io/platforms/go/guides/zerolog/
1. Create a GH discussion that contains instructions for setting up the SDK in your repo. Eventually this will be moved into the primary docs. See the [JS SDK's GH discussion](https://github.com/getsentry/sentry-javascript/discussions/15916) for inspiration.

giortzisg:
Closes #1000.

Example usage of logger
```go
package main

import (
	"context"
	"github.com/getsentry/sentry-go/attribute"
	"time"
)

func main() {
		Dsn:        "",
		EnableLogs: true,
	})
	if err != nil {
		panic(err)
	}

	ctx := context.Background()
	logger := sentry.NewLogger(ctx)

	// you can use the logger like [fmt.Print]
	logger.Info(ctx, "Expecting ", 2, " params")
	// or like [fmt.Printf]
	logger.Infof(ctx, "format: %v", "value")

	// and you can also set attributes on the log like this
	logger.SetAttributes(
		attribute.Int("key.int", 42),
		attribute.Bool("key.boolean", true),
		attribute.Float64("key.float", 42.4),
		attribute.String("key.string", "string"),
	)
	logger.Warnf(ctx, "I have params %v and attributes", "example param")
}
```


log.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	return &noopLogger{} // fallback: does nothing
+}
+
+func (l *sentryLogger) Write(p []byte) (int, error) {
+	// Avoid sending double newlines to Sentry
+	msg := strings.TrimRight(string(p), "\n")
+	l.log(context.Background(), LevelInfo, LogSeverityInfo, msg)
+	return len(p), nil
+}
```
giortzisg, line 75: We need to implement this function to support the `io.Writer` interface, but can't really pass the context somehow. However the `Hub` is context aware and we need it to get trace & span details, so currently wrapping the custom logger with the `log` package won't be able to support binding the log to the originating span.

_examples/http/main.go
```diff
@@ -264,8 +264,8 @@ func NewImage(ctx context.Context, width, height int, seed []byte) image.Image {
 // The only reason to change logger configuration in this example is aesthetics.
 func configureLoggers() {
 	logFlags := log.Ldate | log.Ltime
-	sentry.Logger.SetPrefix("[sentry sdk]   ")
-	sentry.Logger.SetFlags(logFlags)
+	sentry.DebugLogger.SetPrefix("[sentry sdk]   ")
+	sentry.DebugLogger.SetFlags(logFlags)
```
AbhiPrasad, line 268: IMO we should extract this `DebugLogger` change to another PR to reduce the diff here. Makes the logs PR easier to review.

client.go
```diff
@@ -223,6 +223,9 @@ type ClientOptions struct {
 	MaxErrorDepth int
 	// Default event tags. These are overridden by tags set on a scope.
 	Tags map[string]string
+	// EnableLogs is a boolean flag to control if log envelopes will be generated and
+	// sent to Sentry via the logging API.
+	EnableLogs bool
```
AbhiPrasad, line 228: This should be in an experimental namespace.

Passing in a struct is probably a pain for users, so I'd be open to actually calling this `ExperimentalEnableLogs`

interfaces.go
```diff
@@ -547,3 +565,19 @@ type EventHint struct {
 	Request            *http.Request
 	Response           *http.Response
 }
+
+// Log contains information about a logging event sent to Sentry.
+type Log struct {
+	Timestamp  time.Time            `json:"timestamp,omitempty"`
+	TraceID    TraceID              `json:"trace_id,omitempty"`
+	Level      Level                `json:"level,omitempty"`
```
AbhiPrasad, line 573:
```suggestion
	Level      Level                `json:"level"`
```

sending a level is required.

interfaces.go
```diff
@@ -547,3 +565,19 @@ type EventHint struct {
...
+	Severity   int                  `json:"severity_number"`
```
AbhiPrasad, line 574:
```suggestion
	Severity   int                  `json:"severity_number,omitempty"`
```

`severity_number` is optional, it can be ignored.

log.go
```diff
@@ -0,0 +1,212 @@
...
+// sentryLogger implements a custom logger that writes to Sentry.
+type sentryLogger struct {
+	client  *Client
+	options LoggerOptions
+}
+
+type LoggerOptions struct {
+	// BeforeSendLog is called before a log event is sent to Sentry.
+	// Use it to mutate the log event or return nil to discard.
+	BeforeSendLog func(log *Log) *Log
```
AbhiPrasad, line 52: This should be a client option - we can also give it an experimental prefix.

log.go
```diff
@@ -0,0 +1,212 @@
...
+}
+func (l *sentryLogger) Warn(ctx context.Context, v ...interface{}) {
+	l.log(ctx, LogLevelWarning, LogSeverityWarning, v...)
+}
+func (l *sentryLogger) Error(ctx context.Context, v ...interface{}) {
+	l.log(ctx, LogLevelError, LogSeverityError, v...)
+}
+func (l *sentryLogger) Fatal(ctx context.Context, v ...interface{}) {
+	l.log(ctx, LogLevelFatal, LogSeverityFatal, v...)
+	os.Exit(1)
```
AbhiPrasad, line 191: why do we exit here? I don't think users would expect their 3rd party logging apis would exit for them.

log.go
```diff
@@ -0,0 +1,212 @@
...
+}
+func (l *sentryLogger) Panic(ctx context.Context, v ...interface{}) {
+	l.log(ctx, LogLevelFatal, LogSeverityFatal, v...)
+	panic(fmt.Sprint(v...))
```
AbhiPrasad, line 195: This is fair, but I'd rather we don't implement this atm, and add it after we get user requests for it. Library code calling `panic` is a pattern I'm not super comfortable with.

log.go
AbhiPrasad: shoutout a classic go thread: https://github.com/golang/go/issues/20280

you're right that the span id might not be accurate in this scenario, so perhaps we need to just not attach it (and only attach the current trace). This means that leveraging the Writer interface means you can't associate spans directly to logs - only traces.

We can document this as a limitation for now.

AbhiPrasad (commented):
Saw this was in progress so did a preliminary review! Do you mind also adding some examples in the PR description of how log should be configured and used with this PR?

client.go
aldy505: The Go SDK has experienced this kind of breaking change with `ProfilesSampleRate`. I guess `EnableLogs` is fine already.

log.go
aldy505: It's actually common in Go that `Fatal()` log method actually executes `os.Exit(1)`. It's on their stdlib anyway.

batch_logger.go
```diff
@@ -0,0 +1,94 @@
...
+		ctx, cancel := context.WithCancel(context.Background())
+		l.cancel = cancel
+		l.wg.Add(1)
+		go l.run(ctx)
+	})
+}
+
+func (l *BatchLogger) Flush() {
+	if l.cancel != nil {
+		l.cancel()
```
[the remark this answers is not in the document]
giortzisg: The batchlogger is already running in the background, so cancel passes a cancel context for the go routine to stop. If you then check the logging logic, I already implemented draining when cancelling, so it's possible to receive a log even after we flushed, so should be ok.

log.go
giortzisg: I'd argue that implementing Panic and Fatal is about giving a complete library to the users. All log libraries have Panic & Fatal, and if the user calls Panic it's expected for the sdk to panic, so it would be a user problem if they use it incorrectly.

giortzisg:
Thanks for the review Ahhh. Will also add examples, nice catch.

client.go
cleptric: Let's keep it as is. Given that we will be in EA soon, I want to avoid removing options a few weeks later.

interfaces.go
```diff
@@ -13,13 +13,16 @@ import (
...
+// eventType is the type of error event.
 const eventType = "event"
 
-// transactionType is the type of a transaction event.
+// transactionType is the type of transaction event.
 const transactionType = "transaction"
 
-// checkInType is the type of a check in event.
+// logType is the type of log event.
+const logType = "log"
```
cleptric, line 23: For later, but let's maybe turn this into a struct, with the item type and its content type.

log.go
cleptric: Signature should be `BeforeSendLog func(event *Event, hint *EventHint) *Event`.

log.go
cleptric: We can keep it as is, see https://github.com/golang/go/blob/c57b0d6c7e54284c8967530153774dbc62190b00/src/log/log.go#L305-L311

log.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+	}
+
+	// if log was called with SetAttributes pass the attributes to attrs
+	if len(l.attributes) > 0 {
+		for k, v := range l.attributes {
+			attrs[k] = v
+		}
+		// flush attributes from logger after send
+		clear(l.attributes)
```
giortzisg, line 111: Not sure if attributes should persist or not.

cleptric:
For `Debug: true`:
```
Sending info event [a710bb92af79490b9135782c8924d613] to o1.ingest.us.sentry.io project:
```
instead we should emit something like
```
Sending X log events [a710bb92af79490b9135782c8924d613] to o1.ingest.us.sentry.io project:
```

log.go
```diff
@@ -0,0 +1,265 @@
...
+
+	// handle metadata
+	if release := l.client.options.Release; release != "" {
+		attrs["sentry.release"] = Attribute{Value: release, Type: "string"}
+	}
+	if environment := l.client.options.Environment; environment != "" {
+		attrs["sentry.environment"] = Attribute{Value: environment, Type: "string"}
+	}
+	if serverAddr := l.client.options.ServerName; serverAddr != "" {
+		attrs["sentry.server.address"] = Attribute{Value: serverAddr, Type: "string"}
```
cleptric, line 122: Let's use `os.Hostname()` if `client.options.ServerName` is `""`.

client.go
```diff
@@ -223,6 +226,9 @@ type ClientOptions struct {
...
```
cleptric, line 231:
```suggestion
	// Enable logs being emitted.
	EnableLogs bool
```

client.go
```diff
@@ -617,17 +632,21 @@ func (client *Client) processEvent(event *Event, hint *EventHint, scope EventMod
...
+				DebugLogger.Println("Transaction dropped due to BeforeSendTransaction callback.")
+				return nil
+			}
 		}
-	} else if event.Type != transactionType && event.Type != checkInType && client.options.BeforeSend != nil {
-		// All other events
-		if event = client.options.BeforeSend(event, hint); event == nil {
-			DebugLogger.Println("Event dropped due to BeforeSend callback.")
-			return nil
+	case checkInType:
```
cleptric, line 643: Do we need this case?

interfaces.go
```diff
@@ -11,15 +11,20 @@ import (
 	"slices"
 	"strings"
 	"time"
+
+	"github.com/getsentry/sentry-go/attribute"
 )
 
-// eventType is the type of an error event.
+// eventType is the type of error event.
```
cleptric, line 18:
```suggestion
// eventType is the default event type.
```

Messages are not errors.

interfaces.go
```diff
@@ -11,15 +11,20 @@ import (
...
 const eventType = "event"
 
-// transactionType is the type of a transaction event.
+// transactionType is the type of transaction event.
```
cleptric, line 21:
```suggestion
// transactionType is the type of an transaction event.
```

interfaces.go
```diff
@@ -11,15 +11,20 @@ import (
...
 const transactionType = "transaction"
 
-// checkInType is the type of a check in event.
+// logType is the type of log event.
```
cleptric, line 24:
```suggestion
// logType is the type of an log event.
```

interfaces.go
```diff
@@ -11,15 +11,20 @@ import (
...
+const logType = "log"
+
+// checkInType is the type of check in event.
```
cleptric, line 27:
```suggestion
// checkInType is the type of an check in event.
```

interfaces.go
```diff
@@ -34,7 +39,7 @@ const (
 	LevelFatal   Level = "fatal"
 )
 
-// SdkInfo contains all metadata about about the SDK being used.
+// SdkInfo contains all metadata about the SDK being used.
```
cleptric, line 42:
```suggestion
// SdkInfo contains all metadata about the SDK.
```

interfaces.go
```diff
@@ -96,6 +101,61 @@ func (b *Breadcrumb) MarshalJSON() ([]byte, error) {
 	return json.Marshal((*breadcrumb)(b))
 }
 
+// Logger allows sending structured logs to sentry.
```
cleptric, line 104:
```suggestion
```

log.go
```diff
@@ -0,0 +1,218 @@
...
+	attribute.STRING:  "string",
+}
+
+// sentryLogger implements a custom logger that writes to Sentry.
+type sentryLogger struct {
+	client     *Client
+	attributes map[string]Attribute
+}
+
+// NewLogger returns a Logger that writes to Sentry if enabled, or discards otherwise.
```
cleptric, line 48:
```suggestion
// NewLogger returns a Logger that emits logs to Sentry. If logging is turned off, all logs get discarded.
```

transport.go
```diff
@@ -627,16 +649,18 @@ func (t *HTTPSyncTransport) SendEventWithContext(ctx context.Context, event *Eve
 		return
 	}
 
-	var eventType string
+	var eventIdentifier string
```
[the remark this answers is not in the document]
giortzisg: it's not only a type, since for log events you also have the number of events sent.

client.go
giortzisg: yes, we don't want to go to the default option.

log.go
cleptric: We can leave it as is for now.

cleptric (approved):
Once https://github.com/getsentry/sentry-go/pull/1010/files/d40ff0a677f0d3d3b54ba2a0ae594ce85837d15d#r2086096970 is addressed, we can get this merged :shipit:

Issue #991, opened 2025-04-08
Improve getting & setting data on spans and transactions


giortzisg:
### Problem Statement

Currently, if I want to get attribute data from an existing span or transaction I need to check the existance and also typecast into the final value.
```go
if d, found := transaction.Data["dataAttr1"]; found {
  if dataAttr1, ok := d.(int); ok {
    transaction.SetData("dataAttr1", dataAttr1.(int)+42)
  }
}
```
For getting multiple values, the code footprint would become fairly large. It would be nicer to be able to handle getting and casting values with a `GetData` function in the transaction/span layer.

### Solution Brainstorm

Not sure about the better way of doing that, but here are two proposals.
- Generic helper function:
```go 
// signature
func GetData[T any](s *Span, name string) (T, error)

// usage 
val, err := sentry.GetData[int](transaction, "int")
if err != nil {
// handle error
}
transaction.SetData("int", val+42)
 ```

pros: small footprint,  less maintenance, custom data structures
cons: casting on runtime (not type safe), how to handle complex cases?

- Custom datastructure and types
not sure on exact implementation but something like [otel](https://github.com/open-telemetry/opentelemetry-go/blob/v1.35.0/attribute/value.go#L22) `Value` type? 
``` go
type Value struct {
	vtype    Type
	numeric  uint64
	stringly string
	slice    interface{}
}
```
pros: concrete data types (no interface{}), type safe
cons: bigger footprint, complex logic, more maintenance, no support for custom data types

giortzisg:
This is already implemented as part of the global attributes API here https://github.com/getsentry/sentry-go/pull/1007

PR #1019, opened 2025-05-29
feat(tracing): get span data

qepo17:
- GetSpanData to get data from a span using a generic approach
- New TestGetSpanData to guard the function from various scenarios

Why Generic? 
- I usually do not store complex data in the span, hence this approach is good with less footprint and maintenance burden compared to Otel's approach mentioned [in the issue](https://github.com/getsentry/sentry-go/issues/991)

### DIFF
--- a/tracing.go
+++ b/tracing.go
@@ -5,7 +5,6 @@ import (
 	"crypto/rand"
 	"encoding/hex"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"net/http"
 	"regexp"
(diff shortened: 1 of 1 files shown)


giortzisg:
Hey @qepo17, thanks for the contribution. However we are looking to migrate away from `map[string]interface{}` on span Data, since Relay can't handle all generic types and we want to enforce only what we support. That's why we've went with the otel `attribute` package in the end as you can see [here](https://github.com/getsentry/sentry-go/blob/master/attribute/value.go). 

To avoid confusion and to not waste your time I would suggest commenting or opening issues and talking to us first before opening pull requests.

qepo17:
Understood, thanks for the info

Issue #1029, opened 2025-06-05
Structured logging support for zerolog


stephanie-anderson:
With the addition of structured logging for sentry with https://github.com/getsentry/sentry-go/pull/1010 we also want to also instrument the logging libraries that we already support:

https://pkg.go.dev/github.com/rs/zerolog / https://docs.sentry.io/platforms/go/guides/zerolog/

AbhiPrasad:
This is even the case when someone uses the context integration? https://github.com/rs/zerolog?tab=readme-ov-file#contextcontext-integration

giortzisg:
Yes, the context passed on the logger is available only for Hooks.

A solution that might work would be to wrap the zerolog logger with our own implementation that matches theirs and extend it with our own `WithCtx` that injects `ctx` to our own sentry logger.

AbhiPrasad:
> A solution that might work would be to wrap the zerolog logger with our own implementation that matches theirs and extend it with our own WithCtx that injects ctx to our own sentry logger.

yeah that sounds good. Let's try experimenting with that. Let's do that with an initial PR before implementing logs.

AbhiPrasad:
could we make the wrapping work if we used something like https://github.com/DataDog/orchestrion?

PR #1031, opened 2025-06-08
feat: structured logging support for zerolog

aldy505:
Closes https://github.com/getsentry/sentry-go/issues/1029

### DIFF
--- a/zerolog/sentrylogger.go
+++ b/zerolog/sentrylogger.go
@@ -1,0 +1,129 @@
+package sentryzerolog
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"time"
+
+	"github.com/getsentry/sentry-go"
+	"github.com/getsentry/sentry-go/attribute"
+	"github.com/rs/zerolog"
+)
+
+// NewSentryLogger creates a new instance of SentryLogger that implements
+// zerolog.LevelWriter interface. This should be used for sending zerolog
+// events to Sentry as logs as opposed to events/errors.
+//
+// If you want to send events/errors to Sentry, use the New() function instead.
+func NewSentryLogger() SentryLogger {
+	return SentryLogger{}
+}
+
+// SentryLogger implements zerolog.LevelWriter.
+// This should be used for sending zerolog events to Sentry as logs as opposed
+// to events/errors.
+type SentryLogger struct{}
+
+var _ zerolog.LevelWriter = (*SentryLogger)(nil)
+var _ io.Closer = (*SentryLogger)(nil)
+
+func (s SentryLogger) Write(p []byte) (n int, err error) {
+	return s.WriteLevel(zerolog.DebugLevel, p)
+}
+
+func (s SentryLogger) WriteLevel(level zerolog.Level, p []byte) (n int, err error) {
+	return s.runContext(context.Background(), level, p)
+}
+
+func (s SentryLogger) runContext(ctx context.Context, level zerolog.Level, p []byte) (n int, err error) {
+	if !sentry.HasHubOnContext(ctx) {
+		hub := sentry.CurrentHub()
+		if hub == nil {
+			hub = sentry.NewHub(nil, sentry.NewScope())
+		}
+
+		ctx = sentry.SetHubOnContext(context.Background(), hub)
+	}
+
+	var evt map[string]any
+	d := json.NewDecoder(bytes.NewReader(p))
+	err = d.Decode(&evt)
+	if err != nil {
+		return 0, fmt.Errorf("cannot decode event: %s", err.Error())
+	}
+
+	l := sentry.NewLogger(ctx)
+	var message string
+	for field, value := range evt {
+		switch field {
+		case zerolog.LevelFieldName:
+			levelString, _ := value.(string)
+			level, err = zerolog.ParseLevel(levelString)
+			if err != nil {
+				level = zerolog.DebugLevel
+			}
+		case zerolog.MessageFieldName:
+			message, _ = value.(string)
+		case zerolog.TimestampFieldName:
+			continue
+		default:
+			switch valueType := value.(type) {
+			case string:
+				l.SetAttributes(attribute.String(field, valueType))
+			case bool:
+				l.SetAttributes(attribute.Bool(field, valueType))
+			case float64:
+				l.SetAttributes(attribute.Float64(field, float64(valueType)))
+			case []any:
+				for i, v := range valueType {
+					switch vv := v.(type) {
+					case string:
+						l.SetAttributes(attribute.String(fmt.Sprintf("%s.%d", field, i), vv))
+					case bool:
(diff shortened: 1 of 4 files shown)


zerolog/sentrylogger_test.go
```diff
@@ -0,0 +1,642 @@
...
+			},
+		},
+		{
+			name: "String array attribute",
+			logFunc: func(c *zerolog.Event) *zerolog.Event {
+				return c.Strs("foo", []string{"bar", "baz"})
+			},
+			wantAttributes: map[string]sentry.Attribute{
+				"foo.0": {Value: "bar", Type: "string"},
+				"foo.1": {Value: "baz", Type: "string"},
```
aldy505, line 373: I'm not sure if this is the right behavior. If we see into https://github.com/open-telemetry/opentelemetry-specification/issues/3982, they said if we have an array attribute, it shouldn't be separated like this. But I don't really know the internals of the Trace Explorer.

The options would be:
1. `foo.0` => `bar`, `foo.1` => `baz`
2. `foo` => `["bar", "baz"]`

Perhaps @AbhiPrasad can help with this?

AbhiPrasad: Let's please not unfurl array properties into their own keys. Let's send them as a string for now, with the intention that in the future sentry ingestion will support arrays.

We do this unfurling pattern for message parameters (`sentry.message.parameter.0`) because we needed to make them indexed + queryable immediately for better product UX in the short term, but I want to avoid arbitrary user arrays to be done this way.

zerolog/sentrylogger.go
```diff
@@ -0,0 +1,129 @@
...
+			level, err = zerolog.ParseLevel(levelString)
+			if err != nil {
+				level = zerolog.DebugLevel
+			}
+		case zerolog.MessageFieldName:
+			message, _ = value.(string)
+		case zerolog.TimestampFieldName:
+			continue
+		default:
+			switch valueType := value.(type) {
```
AbhiPrasad, line 73: shouldn't we have a case for `int` here?

zerolog/sentrylogger.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// Zerolog requires that the original number of bytes is returned.
+	// Otherwise, it will return "short write" error.
+	return len(p), nil
+}
+
+// Close should not be called directly.
+// It should be called internally by zerolog.
+func (s SentryLogger) Close() error {
+	sentry.Flush(time.Second)
```
AbhiPrasad, line 127: should we make this configurable?

zerolog/sentrylogger.go
```diff
@@ -0,0 +1,129 @@
...
+			continue
+		default:
+			switch valueType := value.(type) {
+			case string:
+				l.SetAttributes(attribute.String(field, valueType))
+			case bool:
+				l.SetAttributes(attribute.Bool(field, valueType))
+			case float64:
+				l.SetAttributes(attribute.Float64(field, float64(valueType)))
+			case []any:
```
AbhiPrasad, line 80: we should have a default case here that just serializes the value and sends as a string type.

zerolog/sentrylogger.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func (s SentryLogger) runContext(ctx context.Context, level zerolog.Level, p []byte) (n int, err error) {
+	if !sentry.HasHubOnContext(ctx) {
+		hub := sentry.CurrentHub()
+		if hub == nil {
+			hub = sentry.NewHub(nil, sentry.NewScope())
+		}
+
+		ctx = sentry.SetHubOnContext(context.Background(), hub)
```
AbhiPrasad, line 48:
```suggestion
		ctx = sentry.SetHubOnContext(ctx, hub)
```

zerolog/sentrylogger.go
```diff
@@ -0,0 +1,129 @@
...
+	err = d.Decode(&evt)
+	if err != nil {
+		return 0, fmt.Errorf("cannot decode event: %s", err.Error())
+	}
+
+	l := sentry.NewLogger(ctx)
+	var message string
+	for field, value := range evt {
+		switch field {
+		case zerolog.LevelFieldName:
```
AbhiPrasad, line 62: these names can be customized: https://github.com/rs/zerolog/blob/master/README.md#customize-automatic-field-names

we prob need a helper

```go
func (s SentryLogger) getFieldNames() (level, message, timestamp string) {
	return zerolog.LevelFieldName,
		zerolog.MessageFieldName,
		zerolog.TimestampFieldName
}
```

zerolog/sentrylogger.go
aldy505: I did test it with every possible Go types. Yet everything falls down to just 3 types (plus their array variant): string, boolean, float64

aldy505:
> Instead of using a writer, can we use the hook system in zerolog? https://github.com/rs/zerolog/blob/master/README.md#hooks
> 
> I'd like us to avoid the overhead of encoding/decoding json if we can.
> 
> something like
> 
> ```go
> type SentryLogHook struct{}
> 
> func (h SentryLogHook) Run(e *zerolog.Event, level zerolog.Level, msg string) {
>     ctx := e.GetCtx()
>     l := sentry.NewLogger(ctx)
>     // need more complicated logic to parse the zerolog event
>     l.WithLevel(convertLevel(level)).Log(ctx, msg)
> }
> 
> logger := zerolog.New(os.Stdout).Hook(SentryLogHook{})
> ```

@AbhiPrasad I've tried the Hook, didn't work. The Hook has this interface signature:
```go
Run(e *Event, level Level, message string)
```
In which:
* `e *Event` is write only, you cannot read the properties/attributes already set to that event
* `level Level` is useful yes, but for level only
* `message string` only contains the message, you can't acquire the attributes here

> If we do have to use a writer, we should make it async.

Yes we might want to go to this direction.

zerolog/sentrylogger_test.go
aldy505: Good to know. Thanks Abhi!

zerolog/sentrylogger.go
aldy505: Perhaps, yes.

zerolog/sentrylogger.go
aldy505: No can't do. You must assume that the previous `ctx` is `nil`.

cleptric:
Can we please discuss implementation prior opening PRs @aldy505. This is really unproductive else.

zerolog/sentrylogger.go
AbhiPrasad: We'll want to still explicitly use `attribute.Int` because we treat these differently in the backend. In the future there will be operations you can only do on integers you can can't use on floats.

zerolog/sentrylogger.go
AbhiPrasad: really? do you have links to code/docs about this?

PR #1033, opened 2025-06-10
Add slog logs integration

### LINKED ISSUE #1027: Structured logging support for slog
With the addition of structured logging for sentry with https://github.com/getsentry/sentry-go/pull/1010 we also want to also instrument the logging libraries that we already support:

https://pkg.go.dev/log/slog / https://docs.sentry.io/platforms/go/guides/slog/

giortzisg:
### Description 

Closes #1027. 
Changing the current `sentryslog` implementation to add support for creating sentry Logs, using different option Levels:
```go
        ctx := context.Background()
	logger := slog.New(sentryslog.Option{
		LogLevel:   []slog.Level{slog.LevelWarn, slog.LevelInfo},
		EventLevel: []slog.Level{slog.LevelError, sentryslog.LevelFatal},
	}.NewSentryHandler(ctx))
	logger = logger.With("release", "v1.0.0")

	// message level is Error and will be handled by both the event & log handler.
	logger.
		With(
			slog.Group("user",
				slog.String("id", "user-123"),
				slog.Time("created_at", time.Now()),
			),
		).
		With("environment", "dev").
		With("error", fmt.Errorf("an error")).
		Error("a message")
```
Somes notes:
- It would be good to add `context.Context` while creating the `NewSentryHandler`, but that would be a breaking change.


slog/sentryslog.go
```diff
@@ -69,9 +78,17 @@ type Option struct {
...
+	if o.EventLevel == nil {
+		// backwards compatibility
+		if o.Level != nil {
+			o.EventLevel = o.Level
+		} else {
+			o.EventLevel = slog.LevelError
+		}
+	}
+	if o.LogLevel == nil {
+		o.LogLevel = slog.LevelDebug
```
giortzisg: Based on defaults, if a user upgrades, log messages will be sent. Should we handle the default level differently for now?

slog/converter.go
```diff
@@ -121,3 +125,45 @@ func handleRequestAttributes(v slog.Value, event *sentry.Event) {
 		}
 	}
 }
+
+func attrToSentryLog(group string, a slog.Attr) []attribute.Builder {
+	switch a.Value.Kind() {
+	case slog.KindAny:
+		return []attribute.Builder{attribute.String(group+a.Key, fmt.Sprintf("%+v", a.Value.Any()))}
```
AbhiPrasad: l: Let's move `group+a.Key` to the top `key := group + a.Key` and pass in `key` everywhere.

log.go
```diff
@@ -144,7 +144,9 @@ func (l *sentryLogger) log(ctx context.Context, level LogLevel, severity int, me
 	if sdkVersion := l.client.sdkVersion; sdkVersion != "" {
 		attrs["sentry.sdk.version"] = Attribute{Value: sdkVersion, Type: "string"}
 	}
-	attrs["sentry.origin"] = Attribute{Value: "auto.logger.log", Type: "string"}
+	if _, ok := attrs["sentry.origin"]; !ok {
+		attrs["sentry.origin"] = Attribute{Value: "auto.logger.log", Type: "string"}
```
AbhiPrasad: h: Origin should not be set by default, only for logs generated by logging integrations.

This is because explicit calls to the logging API are assumed to be `manual`, so no applicable origin.

slog/sentryslog.go
```diff
@@ -42,9 +43,17 @@ var (
 )
 
 type Option struct {
-	// Level sets the minimum log level to capture and send to Sentry.
-	// Logs at this level and above will be processed. The default level is debug.
+	// Deprecated: Level is kept for backwards compatibility and defaults to EventLevel.
```
AbhiPrasad: If we are going to deprecate, we need to provide the alternative (don't use it, use X instead, etc.).

It would also be nice to throw deprecations in https://github.com/getsentry/sentry-go/blob/master/MIGRATION.md so that they can be easily searched for and linked in the changelog, but I'll leave the exact process up to you.

slog/sentryslog.go
AbhiPrasad: It's still gated behind `EnableLogs` right? I think thats fine.

slog/sentryslog.go
```diff
@@ -82,24 +99,80 @@ func (o Option) NewSentryHandler() slog.Handler {
...
+func (h *SentryHandler) Enabled(ctx context.Context, level slog.Level) bool {
+	return h.eventHandler.Enabled(ctx, level) || h.logHandler.Enabled(ctx, level)
+}
+
+func (h *SentryHandler) Handle(ctx context.Context, record slog.Record) error {
+	var err error
+
+	if h.eventHandler.Enabled(ctx, record.Level) {
+		if handleErr := h.eventHandler.Handle(ctx, record); handleErr != nil {
+			err = handleErr
```
AbhiPrasad: m: shouldn't we return the first error that we get? Otherwise I'd prefer if we construct an error that is the combo of of both if we hit both `eventHandler` and `logHandler` errors.

slog/sentryslog.go
```diff
@@ -82,24 +99,80 @@ func (o Option) NewSentryHandler() slog.Handler {
 		o.AttrFromContext = []func(ctx context.Context) []slog.Attr{}
 	}
 
-	return &SentryHandler{
+	logger := sentry.NewLogger(ctx)
```
AbhiPrasad, line 106: I wonder if this will lead to stale contexts.

Should we just create a new logger everytime on `Handle`? Overhead might be more, so perhaps we can just set new contexts if it changes?

slog/converter.go
```diff
@@ -121,3 +125,45 @@ func handleRequestAttributes(v slog.Value, event *sentry.Event) {
...
+		return []attribute.Builder{attribute.String(group+a.Key, a.Value.Time().Format(time.RFC3339))}
+	case slog.KindUint64:
+		val := a.Value.Uint64()
+		if val <= math.MaxInt64 {
+			return []attribute.Builder{attribute.Int64(group+a.Key, int64(val))}
+		} else {
+			return []attribute.Builder{attribute.String(group+a.Key, strconv.FormatUint(val, 10))}
+		}
+	case slog.KindLogValuer:
+		return []attribute.Builder{attribute.String(group+a.Key, a.Value.LogValuer().LogValue().String())}
```
AbhiPrasad: This risks infinite recursion right? I wonder if we need to just special case this somehow.

slog/converter.go
```diff
@@ -121,3 +125,45 @@ func handleRequestAttributes(v slog.Value, event *sentry.Event) {
...
+	case slog.KindGroup:
+		// Handle nested group attributes
+		var attrs []attribute.Builder
```
AbhiPrasad, line 165: I wonder if we can estimate a pre-allocation here. Maybe too difficult.

slog/sentryslog.go
giortzisg: Yeap, still gated behind it.

slog/sentryslog.go
giortzisg: We need to handle both events and logs so can't return on first error. But it would be probably nicer to wrap them.

slog/sentryslog.go
giortzisg: We are already using the `ctx` passed on sentryLogger.log to update the hub if possible, so we don't need to create a new logger.

slog/converter.go
giortzisg: Was the comment meant for `slog.KindGroup`? Valuer should be fine.

slog/converter.go
giortzisg: I'd say that we can't really know how many sub groups we have without checking first. Also performance should be fine.

slog/sentryslog.go
AbhiPrasad: yeah, let's wrap them.

slog/sentryslog.go
```diff
@@ -42,9 +44,17 @@ var (
...
 	Level slog.Leveler
+	// EventLevel sets the minimum log level to capture and send to Sentry as an Event.
+	// Logs at this level and above will be processed as events.
+	// Defaults to slog.LevelError.
+	EventLevel slog.Leveler
+
+	// LogLevel sets the minimum log level to capture and send to Sentry as a Log entry.
+	// Logs at this level and above will be processed as log entries.
+	// Defaults to slog.LevelDebug.
+	LogLevel slog.Leveler
```
AbhiPrasad: I know I said previously that `EnableLogs` gating this is fine, but I tried running this on a local go app and now I'm now debating if there should be a way to turn this off.

I can see users wanting to disable `EventLevel` completely and only rely on `LogLevel` for example.

Maybe we expose separate boolean options? What do you think?

slog/converter.go
AbhiPrasad: Yes sorry - `slog.KindGroup`.

giortzisg: Probably can, but not while using `WithGroup`.  You can possibly create : 
```go
attr1 := slog.Attr{
	Key:   "group1",
	Value: slog.GroupValue(),
}

attr2 := slog.Attr{
	Key:   "group2",
	Value: slog.GroupValue(attr1),
}
```
and then create another group reference to the second one `attr1.Value = slog.GroupValue(attr2)`. Should we guard against that? Seems like a pretty wacky case to me.

slog/sentryslog.go
giortzisg: Don't know if i like adding also boolean options. 

It might be nicer to specify an array of log levels so the user directly specifies all the levels that wants to be captured. This also provides more fine grained control, since maybe you don't want to catch `Error` level for logs and want only events.

But then I don't know how we can handle and compare custom slog levels.

EDIT: we actually don't support them with the current implementation, so let me change that as well.

AbhiPrasad: I like the array suggestion.

slog/converter.go
giortzisg: Delved a bit deeper into it, slog.Attr and slog.Groups are values, so we can't really have a cycle. If a user manually creates a `slog.Attr` with its' `Value` being a pointer to a GroupValue it should be handled by `slog.KindAny` not `slog.KindGroup`.

slog/README.MD
```diff
@@ -46,24 +46,26 @@ func main() {
...
 
 ## Configuration
 
 The slog-sentry package offers several options to customize how logs are handled and sent to Sentry. These are specified through the Option struct:
 
-- `EventLevel`: Minimum level to send logs as events to Sentry. Defaults to `slog.LevelError`.
+- `EventLevel`: Slice of specific levels to send `Events` to Sentry. Defaults to `[]slog.Level{slog.LevelError, LevelFatal}`.
 
-- `LogLevel`: Minimum level to send logs to Sentry. Defaults to `slog.LevelDebug`.
+- `LogLevel`: Slice of specific levels to send `Log` entries to Sentry. Defaults to `[]slog.Level{slog.LevelError, LevelFatal}`.
```
AbhiPrasad: Is this right? From below I can see `o.LogLevel = []slog.Level{slog.LevelDebug, slog.LevelInfo, slog.LevelWarn, slog.LevelError, LevelFatal}`

AbhiPrasad (approved):
One note on the README, otherwise we are good

slog/README.MD
giortzisg: Yes that was missed. Nice catch Abhi!

PR #1035, opened 2025-06-12
fix: data race issue occurring when multiple goroutines concurrently write to hub

japananh:


giortzisg:
Closing this since #1050 should fix the issue

PR #1039, opened 2025-06-17
Add slog fingerprint handler

golemiso:
This pull request introduces support for handling the fingerprint attribute in the slog-to-Sentry event conversion process.

- Added logic in attrToSentryEvent to recognize the "fingerprint" key and assign its value to the Sentry event's Fingerprint field.
- Implemented the handleFingerprint function to process the attribute when it is a []string.
- Added corresponding test cases to ensure correct behavior.

This enhancement allows users to specify custom fingerprints via slog attributes, enabling more flexible event grouping in Sentry.


giortzisg (approved):
Thanks for the contribution @golemiso

PR #1044, opened 2025-06-25
Fix race condition on GetSpan and guard transport Close

giortzisg:
Partial fix for #1047.

## Description

- Fixing a race condition on GetSpan by adding a mutex.
- Guarding transport so that you can only close the `done` channel once, to avoid panics.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -303,7 +303,7 @@ type HTTPTransport struct {
 	// current in-flight items and starts a new batch for subsequent events.
 	buffer chan batch
 
-	start     sync.Once
+	startOnce sync.Once
 	closeOnce sync.Once
 
 	// Size of the transport buffer. Defaults to 30.
@@ -365,7 +365,7 @@ func (t *HTTPTransport) Configure(options ClientOptions) {
 		}
 	}
 
-	t.start.Do(func() {
+	t.startOnce.Do(func() {
 		go t.worker()
 	})
 }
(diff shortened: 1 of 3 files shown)


cleptric:
@sentry review

transport_test.go
```diff
@@ -490,6 +490,28 @@ func TestHTTPTransport(t *testing.T) {
...
+	for i := 0; i < 10; i++ {
+		transport.Close()
+	}
+
+	// Verify the done channel is closed
+	select {
+	case <-transport.done:
+		// Expected - channel should be closed
+	case <-time.After(time.Second):
+		t.Fatal("tranposrt.done not closed")
```
lcian:
```suggestion
		t.Fatal("transport.done not closed")
```

transport.go
```diff
@@ -303,7 +303,8 @@ type HTTPTransport struct {
 	// current in-flight items and starts a new batch for subsequent events.
 	buffer chan batch
 
-	start sync.Once
+	start     sync.Once
```
lcian: `l`: is naming things `<something>Once` a convention in go? Then maybe we could rename this to `startOnce` also for consistency between the two (could also be done in a separate PR though)

lcian (approved):

transport.go
giortzisg: kinda is. Makes sense will change for consistency.

PR #1045, opened 2025-06-25
In Event.SetException support errors implementing Unwrap() []error

kamstrup:
# Background

At some point in time Go grew types of errors that do not support the standard `errors.Unwrap()` function. Notably `errors.Join(...)` and `fmt.Errorf("%w: multiple errors here %w", err1, err2)`. These errors expose an interface `Unwrap() []error`. This is also described in the Go documentation: https://pkg.go.dev/errors#Unwrap

This new Unwrap interface is not supported in the Sentry client, leading to not picking up stacktraces correctly for errors that have been wrapped this way.

# Description
In this PR I update `Event.SetException` to support these new "multi-errors".

It is done by implementing a Go-1.23-style iterator `unwrapAll(err)`. The code does currently not use range-over-func so requires no bump in Go dependency; but is ready for that when the language level is bumped one day.

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -3,7 +3,6 @@ package sentry
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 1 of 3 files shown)


giortzisg:
Hey @kamstrup, thanks for the contribution. From a quick look I had, this breaks issue grouping, although maybe it's unavoidable.

giortzisg:
Also if we want to support this, we need to properly handle Exception Groups ([RFC](https://github.com/getsentry/rfcs/blob/32ac291a543751e457d36d405106ae4e1cc6e09b/text/0079-exception-groups.md)) and build an error tree for the wrapped errors.

giortzisg:
> My main motivation was just to "get stack traces back in our Sentry issues", because we lost them on around ~50% of the generated issues after developers started using `errors.Join()` and `fmt.Errorf()` with multiple `%w`s.

On my testing I could only see stack traces only on the last exception.

> I would expect that the full fledged solution to this issue would take a good while to develop - in which case it might be worthwhile considering some stop-gap solution for any Sentry users with similar problems.

The problem is, the way the partial solution is currently implemented, it links every error as inherited from the previous one, and results in an incorrect error tree on the UI, so I wouldn't want to introduce it in this state.

giortzisg:
Will be closing this then, except if you have anything else to add @kamstrup. But we will be having a look to properly support Exception Groups in the future.

PR #1046, opened 2025-06-27
Modify log API

### LINKED ISSUE #1042: Modify sentryLogger API signature
## Problem

The current api and the use of `setAttributes` on the logger is confusing, since users would expect for the attributes to be permanently attached to the logger. Since we want to support both attributes and format parameters, we need to use chaining for our temporary attributes, and the `setAttributes` func would work as expected, permanently attaching the attributes on the logger instance.

## Solution

The API should be clearer and temporary attributes should be attached on the log entry itself. The proposed solution then should be:

* logger.Info() -> log entry with specified level
* logEntry.String("key", "value") -> log entry with temporary attributes
* logEnty.Msg() -> invoke sentrylogger to send the message.

Then an example of usage would be: 

```
logger := sentry.NewLogger()
logger.SetAttributes(attribute.String("key.permanent", "value")

logger.Info().String("key.temp", "value2").Msg()
```

The API should support all currently supported values on the `attribute` package.

giortzisg:
Closes #1042. See issue for more details.

## Description
Modifying existing API to make usage clearer. Introducing chaining of attributes:
- Multiple levels `Trace`, `Debug`, `Info`, `Warn`, `Error`, `Panic`, `Fatal` that return a `LogEntry`
- Attribute API on each `LogEntry`, to set `String`, `Int`, ...
- WithCtx func to inject context to entry
- Emit & Emitf functions to create a message and send the log

 Example:
```go
loggerWithVersion := sentry.NewLogger(ctx)

// Setting  attributes (these persist across log calls)
loggerWithVersion.SetAttributes(
    attribute.String("permanent.version", "1.0.0"),
)

// Add attributes to log entry
loggerWithVersion.Info().WithCtx(ctx).
    String("key.string", "value").
    Int("key.int", 42).
    Bool("key.bool", true).
    // don't forget to call Emit to send the logs to Sentry
    Emitf("Message with parameters %d and %d", 1, 2)

logger := sentry.NewLogger(ctx)

// you can also use Attributes for setting many attributes in the log entry (these are temporary)
logger.Info().
    Attributes(
	attribute.String("key.temp.string", "value"),
	attribute.Int("key.temp.int", 42),
    ).Emit("doesn't contain permanent.version")
```

### DIFF
--- a/_examples/logs/main.go
+++ b/_examples/logs/main.go
@@ -53,7 +53,7 @@ type MyHandler struct {
 
 // ServeHTTP example of a handler
 // To correlate logs with transactions, [context.Context] needs to be passed to the [LogEntry] with the [WithCtx] func.
-// Assuming you are using and sentry tracing integration.
+// Assuming you are using a Sentry tracing integration.
 func (h MyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	// By using [WithCtx] the log entry will be associated with the transaction from the request
--- a/_examples/logs/main.go
+++ b/_examples/logs/main.go
@@ -21,8 +21,7 @@ func main() {
 
 	ctx := context.Background()
 	loggerWithAttrs := sentry.NewLogger(ctx)
-	// Attaching permanent attributes on the logger. The attributes are used for indexing reasons and thus, only
-	// the basic primitive types are supported.
+	// Attaching permanent attributes on the logger.
 	loggerWithAttrs.SetAttributes(
 		attribute.String("version", "1.0.0"),
 	)
--- a/_examples/logs/main.go
+++ b/_examples/logs/main.go
@@ -11,7 +11,7 @@ import (
 func main() {
 	err := sentry.Init(sentry.ClientOptions{
 		Dsn:        "",
-		EnableLogs: true,
+		EnableLogs: true, // you need to have EnableLogs set to true
 	})
 	if err != nil {
 		panic(err)
@@ -19,19 +19,24 @@ func main() {
 	defer sentry.Flush(2 * time.Second)
 
 	ctx := context.Background()
-	logger := sentry.NewLogger(ctx)
-
-	// You can use the logger like [fmt.Print]
-	logger.Info(ctx, "Expecting ", 2, " params")
-	// or like [fmt.Printf]
-	logger.Infof(ctx, "format: %v", "value")
-
-	// Additionally, you can also set attributes on the log like this
-	logger.SetAttributes(
-		attribute.Int("key.int", 42),
-		attribute.Bool("key.boolean", true),
-		attribute.Float64("key.float", 42.4),
-		attribute.String("key.string", "string"),
+	loggerWithVersion := sentry.NewLogger(ctx)
+	// Setting  attributes (these persist across log calls)
+	loggerWithVersion.SetAttributes(
+		attribute.String("permanent.version", "1.0.0"),
 	)
-	logger.Warnf(ctx, "I have params %v and attributes", "example param")
+	// Add attributes to log entry
+	loggerWithVersion.Info().WithCtx(ctx).
+		String("key.string", "value").
+		Int("key.int", 42).
+		Bool("key.bool", true).
+		// don't forget to call Emit to send the logs to Sentry
+		Emitf("Message with parameters %d and %d", 1, 2)
+
+	logger := sentry.NewLogger(ctx)
+	// you can also use Attributes for setting many attributes in the log entry (these are temporary)
+	logger.Info().
+		Attributes(
+			attribute.String("key.temp.string", "value"),
+			attribute.Int("key.temp.int", 42),
+		).Emit("doesn't contain permanent.version")
 }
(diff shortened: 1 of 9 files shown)


log.go
```diff
@@ -63,11 +73,11 @@ func NewLogger(ctx context.Context) Logger {
 func (l *sentryLogger) Write(p []byte) (int, error) {
 	// Avoid sending double newlines to Sentry
 	msg := strings.TrimRight(string(p), "\n")
-	l.log(context.Background(), LogLevelInfo, LogSeverityInfo, msg)
+	l.Info().WithCtx(context.Background()).Emit(msg)
```
lcian: Should we use `l.ctx` instead?

giortzisg: Yes we should

log.go
```diff
@@ -182,49 +190,145 @@ func (l *sentryLogger) SetAttributes(attrs ...attribute.Builder) {
 	}
 }
 
-func (l *sentryLogger) Trace(ctx context.Context, v ...interface{}) {
-	l.log(ctx, LogLevelTrace, LogSeverityTrace, fmt.Sprint(v...))
+func (l *sentryLogger) Trace() LogEntry {
+	return &logEntry{
+		logger:     l,
+		ctx:        context.Background(),
```
lcian: Also here I think it's the same issue. I think the log entry should inherit the logger context, right?

lcian: But haven't looked at LogEntry yet so this might be wrong

log.go
```diff
@@ -182,49 +190,145 @@ func (l *sentryLogger) SetAttributes(attrs ...attribute.Builder) {
...
+
+func (e *logEntry) Float64(key string, value float64) LogEntry {
+	e.attributes[key] = Attribute{Value: value, Type: "double"}
+	return e
 }
-func (l *sentryLogger) Warnf(ctx context.Context, format string, v ...interface{}) {
-	l.log(ctx, LogLevelWarn, LogSeverityWarning, format, v...)
+
+func (e *logEntry) Bool(key string, value bool) LogEntry {
+	e.attributes[key] = Attribute{Value: value, Type: "boolean"}
```
lcian: I think we could change the `type` to a map.

interfaces.go
```diff
@@ -101,58 +101,48 @@ func (b *Breadcrumb) MarshalJSON() ([]byte, error) {
...
+
+// LogEntry defines the interface for a log entry that supports chaining attributes.
+type LogEntry interface {
+	WithCtx(ctx context.Context) LogEntry
+	String(key, value string) LogEntry
+	Int(key string, value int) LogEntry
+	Int64(key string, value int64) LogEntry
+	Float64(key string, value float64) LogEntry
+	Bool(key string, value bool) LogEntry
+	Attributes(attrs ...attribute.Builder) LogEntry
```
lcian: We can consider removing this as it's redundant with the other methods.

lcian (commented):
Left some comments

_examples/logs/main.go
```diff
@@ -11,27 +12,51 @@ import (
...
-		attribute.String("key.string", "string"),
-	)
-	logger.Warnf(ctx, "I have params %v and attributes", "example param")
+type MyHandler struct {
+	logger sentry.Logger
+}
+
+// ServeHTTP example of a handler
+// To correlate logs with transactions, [context.Context] needs to be passed to the [LogEntry] with the [WithCtx] func.
+// Assuming you are using and sentry tracing integration.
```
lcian:
```suggestion
// Assuming you are using a Sentry tracing integration.
```

_examples/logs/main.go
```diff
@@ -11,27 +12,51 @@ import (
...
+		EnableLogs: true, // you need to have EnableLogs set to true
 	})
 	if err != nil {
 		panic(err)
 	}
 	defer sentry.Flush(2 * time.Second)
 
 	ctx := context.Background()
+	loggerWithAttrs := sentry.NewLogger(ctx)
+	// Attaching permanent attributes on the logger. The attributes are used for indexing reasons and thus, only
```
cleptric: indexing reason, which means?

log.go
```diff
@@ -90,59 +100,57 @@ func (l *sentryLogger) log(ctx context.Context, level LogLevel, severity int, me
...
-				attrs["user.name"] = Attribute{Value: user.Name, Type: "string"}
+				attrs["user.name"] = Attribute{Value: user.Name, Type: AttributeString}
 			}
 			if user.Email != "" {
-				attrs["user.email"] = Attribute{Value: user.Email, Type: "string"}
+				attrs["user.email"] = Attribute{Value: user.Email, Type: AttributeString}
 			}
 		}
 	}
 	if spanID.String() != "0000000000000000" {
```
cleptric: How can this even happen?

_examples/logs/main.go
giortzisg: Wanted to say that the attributes are getting indexed so that you can search your logs. But yeah, initial wording is cumbersome.

log.go
giortzisg: If there is no propagation context, then the spanID would be empty. So just emitting it in that case.

Issue #1048, opened 2025-06-30
Add tests for simulating race conditions


giortzisg:
Create tests that should run on the pipeline to try and catch race conditions on the hub & scope.

PR #1050, opened 2025-07-02
Fix race conditions on log when accessing scope

giortzisg:
Closes both #1048  and #1049.

## Description

Adds tests for the logger to identify race conditions and fix everything uncovered.

Changes include: 
* Add locks when using `SetAttributes` to guard for panics.
* Lock scope when trying to access propagation context on `log`.

### DIFF
--- a/log.go
+++ b/log.go
@@ -6,6 +6,7 @@ import (
 	"maps"
 	"os"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/getsentry/sentry-go/attribute"
@@ -43,6 +44,7 @@ type sentryLogger struct {
 	ctx        context.Context
 	client     *Client
 	attributes map[string]Attribute
+	mu         sync.RWMutex
 }
 
 type logEntry struct {
@@ -64,7 +66,12 @@ func NewLogger(ctx context.Context) Logger {
 
 	client := hub.Client()
 	if client != nil && client.batchLogger != nil {
-		return &sentryLogger{ctx, client, make(map[string]Attribute)}
+		return &sentryLogger{
+			ctx:        ctx,
+			client:     client,
+			attributes: make(map[string]Attribute),
+			mu:         sync.RWMutex{},
+		}
 	}
 
 	DebugLogger.Println("fallback to noopLogger: enableLogs disabled")
@@ -89,13 +96,21 @@ func (l *sentryLogger) log(ctx context.Context, level LogLevel, severity int, me
 
 	var traceID TraceID
 	var spanID SpanID
+	var span *Span
+	var user User
 
-	span := hub.Scope().span
-	if span != nil {
-		traceID = span.TraceID
-		spanID = span.SpanID
-	} else {
-		traceID = hub.Scope().propagationContext.TraceID
+	scope := hub.Scope()
+	if scope != nil {
+		scope.mu.Lock()
+		span = scope.span
+		if span != nil {
+			traceID = span.TraceID
+			spanID = span.SpanID
+		} else {
+			traceID = scope.propagationContext.TraceID
+		}
+		user = scope.user
+		scope.mu.Unlock()
 	}
 
 	attrs := map[string]Attribute{}
@@ -110,9 +125,12 @@ func (l *sentryLogger) log(ctx context.Context, level LogLevel, severity int, me
 		}
 	}
 
+	l.mu.RLock()
 	for k, v := range l.attributes {
 		attrs[k] = v
 	}
+	l.mu.RUnlock()
+
 	for k, v := range entryAttrs {
 		attrs[k] = v
 	}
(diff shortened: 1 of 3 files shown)


hub.go
```diff
@@ -407,9 +428,15 @@ func (hub *Hub) GetTraceparent() string {
...
+	}
+	scope.mu.RLock()
+	defer scope.mu.RUnlock()
 
-	if scope.span != nil {
-		return scope.span.ToBaggage()
+	span := scope.span
+	if span != nil {
+		return span.ToBaggage()
 	}
```
lcian:
```suggestion

	if span := scope.GetSpan(); span != nil {
		return span.ToBaggage()
	}
```

hub.go
```diff
@@ -393,9 +408,15 @@ func (hub *Hub) FlushWithContext(ctx context.Context) bool {
...
+	}
+	scope.mu.RLock()
+	defer scope.mu.RUnlock()
 
-	if scope.span != nil {
-		return scope.span.ToSentryTrace()
+	span := scope.span
+	if span != nil {
+		return span.ToSentryTrace()
 	}
```
lcian:
```suggestion
	if span := scope.GetSpan(); span != nil {
		return span.ToSentryTrace()
	}
```

hub.go
lcian: `scope.GetSpan()` already does the locking. Additionally we can do it like this which afaik is more idiomatic :)

lcian:
Ah yes, I haven't reviewed the tests but I assume they correctly test that there are no race conditions.

lcian:
About the Cursor comment:
I guess what could happen in theory is that after calling `hub.Scope()` and retrieving the scope, something else could call `hub.PopScope()`, so you would be acting on a scope that is not the top anymore.
But I don't understand if this can occur in practice and if it would be problematic.

hub.go
giortzisg: Actually just remembered why I did the change like this. After the if block we are still accessing `scope.propagationContext` so we need to lock the scope again.

hub.go
giortzisg: Same as above comment.

transport_test.go
```diff
@@ -789,8 +789,7 @@ func TestHTTPTransportDoesntLeakGoroutines(t *testing.T) {
 
 	transport := NewHTTPTransport()
 	transport.Configure(ClientOptions{
-		Dsn:        "https://test@foobar/1",
-		HTTPClient: http.DefaultClient,
+		Dsn: "https://test@foobar/1",
```
lcian: Leftover from testing?

log_test.go
```diff
@@ -663,12 +662,6 @@ func Test_sentryLogger_TracePropagationWithTransaction(t *testing.T) {
 
 func TestSentryLogger_DebugLogging(t *testing.T) {
 	var buf bytes.Buffer
-	debugLogger := log.New(&buf, "", 0)
-	originalLogger := DebugLogger
-	DebugLogger = debugLogger
-	defer func() {
-		DebugLogger = originalLogger
-	}()
```
lcian, line 671: Leftover from testing?

log.go
```diff
@@ -129,19 +147,16 @@ func (l *sentryLogger) log(ctx context.Context, level LogLevel, severity int, me
...
-				attrs["user.id"] = Attribute{Value: user.ID, Type: AttributeString}
-			}
-			if user.Name != "" {
-				attrs["user.name"] = Attribute{Value: user.Name, Type: AttributeString}
-			}
-			if user.Email != "" {
-				attrs["user.email"] = Attribute{Value: user.Email, Type: AttributeString}
-			}
+
+	if !user.IsEmpty() {
```
lcian, line 151: Do we need to check for `user != nil` too?

lcian: If `scope` was `nil` above, then `user` will also be `nil` here

log.go
```diff
@@ -89,13 +96,21 @@ func (l *sentryLogger) log(ctx context.Context, level LogLevel, severity int, me
...
+	scope := hub.Scope()
+	if scope != nil {
+		scope.mu.Lock()
+		span = scope.span
+		if span != nil {
+			traceID = span.TraceID
+			spanID = span.SpanID
+		} else {
+			traceID = hub.Scope().propagationContext.TraceID
+		}
```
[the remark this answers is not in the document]
lcian: Comment from Cursor seems correct, we can just use `scope`

log.go
lcian: If we consider `scope == nil` possible (and it looks like we do based on the check above), then we should also guard this with `user != nil`.

lcian (commented):
Mostly looks good, please see the comments

transport_test.go
giortzisg: both should be fine. If an HTTPClient is missing you init a new one.

log_test.go
giortzisg: Actually not. Changing the global instance on the DebugLogger would cause race conditions. At some point there was a test which was overwriting it. But the test was removed so we need to remove this as well.

log.go
giortzisg: User can't be nil, it would be an empty struct.

giortzisg:
@lcian For the cursor comment about nil scope. If you clone the Hub and scope should never be empty. We can still guard with nil checks on the `scope.Set` methods, but it should be ok as is right now also.

lcian:
> @lcian For the cursor comment about nil scope. If you clone the Hub and scope should never be empty. We can still guard with nil checks on the `scope.Set` methods, but it should be ok as is right now also.

Yep I see. I was talking about this one: https://github.com/getsentry/sentry-go/pull/1050#discussion_r2253885071
Screenshot 2025-08-07 at 11 38 55

hub_test.go
```diff
@@ -553,3 +555,19 @@ func TestHub_FlushWithContext(t *testing.T) {
 		t.Fatalf("expected message to be %v, got %v", wantEvent.Message, gotEvents[0].Message)
 	}
 }
+
+func TestHubEmptyClientShouldBeSafe(t *testing.T) {
```
cleptric: What does this mean? 😬

giortzisg: That the hub shouldn't panic with an empty client 😅

cleptric: Maybe we use a a test name with proper grammar then?

PR #1058, opened 2025-07-18
feat(tracing): add SetMeasurement support to spans and transactions

yigithanyucedag:
This PR adds support for custom **span measurements** in the Go SDK (`sentry-go`) by introducing a new method:

```go
SetMeasurement(name string, value float64, unit string)
```

This method allows instrumentation of custom metrics (e.g., `duration_ms`, `tokens_used`, etc.) and sends them under the `measurements` field in the transaction payload.

---

## Why

Currently, the Go SDK lacks support for custom `measurements` on spans/transactions — a feature already available in the JS and Python SDKs.

---

## How

- Adds `Span.SetMeasurement` method
- Adds a `measurements` map to the `Span` struct
- Exports this map as part of the final `Event` in `toEvent()`
- Includes unit and value
- Adds `Measurement` struct matching expected envelope format
- Includes a test case (`TestSetMeasurement`) to verify behavior

---

## Example

```go
tx := sentry.StartTransaction(ctx, "Predict")
tx.SetMeasurement("tokens_used", 1423.0, "token")
tx.SetMeasurement("duration_ms", 512.6, "millisecond")
tx.Finish()
```

Sends:

```json
"measurements": {
  "tokens_used": {
    "value": 1423.0,
    "unit": "token"
  },
  "duration_ms": {
    "value": 512.6,
    "unit": "millisecond"
  }
}
```

---

## Notes

- This change is fully backward-compatible.
- If no measurements are set, nothing is serialized.

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -372,30 +372,31 @@ type Context = map[string]interface{}
 
 // Event is the fundamental data structure that is sent to Sentry.
 type Event struct {
-	Breadcrumbs []*Breadcrumb          `json:"breadcrumbs,omitempty"`
-	Contexts    map[string]Context     `json:"contexts,omitempty"`
-	Dist        string                 `json:"dist,omitempty"`
-	Environment string                 `json:"environment,omitempty"`
-	EventID     EventID                `json:"event_id,omitempty"`
-	Extra       map[string]interface{} `json:"extra,omitempty"`
-	Fingerprint []string               `json:"fingerprint,omitempty"`
-	Level       Level                  `json:"level,omitempty"`
-	Message     string                 `json:"message,omitempty"`
-	Platform    string                 `json:"platform,omitempty"`
-	Release     string                 `json:"release,omitempty"`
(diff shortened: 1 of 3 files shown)


yigithanyucedag:
It would be really helpful if Measurements could remain supported until the new API for span data becomes available. From what I can see, this feature has been anticipated for a while, and currently, span data does not allow setting measurement values in a way that supports features like avg() or custom dashboards.

cleptric:
Both things are indeed supported with data attributes. Here is how they are set in the SDK https://docs.sentry.io/platforms/go/tracing/instrumentation/custom-instrumentation/#adding-span--transaction-data-attributes.

And here are all docs for our Trace Explorer https://docs.sentry.io/product/explore/trace-explorer/, where you can plot avg(), p95(), sum() and so on and create a dashboard.

https://sentry.io/explore/traces/ to access is in the UI. The only thing missing are units, but they will follow later this year.

cleptric:
image

You can indeed save it as a dashboard widget.

PR #1061, opened 2025-07-21
Add a RecoveredErrorLogger option to the Handler that allows logging …

spanwp:
This PR adds an option to pass the recovered panic error to a defined function. This provides the opportunity to handle the recovered error in more ways. For example, in our company, any log sent to Sentry is also printed to standard output. The recovery method currently disallows us from achieving that in a simple way.

Please give this Pull Request your consideration.


cleptric:
Wouldn't [`BeforeSend`](https://docs.sentry.io/platforms/go/configuration/filtering/#using-before-send) also allow you to do this?

PR #1075, opened 2025-09-04
feat: properly support error capturing

### LINKED ISSUE #977: Decompose `errors.Join` errors in `SetException`
### Problem Statement

I have rich errors with Sentry-compatible stack traces. I use the Go standard library to `Join` them:

```
first := NewError(...)
second := NewError(...)

joined := errors.Join(first, second)
```

This is a common pattern for error aggregation: it was introduced into the standard library in Go 1.20 (Feb. 2023), and [GitHub indexes 53.2k public instances](https://github.com/search?q=language%3Ago+%22errors.Join%22&type=code).

I'd like to capture `joined` with the Sentry SDK and retain _both_ errors — with their individual stack traces — on the resulting Sentry event.

_Today,_ `CaptureException` yields an event with a _single_ `Exception` (representing `joined`), with _no_ stack trace information associated with it.

### Solution Brainstorm

# Proposal ([Draft PR](https://github.com/getsentry/sentry-go/pull/976))

This SDK plays nice with `errors.Wrap` (see https://github.com/getsentry/sentry-go/pull/792) and `(github.com/pkg/errors).Wrap`. It can introduce analogous support for `errors.Join`.

| Compose                        | Decompose                                                                                                                                             |
|--------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------|
| `errors.Wrap`                  | `errors.Unwrap`                                                                                                                                       |
| `(github.com/pkg/errors).Wrap` | `interface{ Cause() error }` ([source](https://github.com/getsentry/sentry-go/blob/031ec47dc927e3fb737fa818285ccf3f46b47c7f/interfaces.go#L364-L368)) |
| ✨ `errors.Join`                  | `interface{ Unwrap() []error }` ([docs](https://pkg.go.dev/errors))                                                                                   |

## Implementation

Two relevant excerpts from the standard library [errors docs](https://pkg.go.dev/errors):

> A non-nil error returned by `Join` implements the `Unwrap() []error` method.

> It is invalid for an `Unwrap` method to return an `[]error` containing a `nil` error value.

This represents a major change in the data model for `error` composition: _wrapped_ errors form a clear causal chain, whereas _joined_ errors are a tree that can contain chains of wrapped errors. Rather than looping linearly over the causal chain, `SetException` has to 

1. traverse a tree of decomposed errors
2. handle more complex parent/child relationships — a "child" (composed `error`) can have _several_ parents (`first` and `second` in the example above)

The draft diff linked above implements depth-first traversal of the error tree in a new `errorTree` type. It passes the `SetException` unit tests, plus additional cases including joined errors.

# Open questions

- [ ] Does Sentry's treatment of exceptions support multi-parental relationships/trees of exceptions?
    - [ ] Accordingly, how should joined errors be represented as `[]sentry.Exception`s?
- [ ] How should `maxExceptionDepth` translate to a tree data model?

giortzisg:
## Description

This PR adds support for capturing all possible errors (including wrapped ones) in Go. The following changes were made: 
- add support for unwrapping errors that implement the `Unwrap() []error` interface, by parsing all errors in a dfs-like traversal.
- properly handle `is_exception_group`. In the end made the decision to only capture `Unwrap() []error` errors as exception_groups. This breaks the previous behavior of including any chained error in a group.
- support for third party libraries using `Cause() error`.
- change the default maxErrorDepth to 100.

### maxErrorDepth

For the change to the default after a quick investigation these are some rough payload sizes based on the amount of errors that get unwrapped. Setting it to 100 since the payload sizes are not that big to cause any issues.

**Built-in errors**

Depth | max=unlimited | max=1 | max=5 | max=10 | max=20
-- | -- | -- | -- | -- | --
1 | 0.5KB | 0.5KB | 0.5KB | 0.5KB | 0.5KB
3 | 0.8KB | 0.6KB | 0.8KB | 0.8KB | 0.8KB
5 | 1.2KB | 0.6KB | 1.2KB | 1.2KB | 1.2KB
10 | 2.3KB | 0.6KB | 1.2KB | 2.3KB | 2.3KB
20 | 5.1KB | 0.6KB | 1.2KB | 2.3KB | 5.1KB
50 | 19.4KB | 0.6KB | 1.2KB | 2.3KB | 5.1KB
100 | 62.9KB | 0.6KB | 1.2KB | 2.3KB | 5.1KB

**Third-party pkg/errors**

Depth | max=unlimited | max=1 | max=5 | max=10 | max=20
-- | -- | -- | -- | -- | --
1 | 0.9KB | 0.9KB | 0.9KB | 0.9KB | 0.9KB
3 | 2.6KB | 1.0KB | 2.6KB | 2.6KB | 2.6KB
5 | 4.3KB | 1.0KB | 2.6KB | 4.3KB | 4.3KB
10 | 8.8KB | 1.0KB | 2.6KB | 4.5KB | 8.8KB
20 | 19.3KB | 1.0KB | 2.6KB | 4.5KB | 9.1KB
50 | 62.4KB | 1.0KB | 2.6KB | 4.5KB | 9.1KB
100 | 173.6KB | 1.0KB | 2.7KB | 4.6KB | 9.1KB

#### Issues
- resolves #977.
- resolves GO-4.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -40,7 +40,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,10 +18,10 @@ jobs:
     name: Lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/setup-go@v5
+      - uses: actions/setup-go@v6
         with:
           go-version: "1.24"
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: golangci-lint
         uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 # pin@v8.0
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,11 +15,11 @@ jobs:
     steps:
       - name: Get auth token
         id: token
-        uses: actions/create-github-app-token@df432ceedc7162793a195dd1713ff69aefc7379e # v2.0.6
+        uses: actions/create-github-app-token@67018539274d69449ef7c02e8e71183d1719ab42 # v2.1.4
         with:
           app-id: ${{ vars.SENTRY_RELEASE_BOT_CLIENT_ID }}
           private-key: ${{ secrets.SENTRY_RELEASE_BOT_PRIVATE_KEY }}
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           token: ${{ steps.token.outputs.token }}
           fetch-depth: 0
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,10 +26,10 @@ jobs:
       # this step for the whole matrix.
       RUN_RACE_TESTS: ${{ github.event_name != 'pull_request' || (matrix.go == '1.24' && matrix.os == 'ubuntu') }}
     steps:
-      - uses: actions/setup-go@v5
+      - uses: actions/setup-go@v6
         with:
           go-version: ${{ matrix.go }}
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - uses: actions/cache@v4
         with:
           # In order:
@@ -47,7 +47,7 @@ jobs:
             ${{ runner.os }}-go-${{ matrix.go }}-
       - name: Tidy for min version
         run: make mod-tidy
-        if: ${{ matrix.go == '1.22' }}
+        if: ${{ matrix.go == '1.23' }}
       - name: Build
         run: make build
       - name: Vet
@@ -62,6 +62,6 @@ jobs:
     timeout-minutes: 15
     strategy:
       matrix:
-        go: ["1.24", "1.23", "1.22"]
+        go: ["1.25", "1.24", "1.23"]
         os: [ubuntu, windows, macos]
       fail-fast: false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The Sentry SDK team is happy to announce the immediate availability of Sentry Go
 
 ### Breaking Changes
 
+- Behavioral change for the `MaxBreadcrumbs` client option. Removed the hard limit of 100 breadcrumbs, allowing users to set a larger limit and also changed the default limit from 30 to 100 ([#1106](https://github.com/getsentry/sentry-go/pull/1106)))
+
 - The changes to error handling ([#1075](https://github.com/getsentry/sentry-go/pull/1075)) will affect issue grouping. It is expected that any wrapped and complex errors will be grouped under a new issue group.
 
 ### Features
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,7 @@ mod-tidy: ## Check go.mod tidiness
 	set -e ; \
 	for dir in $(ALL_GO_MOD_DIRS); do \
 		echo ">>> Running 'go mod tidy' for module: $${dir}"; \
-		(cd "$${dir}" && go mod tidy -go=1.22 -compat=1.22); \
+		(cd "$${dir}" && go mod tidy -go=1.23 -compat=1.23); \
 	done; \
 	git diff --exit-code;
 .PHONY: mod-tidy
--- a/batch_logger.go
+++ b/batch_logger.go
@@ -125,7 +125,8 @@ func (l *BatchLogger) run(ctx context.Context) {
 func (l *BatchLogger) processEvent(logs []Log) {
 	event := NewEvent()
 	event.Timestamp = time.Now()
+	event.EventID = EventID(uuid())
 	event.Type = logEvent.Type
 	event.Logs = logs
-	l.client.CaptureEvent(event, nil, nil)
+	l.client.Transport.SendEvent(event)
 }
--- a/client.go
+++ b/client.go
@@ -5,7 +5,6 @@ import (
 	"crypto/x509"
 	"fmt"
 	"io"
-	"log"
 	"math/rand"
 	"net/http"
 	"os"
@@ -15,24 +14,31 @@ import (
 	"time"
 
 	"github.com/getsentry/sentry-go/internal/debug"
+	"github.com/getsentry/sentry-go/internal/debuglog"
 )
 
 // The identifier of the SDK.
 const sdkIdentifier = "sentry.go"
 
-// maxErrorDepth is the maximum number of errors reported in a chain of errors.
-// This protects the SDK from an arbitrarily long chain of wrapped errors.
-//
-// An additional consideration is that arguably reporting a long chain of errors
-// is of little use when debugging production errors with Sentry. The Sentry UI
-// is not optimized for long chains either. The top-level error together with a
-// stack trace is often the most useful information.
-const maxErrorDepth = 100
+const (
+	// maxErrorDepth is the maximum number of errors reported in a chain of errors.
+	// This protects the SDK from an arbitrarily long chain of wrapped errors.
+	//
+	// An additional consideration is that arguably reporting a long chain of errors
+	// is of little use when debugging production errors with Sentry. The Sentry UI
+	// is not optimized for long chains either. The top-level error together with a
+	// stack trace is often the most useful information.
+	maxErrorDepth = 100
 
-// defaultMaxSpans limits the default number of recorded spans per transaction. The limit is
-// meant to bound memory usage and prevent too large transaction events that
-// would be rejected by Sentry.
-const defaultMaxSpans = 1000
+	// defaultMaxSpans limits the default number of recorded spans per transaction. The limit is
+	// meant to bound memory usage and prevent too large transaction events that
+	// would be rejected by Sentry.
+	defaultMaxSpans = 1000
+
+	// defaultMaxBreadcrumbs is the default maximum number of breadcrumbs added to
+	// an event. Can be overwritten with the MaxBreadcrumbs option.
+	defaultMaxBreadcrumbs = 100
+)
 
 // hostname is the host name reported by the kernel. It is precomputed once to
 // avoid syscalls when capturing events.
@@ -78,8 +84,8 @@ type usageError struct {
 }
 
 // DebugLogger is an instance of log.Logger that is used to provide debug information about running Sentry Client
-// can be enabled by either using DebugLogger.SetOutput directly or with Debug client option.
-var DebugLogger = log.New(io.Discard, "[Sentry] ", log.LstdFlags)
+// can be enabled by either using debuglog.SetOutput directly or with Debug client option.
+var DebugLogger = debuglog.GetLogger()
 
 // EventProcessor is a function that processes an event.
 // Event processors are used to change an event before it is sent to Sentry.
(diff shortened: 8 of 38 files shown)


error_group.go
```diff
@@ -0,0 +1,94 @@
...
+	if len(exceptions) == 1 {
+		exceptions[0].Mechanism = nil
+	}
+
+	slices.Reverse(exceptions)
+
+	return exceptions
+}
+
+func convertErrorDFS(err error, exceptions *[]Exception, parentID *int, source string, visited map[error]bool) {
```
lcian: `m`: This could use a comment describing what the params are for, to help readers.

exception.go
```diff
@@ -0,0 +1,94 @@
...
+	MechanismSourceCause string = "cause"
+)
+
+func convertErrorToExceptions(err error) []Exception {
+	var exceptions []Exception
+	visited := make(map[error]bool)
+	convertErrorDFS(err, &exceptions, nil, "", visited)
+
+	if len(exceptions) == 1 {
+		exceptions[0].Mechanism = nil
```
lcian, line 24: `l`: For the record, the spec https://develop.sentry.dev/sdk/data-model/event-payloads/exception/#exception-mechanism says 

> For user-invoked exception captures (e.g. via captureException), the type MUST be set to 'generic'.

Although I suspect we won't be able to distinguish between user vs integration invoked captures without changing the code significantly, so this is not a big deal to me if we cannot do it.

exception.go
```diff
@@ -0,0 +1,94 @@
...
+	switch v := err.(type) {
+	case interface{ Unwrap() []error }:
+		unwrapped := v.Unwrap()
+		for i := range unwrapped {
+			if unwrapped[i] != nil {
+				childSource := fmt.Sprintf("errors[%d]", i)
+				convertErrorDFS(unwrapped[i], exceptions, &currentID, childSource, visited)
+			}
+		}
+	case interface{ Unwrap() error }:
```
lcian, line 92: `l`: Should we use the `source` here too, to indicate e.g. `Unwrap`/`unwrap` or `Cause`/`cause`?

interfaces.go
```diff
@@ -425,64 +422,14 @@ func (e *Event) SetException(exception error, maxErrorDepth int) {
...
-		// using the github.com/pkg/errors package.
-		err = cause.Cause()
-	}
-
-	// Add a trace of the current stack to the most recent error in a chain if
-	// it doesn't have a stack trace yet.
-	// We only add to the most recent error to avoid duplication and because the
-	// current stack is most likely unrelated to errors deeper in the chain.
-	if e.Exception[0].Stacktrace == nil {
-		e.Exception[0].Stacktrace = NewStacktrace()
```
lcian, line 465: `h`: Removing this means that capturing a simple `errors.New(<message>)` (and possibly other things, see also tests above, but that's the one I've tried) won't carry a stack trace anymore.
Therefore, the grouping will happen by message rather than stack trace (previous behavior).

Just checking to make sure this change is intentional and it's the behavior we want.

error_group.go
```diff
@@ -0,0 +1,94 @@
...
+	MechanismSourceCause string = "cause"
+)
+
+func convertErrorToExceptions(err error) []Exception {
+	var exceptions []Exception
+	visited := make(map[error]bool)
+	convertErrorDFS(err, &exceptions, nil, "", visited)
+
+	if len(exceptions) == 1 {
+		exceptions[0].Mechanism = nil
```
giortzisg: I think that applies to normal exceptions. For exception groups according to the rfc here https://github.com/getsentry/rfcs/blob/main/text/0079-exception-groups.md#mechanism-type, the root exception should be set to generic. Maybe we need to clarify it on the develop docs.

exception.go
giortzisg: We actually use the hardcoded `cause` as a source. According to [this](https://github.com/getsentry/rfcs/blob/main/text/0079-exception-groups.md#source), it made sense to follow python's approach and just use `cause`. Maybe it would make sense to have `unwrap` as well.

interfaces.go
giortzisg: Yes you are correct, we need to keep the behavior.

exception.go
giortzisg: we don't really have a way to distinguish the origin of where the error is coming from right now. Also the mechanism is only for debugging purposes, so I guess setting it to nil is ok. For wrapped errors we normally set it to `chained` as expected.

error_group.go
```diff
@@ -0,0 +1,103 @@
...
+		unwrapped := v.Unwrap()
+		if unwrapped != nil {
+			unwrappedTypeStr := reflect.TypeOf(unwrapped).String()
+			currentTypeStr := reflect.TypeOf(err).String()
+			// This specifically catches cases like go-errors.New() where the error wraps a string
+			if unwrapped.Error() == err.Error() && unwrappedTypeStr == "*errors.errorString" && currentTypeStr == "*errors.Error" {
+				exception.Mechanism.IsExceptionGroup = false
+			} else {
+				convertErrorDFS(unwrapped, exceptions, &currentID, MechanismSourceCause, visited)
+			}
```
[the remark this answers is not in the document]
giortzisg: We are also checking that unwrappedTypeStr == "*errors.errorString" && currentTypeStr == "*errors.Error", so that we don't terminate early.

interfaces.go
```diff
@@ -425,64 +422,25 @@ func (e *Event) SetException(exception error, maxErrorDepth int) {
...
-			IsExceptionGroup: true,
-			ExceptionID:      i,
-			Type:             "generic",
-		}
-		if i == 0 {
-			continue
-		}
-		e.Exception[i].Mechanism.ParentID = Pointer(i - 1)
-	}
+	e.Exception = exceptions
```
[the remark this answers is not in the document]
lcian: The comment from Cursor is correct.
If you have a chain of length `maxErrorDepth + 1` you would cut off the top-level one, which is what we just attached the stacktrace to.
Another problematic case is if you have say 1 parent error with `maxErrorDepth` children (with `Causes error[]`). You would remove the parent, even though the total depth is just 2.
This proves we're not really considering the depth, we're just counting the overall number of errors.

The correct implementation would be to pass this into `convertErrorDFS` and stop the visit when you reach the depth limit.

error_group.go
```diff
@@ -0,0 +1,103 @@
...
+
+	if visited[err] {
+		return
+	}
+	visited[err] = true
+
+	var isExceptionGroup bool
+
+	switch err.(type) {
+	case interface{ Unwrap() []error }, interface{ Unwrap() error }, interface{ Cause() error }:
```
lcian: I believe this is not correct according to the RFC and to how other SDKs work.
I see you've added a workaround on line 91 to handle some specific cases (which I haven't fully though about) but the problem still stands.

The correct thing here would be to set `isExceptionGroup = true` only when we have `Unwrap() []error`.
That really matches the concept of exception **group**. 
In other cases, what you have is really a **chain**.

Please try the following example:
```go
	deepErr := errors.New("root cause")
	for i := 1; i <= 5; i++ {
		deepErr = fmt.Errorf("layer %d: %w", i, deepErr)
	}
	deepErr = fmt.Errorf("top level error: %w", deepErr)
	fmt.Println(deepErr)
	sentry.CaptureException(deepErr)
```

With the current code, you would get this: https://sentry-sdks.sentry.io/issues/6915067203/events/535bf98e4823453fb4ae3cb85fcb9837/

With the "correct" version (change this line to `case interface{ Unwrap []error }:`) you would get this:
https://sentry-sdks.sentry.io/issues/6915069781/?project=4508683222843393&query=&referrer=issue-stream

Note the title difference.
Between the two, I would prefer the second one, because I see the full top-level message, not the innermost child.

The good thing about the "incorrect" approach that we have in the current commit I'm reviewing, is that it doesn't break grouping with respect to issues created by older versions of the SDK.
This is because so far the SDK has incorrectly reported chains as groups, see https://github.com/getsentry/sentry-go/blame/8921d3c248d9619dea4bf9f1fe909a76f5982d48/interfaces.go#L477 and the git blame.
In other words, handling chains correctly would create new groups.

So the tradeoff here is between correctness + better titling vs not breaking grouping for existing chains, and I think we should consider it or at least be aware of it.

lcian: We could also leave it as is (given that the problem was already there) and consider changing it in a future version.

giortzisg: In the end it's what we consider a group vs what we don't in the language, so IMO chained errors can be considered groups. The only thing that changes, behavior wise is that in the UI, the chained errors get hidden (and grouped together), so I was personally in favor of that as an approach. However if it affects titles, we should follow the "correct" way so that the whole title message is included.

exception.go
```diff
@@ -0,0 +1,108 @@
...
+		ParentID:         parentID,
+		Source:           source,
+		IsExceptionGroup: isExceptionGroup,
+	}
+
+	*exceptions = append(*exceptions, exception)
+
+	if maxErrorDepth >= 0 && currentDepth >= maxErrorDepth {
+		return
+	}
```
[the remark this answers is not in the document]
lcian: (see Cursor comment https://github.com/getsentry/sentry-go/pull/1075/files#r2394819108)
`h`: True, let's just move the check closer to the top

exception.go
```diff
@@ -0,0 +1,103 @@
...
+		for i := range unwrapped {
+			if unwrapped[i] != nil {
+				childSource := fmt.Sprintf("errors[%d]", i)
+				convertErrorDFS(unwrapped[i], exceptions, &currentID, childSource, visited, maxErrorDepth, currentDepth+1)
+			}
+		}
+	case interface{ Unwrap() error }:
+		unwrapped := v.Unwrap()
+		if unwrapped != nil {
+			convertErrorDFS(unwrapped, exceptions, &currentID, MechanismSourceCause, visited, maxErrorDepth, currentDepth+1)
```
lcian: `l`: we could use a new `MechanismSourceUnwrap` here given that `Unwrap` is how we obtain it, not `Source`.

exception.go
```diff
@@ -0,0 +1,103 @@
...
+	currentID := len(*exceptions)
+
+	var mechanismType string
+
+	if parentID == nil {
+		mechanismType = MechanismTypeGeneric
+		source = ""
+	} else {
+		mechanismType = MechanismTypeChained
+	}
```
lcian, line 67: `l`: this is not entirely correct because we would set the `chained` mechanism to groups as well, which are not chains. So maybe we set it to `generic` unless we know for sure it's a chain?

But at this point I'm questioning the entire point of the mechanism type, it seems quite useless the way we're using it for this SDK for both us and SDK users, it doesn't provide any information.
Unfortunately it's required if you want to have a mechanism...
So we could also leave it like this, it can be revisited in the future.

exception.go
```diff
@@ -0,0 +1,103 @@
+package sentry
```
lcian, line 1: `l`: the file is called `error_group.go` but it handles/creates things that are not necessarily groups, maybe call it `exception.go` or something?

exception.go
```diff
@@ -0,0 +1,103 @@
...
+func convertErrorToExceptions(err error, maxErrorDepth int) []Exception {
+	var exceptions []Exception
+	visited := make(map[error]bool)
+	convertErrorDFS(err, &exceptions, nil, "", visited, maxErrorDepth, 0)
+
+	// mechanism type is used for debugging purposes, but since we can't really distinguish the origin of who invoked
+	// captureException, we set it to nil if the error is not chained.
+	if len(exceptions) == 1 {
+		exceptions[0].Mechanism = nil
+	}
```
lcian, line 25: `l`: what's even the point of this now?

lcian (approved):
LGTM, thanks for bearing with me.
Left some more nits that you can also disregard (except the `h` one please).
Issue Grouping changes should be documented in the changelog.

exception.go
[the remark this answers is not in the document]
lcian: Wait disregard, the current version is correct.
We add up to the max (included) and then return.

exception.go
lcian: PR also went out of scope, would consider changing the commit message.

exception.go
giortzisg: We would still get a type `generic`, for single errors if we don't set it to nil and that is misleading.

PR #1082, opened 2025-09-11
add missing rate limit categories

### LINKED ISSUE #1081: Add rate limit for logs
We should add logs (and other categories that were missed) in our RateLimiter.

giortzisg:
Closes #1081. 
Properly handle rate limits by:
- Add missing types for Logs and CheckIns. 
- Add CategoryUnknown to enum for debugging purposes. Unknown categories should not get rate limited.
- Tidy: move category transformation to Event.

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -605,7 +605,7 @@ func (e *Event) checkInMarshalJSON() ([]byte, error) {
 	return json.Marshal(checkIn)
 }
 
-func (e *Event) ToCategory() ratelimit.Category {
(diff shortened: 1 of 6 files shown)


interfaces.go
```diff
@@ -604,6 +605,21 @@ func (e *Event) checkInMarshalJSON() ([]byte, error) {
 	return json.Marshal(checkIn)
 }
 
+func (e *Event) ToCategory() ratelimit.Category {
```
cleptric: Don't think we need to export this, only used in `transport.go`.

PR #1085, opened 2025-09-15
feat: Add Spotlight integration for local development debugging

### LINKED ISSUE #846: Add support for Spotlight
## Summary

If possible, I would love to have spotlight integration to this Sentry SDK.

## Motivation

Spotlight is slowly becoming available across all of the other sentry SDKs and it would be extremely helpful to have it here too.

DAcodedBEAT:
resolves: #846

Add SpotlightTransport that sends events to local Spotlight server for real-time debugging during development. The integration works by wrapping the existing transport and duplicating events to Spotlight.

Features:
- Spotlight transport decorator supporting custom URL overrides
- SENTRY_SPOTLIGHT environment variable for easy enabling
- Example program demonstrating usage patterns
- README documentation update


DAcodedBEAT:
> I'm ok merging this now instead of waiting for our refactoring, as the surface area is indeed rather small. If you could address the lint issue, we can review it and write docs prior the release.

Sorry for missing the lint errors! I've fixed it and re-pushed (trying to keep a clean git history for y'all). Let me know if there's anything else I can do to get this merged!

client.go
```diff
@@ -370,6 +377,12 @@ func NewClient(options ClientOptions) (*Client, error) {
 }
 
 func (client *Client) setupTransport() {
+	if !client.options.Spotlight {
+		if spotlightEnv := os.Getenv("SENTRY_SPOTLIGHT"); spotlightEnv == "true" || spotlightEnv == "1" {
+			client.options.Spotlight = true
+		}
+	}
+
```
giortzisg: nit: This option should be on the `NewClient` method.

transport.go
```diff
@@ -742,3 +746,99 @@ func (noopTransport) FlushWithContext(context.Context) bool {
...
+	if options.SpotlightURL != "" {
+		st.spotlightURL = options.SpotlightURL
+	}
+}
+
+func (st *SpotlightTransport) SendEvent(event *Event) {
+	// Send to the underlying transport (Sentry) unless it's noopTransport
+	if _, isNoop := st.underlying.(noopTransport); !isNoop {
+		st.underlying.SendEvent(event)
+	}
```
giortzisg: We don't really need the check here. If the user incorrectly set the transport it's helpful to get the debug message from the noopTransport when sending the event.

transport.go
```diff
@@ -484,7 +484,11 @@
 	}
 
 fail:
-	DebugLogger.Println("Buffer flushing was canceled or timed out.")
+	if itemCount := len(b.items); itemCount > 0 {
+		DebugLogger.Printf("Buffer flushing was canceled or timed out. %d items were not flushed.", itemCount)
+	} else {
+		DebugLogger.Println("Buffer flushing was canceled or timed out. No items were pending.")
+	}
```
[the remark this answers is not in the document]
giortzisg: Cursor is right here.

transport.go
```diff
@@ -742,3 +746,99 @@ func (noopTransport) FlushWithContext(context.Context) bool {
...
+
+	req, err := http.NewRequest("POST", st.spotlightURL, envelope)
+	if err != nil {
+		DebugLogger.Printf("Failed to create Spotlight request: %v", err)
+		return
+	}
+
+	req.Header.Set("Content-Type", "application/x-sentry-envelope")
+	req.Header.Set("User-Agent", "sentry-go/"+SDKVersion)
```
giortzisg: nit: We should use the `event.Sdk.Name` and `event.Sdk.Version` here.

transport.go
```diff
@@ -742,3 +746,99 @@ func (noopTransport) FlushWithContext(context.Context) bool {
...
+		return
+	}
+
+	envelope, err := envelopeFromBody(event, dsn, time.Now(), eventBody)
+	if err != nil {
+		DebugLogger.Printf("Failed to create Spotlight envelope: %v", err)
+		return
+	}
+
+	req, err := http.NewRequest("POST", st.spotlightURL, envelope)
```
giortzisg: nit: we should use ctx here.

transport.go
```diff
@@ -742,3 +746,99 @@ func (noopTransport) FlushWithContext(context.Context) bool {
...
+		if closeErr := resp.Body.Close(); closeErr != nil {
+			DebugLogger.Printf("Failed to close Spotlight response body: %v", closeErr)
+		}
+	}()
+
+	DebugLogger.Printf("Spotlight response status: %d", resp.StatusCode)
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		DebugLogger.Printf("Spotlight returned non-2xx status: %d", resp.StatusCode)
+		if body, err := io.ReadAll(resp.Body); err == nil && len(body) > 0 {
```
giortzisg: In order to reuse the http connection we should drain the body on every request, not only on failures.

spotlight_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+func (m *mockTransport) Flush(time.Duration) bool {
+	return true
+}
+
+func (m *mockTransport) FlushWithContext(_ context.Context) bool {
+	return true
+}
+
+func (m *mockTransport) Close() {}
```
giortzisg: we already have a `MockTransport` instance that can be used for tests.

spotlight_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+)
+
+func TestSpotlightTransport(t *testing.T) {
```
giortzisg, line 20: It would be nice if we can extend this test to verify that the underlying transport actually flushes and sends the events as well, so that it guards against breaking it.

giortzisg (commented):
Thanks for the contribution. Left some small comments. Let's get them fixed and we can merge this.

DAcodedBEAT:
@giortzisg / @cleptric - may I please get some re-reviews on this?

client.go
```diff
@@ -341,6 +348,13 @@ func NewClient(options ClientOptions) (*Client, error) {
 		options.TraceIgnoreStatusCodes = [][]int{{404}}
 	}
 
+	// Check for Spotlight environment variable
+	if !options.Spotlight {
+		if spotlightEnv := os.Getenv("SENTRY_SPOTLIGHT"); spotlightEnv == "true" || spotlightEnv == "1" {
```
giortzisg: We should adhere to these [values](https://develop.sentry.dev/sdk/expected-features/spotlight/#truthy-values).

We also need to extract the url if provided with `SENTRY_SPOTLIGHT`

client.go
```diff
@@ -408,6 +422,10 @@ func (client *Client) setupTransport() {
 		}
 	}
 
+	if opts.Spotlight {
+		transport = NewSpotlightTransport(transport)
+	}
```
giortzisg, line 580: we should also override the sample rate to be 1.0 and PII to true when spotlight is enabled

transport.go
```diff
@@ -746,6 +746,105 @@ func (noopTransport) FlushWithContext(context.Context) bool {
...
+		st.spotlightURL = options.SpotlightURL
+	}
+}
+
+func (st *SpotlightTransport) SendEvent(event *Event) {
+	// Send to the underlying transport (Sentry)
+	st.underlying.SendEvent(event)
+
+	// Always send to Spotlight
+	st.sendToSpotlight(event)
```
giortzisg: The send to spotlight call is sync here, so it kinda defeats the purpose of the async transport. I think the smart solution here to not couple the spotlight transport with the background worker is to inverse the dependency and modify the Sync and Async transport implementations to just invoke the spotlight transport when it exists. This makes the change to the new transport also minimal, where we would just invoke spotlight.Send on the scheduler, along the http request to the sentry backend.

transport.go
```diff
@@ -746,6 +746,105 @@ func (noopTransport) FlushWithContext(context.Context) bool {
...
+}
+
+func (st *SpotlightTransport) sendToSpotlight(event *Event) {
+	ctx := context.Background()
```
giortzisg: we should also support context cancellation during shutdown. Currently during shutdown spotlight might block and we won't get a graceful shutdown.

transport.go
```diff
@@ -746,6 +746,105 @@ func (noopTransport) FlushWithContext(context.Context) bool {
...
+	client       *http.Client
+	spotlightURL string
+}
+
+func NewSpotlightTransport(underlying Transport) *SpotlightTransport {
+	return &SpotlightTransport{
+		underlying: underlying,
+		client: &http.Client{
+			Timeout: 5 * time.Second,
+		},
```
giortzisg, line 1013: the client should respect the proxy settings. why aren't we using the underlying client here?

giortzisg (commented):
We should update the implementation to also use the `internal/telemetry/scheduler` that we added for the new transport (we need to support both for now unfortunately) and also make the spotlight transport implementation async. Also left some comments for inconsistencies with the spec.

transport.go
```diff
@@ -32,6 +32,7 @@ type Transport interface {
 	FlushWithContext(ctx context.Context) bool
 	Configure(options ClientOptions)
 	SendEvent(event *Event)
+	SendEnvelope(envelope *protocol.Envelope)
```
giortzisg, line 35: We should not implement sendEnvelope for the HttpTransport

transport.go
```diff
@@ -507,6 +508,26 @@ func (t *HTTPTransport) SendEventWithContext(ctx context.Context, event *Event)
 	t.buffer <- b
 }
 
+// SendEnvelope sends an envelope to the Sentry server.
+// Currently converts envelope back to event for compatibility.
+// In the future, this will be the primary method.
+func (t *HTTPTransport) SendEnvelope(envelope *protocol.Envelope) {
```
giortzisg, line 514: same here.

transport.go
```diff
@@ -777,6 +798,23 @@ func (t *HTTPSyncTransport) SendEvent(event *Event) {
 	t.SendEventWithContext(context.Background(), event)
 }
 
+func (t *HTTPSyncTransport) SendEnvelope(envelope *protocol.Envelope) {
```
giortzisg, line 801: and here.

transport.go
```diff
@@ -887,6 +929,271 @@ func (noopTransport) FlushWithContext(context.Context) bool {
...
+	spotlightURL string
+
+	// ctx is cancelled on Close to signal goroutines to stop.
+	// wg tracks in-flight sends so Close can wait for them to finish.
+	ctx    context.Context
+	cancel context.CancelFunc
+	wg     sync.WaitGroup
+}
+
+func NewSpotlightTransport(underlying Transport) *SpotlightTransport {
```
giortzisg, line 945: following from the previous comment, when we setup the transport on the client, we already convert the `internal/http/transport` to a `Transport` interface with the `internalAsyncTransportAdapter` so there is no point in satisfying the `SendEnvelope` on the scope of this PR.

client.go
```diff
@@ -1027,6 +1112,45 @@ func (client *Client) integrationAlreadyInstalled(name string) bool {
 	return false
 }
 
+// buildEnvelopeFromEvent builds an envelope from an event.
+// This is used to send events via the envelope-based transport API.
+func (client *Client) buildEnvelopeFromEvent(event *Event) *protocol.Envelope {
```
giortzisg, line 1117: this is also not needed, the adapter should suffice.

client.go
```diff
@@ -907,6 +986,12 @@ func (client *Client) processEvent(event *Event, hint *EventHint, scope EventMod
 			debuglog.Println("Event dropped: telemetry buffer full or unavailable")
 		}
 	} else {
+		// For Spotlight: build and send envelope for enhanced data collection
+		if client.options.Spotlight {
+			envelope := client.buildEnvelopeFromEvent(event)
+			client.Transport.SendEnvelope(envelope)
+		}
+		// Default path: send event directly for backwards compatibility
```
giortzisg, line 994: the transport is already wrapped with spotlight, so that part is not needed. the bugbot comments are valid.

transport.go
```diff
@@ -887,6 +929,271 @@ func (noopTransport) FlushWithContext(context.Context) bool {
...
+
+// Flush waits for the underlying transport to send pending events to Sentry.
+// In-flight Spotlight sends are best-effort and not waited on.
+func (st *SpotlightTransport) Flush(timeout time.Duration) bool {
+	return st.underlying.Flush(timeout)
+}
+
+func (st *SpotlightTransport) FlushWithContext(ctx context.Context) bool {
+	return st.underlying.FlushWithContext(ctx)
+}
```
giortzisg, line 1176: this probably doesn't work with an empty dsn. the noop transport returns from flush immediately so no events would be delivered, we also need to wait for the spotlight events.

client.go
```diff
@@ -365,6 +398,47 @@ func NewClient(options ClientOptions) (*Client, error) {
...
+	if options.Spotlight {
+		if options.SampleRate != 1.0 {
+			debuglog.Printf("Overriding SampleRate from %.2f to 1.0 for Spotlight", options.SampleRate)
+			options.SampleRate = 1.0
+		}
+		if !options.SendDefaultPII {
+			debuglog.Println("Enabling SendDefaultPII for Spotlight")
+			options.SendDefaultPII = true
+		}
+	}
```
giortzisg, line 440: We shouldn't override it globally, because this would affect the sample rate and pii of the default sentry client

giortzisg (changes requested):

giortzisg:
Hey @DAcodedBEAT any updates on this?

giortzisg:
Hey @DAcodedBEAT, I feel that the PR reached a unnecessarily complicated state. I see that this includes changes to the scheduler and has backoff implementation, but I don't feel like these are necessary.  Spotlight should just do the http call and then call the transport. Is there a reason we introduced these? 

I feel that as an initial implementation just a small isolated worker going through the requests should be good enough.

PR #1089, opened 2025-09-19
feat: add TraceIgnoreStatusCodes option

### LINKED ISSUE #1088: Change default behavior to stop tracing 404s unless explicitly enabled
* Default excludes 404s
* Breaking change noted in release notes + migration path documented

giortzisg:
## Description

Add client option for ignoring traces with specific status codes.

## Issues

- Closes GO-81
- Closes #1088

### DIFF
--- a/client.go
+++ b/client.go
@@ -228,6 +228,13 @@ type ClientOptions struct {
 	Tags map[string]string
 	// EnableLogs controls when logs should be emitted.
 	EnableLogs bool
+	// TraceIgnoreStatusCodes is a list of HTTP status codes that should not be traced.
+	// When an HTTP request results in one of these status codes, the transaction will
+	// not be sent to Sentry.
+	//
+	// This is useful for ignoring expected errors like 404s.
+	// By default, this is empty and all status codes are traced.
+	TraceIgnoreStatusCodes []int
 }
 
 // Client is the underlying processor that is used by the main API and Hub
--- a/tracing.go
+++ b/tracing.go
@@ -347,6 +347,51 @@ func (s *Span) SetDynamicSamplingContext(dsc DynamicSamplingContext) {
 	}
 }
 
+// shouldIgnoreStatusCode checks if the transaction should be ignored based on HTTP status code
+func (s *Span) shouldIgnoreStatusCode(hub *Hub) bool {
+	if !s.IsTransaction() {
+		return false
+	}
+
+	client := hub.Client()
+	if client == nil {
+		return false
+	}
+
+	ignoreStatusCodes := client.options.TraceIgnoreStatusCodes
+	if len(ignoreStatusCodes) == 0 {
+		return false
+	}
+
+	s.mu.Lock()
+	statusCodeData, exists := s.Data["http.response.status_code"]
+	s.mu.Unlock()
+
+	if !exists {
+		return false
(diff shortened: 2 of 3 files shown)


lcian (approved):
Implementation looks good to me.

tracing.go
```diff
@@ -347,6 +347,44 @@ func (s *Span) SetDynamicSamplingContext(dsc DynamicSamplingContext) {
...
+		return false
+	}
+
+	ignoreStatusCodes := client.options.TraceIgnoreStatusCodes
+	if len(ignoreStatusCodes) == 0 {
+		return false
+	}
+
+	s.mu.Lock()
+	statusCodeData, exists := s.Data["http.response.status_code"]
```
cleptric, line 362: I don't think it's a good idea to implement it like this. I would rather check for the status code in our integrations and drop the transaction there.

giortzisg: Thought of doing it for every integration as well, but this is what we expect to get as the status code, so it made sense at least for me to have it centralized somewhere. WDYT?

cleptric: My main issue is that we couple this with the presence of an attribute, which feels not right. Focusing on the places where the SDK automatically generates traces is how we did it in other SDKs.

echo/sentryecho.go
```diff
@@ -24,10 +24,21 @@ const (
...
 
+// shouldIgnoreStatus checks if the given HTTP status code should be ignored
+func (h *handler) shouldIgnoreStatus(status int) bool {
+	for _, ignoredCode := range h.traceIgnoreStatusCodes {
+		if status == ignoredCode {
+			return true
+		}
+	}
+	return false
+}
```
cleptric: Do we need to add this to each integration?

fasthttp/sentryfasthttp.go
```diff
@@ -86,6 +105,11 @@ func (h *Handler) Handle(handler fasthttp.RequestHandler) fasthttp.RequestHandle
 			status := ctx.Response.StatusCode()
 			transaction.Status = sentry.HTTPtoSpanStatus(status)
 			transaction.SetData("http.response.status_code", status)
+
+			if h.shouldIgnoreStatus(status) {
+				return
```
cleptric: Set it to not sampled.

echo/sentryecho.go
giortzisg: It seems to me like an integration level option, not a global one.

cleptric (commented):
Missing the debug log from the spec.

client.go
```diff
@@ -228,6 +228,13 @@ type ClientOptions struct {
 	Tags map[string]string
 	// EnableLogs controls when logs should be emitted.
 	EnableLogs bool
+	// TraceIgnoreStatusCodes is a list of HTTP status codes that should not be traced.
+	// When an HTTP request results in one of these status codes, the transaction will
+	// not be sent to Sentry.
+	//
+	// This is useful for ignoring expected errors like 404s.
```
cleptric:
```suggestion
```

cleptric (approved):
Please also add the new option to https://docs.sentry.io/platforms/go/configuration/options/ and https://docs.sentry.io/platforms/go/configuration/filtering/.

Issue #1091, opened 2025-09-24
Implement generic transport buffers


giortzisg:
For the new telemetry architecture, a generic buffer should be implemented, to hold the envelopes created from the event scheduler.

PR #1093, opened 2025-09-24
feat: add ring buffers

giortzisg:
### Description

Add generic ring buffer implementation with metrics functionality.

#### Issues
* resolves: #1091 
* resolves: GO-83

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -40,7 +40,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,10 +18,10 @@ jobs:
     name: Lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/setup-go@v5
+      - uses: actions/setup-go@v6
         with:
           go-version: "1.24"
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: golangci-lint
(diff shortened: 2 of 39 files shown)


internal/telemetry/types.go
```diff
@@ -0,0 +1,75 @@
...
+	case PriorityLow:
+		return "low"
+	case PriorityLowest:
+		return "lowest"
+	default:
+		return "unknown"
+	}
+}
+
+type DataCategory string
```
sl0thentr0py: in many SDKs and relay, we use the term 'data category' to mean these
https://github.com/getsentry/sentry-go/blob/de5b70a7b3c87af5459bcc2287d9780685e5c0d5/internal/ratelimit/category.go#L15-L25

so I'd either call this something else or unify this with those types. Unifying might make more sense long term since those data categories are also what relay uses for rate limiting and then each type of telemetry would work seamlessly with those rate limits.

internal/telemetry/types.go
```diff
@@ -0,0 +1,75 @@
...
+	return string(dc)
+}
+
+func (dc DataCategory) GetPriority() Priority {
+	switch dc {
+	case DataCategoryError:
+		return PriorityCritical
+	case DataCategoryCheckIn:
+		return PriorityHigh
+	case DataCategoryLog:
```
sl0thentr0py: can you link the spec for this? intuitively log should be the lowest since it's the most high volume

sl0thentr0py (changes requested):
initial comment about the `DataCategory`, will review buffer in a bit

internal/telemetry/types.go
giortzisg: https://develop.sentry.dev/sdk/miscellaneous/telemetry-buffer/

internal/telemetry/types.go
giortzisg: makes sense to merge them

Issue #1092, opened 2025-09-24
Create new transport that accepts envelopes


giortzisg:
For the telemetry refactoring, we need to migrate from the old async transport that accepts events, to a new one that accepts envelopes.

PR #1094, opened 2025-09-24
feat: add new envelope transport

giortzisg:
### Description

Create the new transport that accepts envelopes and not events. For now only the implementation for the new transport is added, without deprecating the old one. The PR also includes some misc changes: 
- creating an envelope convertible interface, to accept a generic type of events/logs/transactions, etc..
- moving the dsn internally under the protocol folder and re exporting everything on the top level to avoid cyclic imports when using the new transport.

#### Issues
* resolves: #1092 
* resolves: GO-84

### DIFF
--- a/dsn.go
+++ b/dsn.go
@@ -2,221 +2,44 @@ package sentry
 
 import (
 	"encoding/json"
-	"fmt"
-	"net/url"
-	"strconv"
-	"strings"
-	"time"
+
+	"github.com/getsentry/sentry-go/internal/protocol"
 )
 
-type scheme string
+// Re-export protocol types to maintain public API compatibility
 
-const (
-	schemeHTTP  scheme = "http"
-	schemeHTTPS scheme = "https"
-)
-
-func (scheme scheme) defaultPort() int {
-	switch scheme {
-	case schemeHTTPS:
-		return 443
-	case schemeHTTP:
-		return 80
-	default:
-		return 80
-	}
+// Dsn is used as the remote address source to client transport.
+type Dsn struct {
+	*protocol.Dsn
 }
 
 // DsnParseError represents an error that occurs if a Sentry
 // DSN cannot be parsed.
-type DsnParseError struct {
-	Message string
-}
-
-func (e DsnParseError) Error() string {
-	return "[Sentry] DsnParseError: " + e.Message
-}
-
-// Dsn is used as the remote address source to client transport.
-type Dsn struct {
-	scheme    scheme
-	publicKey string
-	secretKey string
-	host      string
-	port      int
-	path      string
-	projectID string
-}
+type DsnParseError = protocol.DsnParseError
 
 // NewDsn creates a Dsn by parsing rawURL. Most users will never call this
 // function directly. It is provided for use in custom Transport
 // implementations.
 func NewDsn(rawURL string) (*Dsn, error) {
-	// Parse
-	parsedURL, err := url.Parse(rawURL)
+	protocolDsn, err := protocol.NewDsn(rawURL)
 	if err != nil {
-		return nil, &DsnParseError{fmt.Sprintf("invalid url: %v", err)}
+		return nil, err
 	}
-
-	// Scheme
-	var scheme scheme
-	switch parsedURL.Scheme {
-	case "http":
-		scheme = schemeHTTP
-	case "https":
-		scheme = schemeHTTPS
-	default:
-		return nil, &DsnParseError{"invalid scheme"}
-	}
-
-	// PublicKey
-	publicKey := parsedURL.User.Username()
-	if publicKey == "" {
-		return nil, &DsnParseError{"empty username"}
-	}
-
-	// SecretKey
-	var secretKey string
-	if parsedSecretKey, ok := parsedURL.User.Password(); ok {
-		secretKey = parsedSecretKey
-	}
-
-	// Host
-	host := parsedURL.Hostname()
-	if host == "" {
-		return nil, &DsnParseError{"empty host"}
-	}
-
-	// Port
-	var port int
-	if p := parsedURL.Port(); p != "" {
-		port, err = strconv.Atoi(p)
-		if err != nil {
-			return nil, &DsnParseError{"invalid port"}
-		}
-	} else {
-		port = scheme.defaultPort()
-	}
-
-	// ProjectID
-	if parsedURL.Path == "" || parsedURL.Path == "/" {
-		return nil, &DsnParseError{"empty project id"}
-	}
-	pathSegments := strings.Split(parsedURL.Path[1:], "/")
-	projectID := pathSegments[len(pathSegments)-1]
-
-	if projectID == "" {
-		return nil, &DsnParseError{"empty project id"}
-	}
-
-	// Path
-	var path string
-	if len(pathSegments) > 1 {
-		path = "/" + strings.Join(pathSegments[0:len(pathSegments)-1], "/")
-	}
-
-	return &Dsn{
-		scheme:    scheme,
-		publicKey: publicKey,
-		secretKey: secretKey,
-		host:      host,
-		port:      port,
-		path:      path,
-		projectID: projectID,
-	}, nil
+	return &Dsn{Dsn: protocolDsn}, nil
 }
 
-// String formats Dsn struct into a valid string url.
-func (dsn Dsn) String() string {
-	var url string
-	url += fmt.Sprintf("%s://%s", dsn.scheme, dsn.publicKey)
-	if dsn.secretKey != "" {
-		url += fmt.Sprintf(":%s", dsn.secretKey)
-	}
-	url += fmt.Sprintf("@%s", dsn.host)
-	if dsn.port != dsn.scheme.defaultPort() {
-		url += fmt.Sprintf(":%d", dsn.port)
-	}
-	if dsn.path != "" {
-		url += dsn.path
-	}
-	url += fmt.Sprintf("/%s", dsn.projectID)
-	return url
-}
-
-// Get the scheme of the DSN.
-func (dsn Dsn) GetScheme() string {
-	return string(dsn.scheme)
-}
-
-// Get the public key of the DSN.
-func (dsn Dsn) GetPublicKey() string {
-	return dsn.publicKey
-}
-
-// Get the secret key of the DSN.
-func (dsn Dsn) GetSecretKey() string {
-	return dsn.secretKey
-}
-
-// Get the host of the DSN.
-func (dsn Dsn) GetHost() string {
-	return dsn.host
-}
-
-// Get the port of the DSN.
-func (dsn Dsn) GetPort() int {
-	return dsn.port
-}
-
-// Get the path of the DSN.
-func (dsn Dsn) GetPath() string {
-	return dsn.path
-}
-
-// Get the project ID of the DSN.
-func (dsn Dsn) GetProjectID() string {
-	return dsn.projectID
-}
-
-// GetAPIURL returns the URL of the envelope endpoint of the project
-// associated with the DSN.
-func (dsn Dsn) GetAPIURL() *url.URL {
-	var rawURL string
-	rawURL += fmt.Sprintf("%s://%s", dsn.scheme, dsn.host)
-	if dsn.port != dsn.scheme.defaultPort() {
-		rawURL += fmt.Sprintf(":%d", dsn.port)
-	}
-	if dsn.path != "" {
-		rawURL += dsn.path
-	}
-	rawURL += fmt.Sprintf("/api/%s/%s/", dsn.projectID, "envelope")
-	parsedURL, _ := url.Parse(rawURL)
-	return parsedURL
-}
-
-// RequestHeaders returns all the necessary headers that have to be used in the transport when seinding events
+// RequestHeaders returns all the necessary headers that have to be used in the transport when sending events
 // to the /store endpoint.
 //
 // Deprecated: This method shall only be used if you want to implement your own transport that sends events to
 // the /store endpoint. If you're using the transport provided by the SDK, all necessary headers to authenticate
 // against the /envelope endpoint are added automatically.
-func (dsn Dsn) RequestHeaders() map[string]string {
-	auth := fmt.Sprintf("Sentry sentry_version=%s, sentry_timestamp=%d, "+
-		"sentry_client=sentry.go/%s, sentry_key=%s", apiVersion, time.Now().Unix(), SDKVersion, dsn.publicKey)
-
-	if dsn.secretKey != "" {
-		auth = fmt.Sprintf("%s, sentry_secret=%s", auth, dsn.secretKey)
-	}
-
-	return map[string]string{
-		"Content-Type":  "application/json",
-		"X-Sentry-Auth": auth,
-	}
+func (dsn *Dsn) RequestHeaders() map[string]string {
+	return dsn.Dsn.RequestHeaders(SDKVersion)
 }
 
 // MarshalJSON converts the Dsn struct to JSON.
-func (dsn Dsn) MarshalJSON() ([]byte, error) {
+func (dsn *Dsn) MarshalJSON() ([]byte, error) {
 	return json.Marshal(dsn.String())
 }
 
--- a/interfaces.go
+++ b/interfaces.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/getsentry/sentry-go/attribute"
+	"github.com/getsentry/sentry-go/internal/protocol"
 	"github.com/getsentry/sentry-go/internal/ratelimit"
 )
 
@@ -249,11 +250,11 @@ var sensitiveHeaders = map[string]struct{}{
 // NewRequest avoids operations that depend on network access. In particular, it
 // does not read r.Body.
 func NewRequest(r *http.Request) *Request {
-	protocol := schemeHTTP
+	prot := protocol.SchemeHTTP
 	if r.TLS != nil || r.Header.Get("X-Forwarded-Proto") == "https" {
-		protocol = schemeHTTPS
+		prot = protocol.SchemeHTTPS
 	}
-	url := fmt.Sprintf("%s://%s%s", protocol, r.Host, r.URL.Path)
+	url := fmt.Sprintf("%s://%s%s", prot, r.Host, r.URL.Path)
 
 	var cookies string
 	var env map[string]string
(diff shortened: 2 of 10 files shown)


dsn.go
```diff
@@ -1,233 +1,37 @@
...
-	case schemeHTTPS:
-		return 443
-	case schemeHTTP:
-		return 80
-	default:
-		return 80
-	}
+// Dsn is used as the remote address source to client transport.
+type Dsn struct {
+	protocol.Dsn
```
sl0thentr0py, line 11: not familiar with go, why embed instead of alias?

interfaces.go
```diff
@@ -485,6 +476,78 @@ func (e *Event) SetException(exception error, maxErrorDepth int) {
...
+
+	if dsn != nil {
+		header.Dsn = dsn.String()
+	}
+
+	header.Sdk = &e.Sdk
+
+	envelope := protocol.NewEnvelope(header)
+
+	eventBody, err := json.Marshal(e)
```
sl0thentr0py, line 455: this is fine for now, but eventually, each envelope item type should know how to serialize itself instead of this hacky logic
https://github.com/getsentry/sentry-go/blob/1ed184b879d4b7a9e282cfa5f97802376d60103d/interfaces.go#L558-L574

sl0thentr0py (approved):
ideally we'd also abstract the older event types as envelope items, but I'll leave it upto you if you want to do that now or separately later

interfaces.go
giortzisg: Agree with that, I didn't want to completely refactor everything for now, but the proper way would be for every event type to be each own interface implementation of a generic event.

dsn.go
giortzisg: I am extending the `RequestHeaders` method to pass the sdkVersion on the protocol level.

internal/http/transport.go
```diff
@@ -0,0 +1,642 @@
...
+	return http.ProxyFromEnvironment
+}
+
+func getTLSConfig(options TransportOptions) *tls.Config {
+	if options.CaCerts != nil {
+		// #nosec G402 -- We should be using `MinVersion: tls.VersionTLS12`,
+		// 				 but we don't want to break peoples code without the major bump.
+		return &tls.Config{
+			RootCAs: options.CaCerts,
+		}
```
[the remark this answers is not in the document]
cleptric: can we address this?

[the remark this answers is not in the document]
giortzisg: Since go uses tls1.2 we should be good to this as the min version. We already inherit the min version of the runtime anyways.

internal/http/transport.go
```diff
@@ -0,0 +1,505 @@
...
+}
+
+func (t *SyncTransport) SendEnvelope(envelope *protocol.Envelope) error {
+	return t.SendEnvelopeWithContext(context.Background(), envelope)
+}
+
+func (t *SyncTransport) Close() {}
+
+func (t *SyncTransport) SendEvent(event protocol.EnvelopeConvertible) {
+	if envelope, err := event.ToEnvelope(t.dsn); err == nil && envelope != nil {
```
lcian: If this errors, we should log it. We're not logging it right now. Same for other uses of this function.

internal/http/transport.go
```diff
@@ -0,0 +1,505 @@
...
+			RootCAs:    options.CaCerts,
+			MinVersion: tls.VersionTLS12,
+		}
+	}
+
+	return nil
+}
+
+func getSentryRequestFromEnvelope(ctx context.Context, dsn *protocol.Dsn, envelope *protocol.Envelope) (r *http.Request, err error) {
+	defer func() {
```
lcian, line 75: Cool!

internal/http/transport.go
```diff
@@ -0,0 +1,505 @@
...
+		case protocol.EnvelopeItemTypeTransaction:
+			return ratelimit.CategoryTransaction
+		case protocol.EnvelopeItemTypeCheckIn:
+			return ratelimit.CategoryMonitor
+		case protocol.EnvelopeItemTypeLog:
+			return ratelimit.CategoryLog
+		case protocol.EnvelopeItemTypeAttachment:
+			continue
+		default:
+			return ratelimit.CategoryAll
```
lcian, line 130: In theory an envelope can contain multiple items, here we're only considering the first we find a match for.

internal/http/transport.go
```diff
@@ -0,0 +1,505 @@
...
+
+func NewSyncTransport(options TransportOptions) *SyncTransport {
+	transport := &SyncTransport{
+		Timeout: defaultTimeout,
+		limits:  make(ratelimit.Map),
+	}
+
+	dsn, err := protocol.NewDsn(options.Dsn)
+	if err != nil {
+		debuglog.Printf("failed to create transport: invalid dsn: %v\n", err)
```
lcian:
```suggestion
		debuglog.Printf("transport is disabled: invalid dsn: %v\n", err)
```

internal/http/transport.go
```diff
@@ -0,0 +1,505 @@
...
+	}
+	response, err := t.client.Do(request)
+	if err != nil {
+		debuglog.Printf("There was an issue with sending an event: %v", err)
+		return err
+	}
+	if response.StatusCode >= 400 && response.StatusCode <= 599 {
+		b, err := io.ReadAll(response.Body)
+		if err != nil {
+			debuglog.Printf("Error while reading response code: %v", err)
```
lcian:
```suggestion
			debuglog.Printf("Error while reading response body: %v", err)
```

internal/http/transport.go
```diff
@@ -0,0 +1,505 @@
...
+			continue
+		default:
+			return ratelimit.CategoryAll
+		}
+	}
+
+	return ratelimit.CategoryAll
+}
+
+type SyncTransport struct {
```
lcian, line 148: Should we copy over some comments from the original transports here (in both)?

e.g. in the original transport we have this.
```
// ================================
// HTTPSyncTransport
// ================================

// HTTPSyncTransport is a blocking implementation of Transport.
//
// Clients using this transport will send requests to Sentry sequentially and
// block until a response is returned.
//
// The blocking behavior is useful in a limited set of use cases. For example,
// use it when deploying code to a Function as a Service ("Serverless")
// platform, where any work happening in a background goroutine is not
// guaranteed to execute.
//
// For most cases, prefer HTTPTransport.
```

And a bunch of comments around that explain certain stuff that is not immediately obvious. Some stuff is obvious but some is not.

As a first-time reader of this code, I appreciate those comments as they make it much easier for me to understand what I'm looking at.
Also at least top-level comments like the example I made above could be useful for end-users as well.

internal/http/transport.go
```diff
@@ -0,0 +1,505 @@
...
+		done:      make(chan struct{}),
+		limits:    make(ratelimit.Map),
+	}
+
+	transport.queue = make(chan *protocol.Envelope, transport.QueueSize)
+	transport.flushRequest = make(chan chan struct{})
+
+	dsn, err := protocol.NewDsn(options.Dsn)
+	if err != nil {
+		debuglog.Printf("%v\n", err)
```
lcian: `l`: please adjust

internal/http/transport.go
```diff
@@ -0,0 +1,505 @@
...
+	case t.queue <- envelope:
+		return nil
+	default:
+		atomic.AddInt64(&t.droppedCount, 1)
+		return ErrTransportQueueFull
+	}
+}
+
+func (t *AsyncTransport) SendEvent(event protocol.EnvelopeConvertible) {
+	if envelope, err := event.ToEnvelope(t.dsn); err == nil && envelope != nil {
```
lcian: same as above, we should log the failure

internal/http/transport.go
```diff
@@ -0,0 +1,505 @@
...
+
+	if response.StatusCode >= 400 && response.StatusCode < 500 {
+		if body, err := io.ReadAll(io.LimitReader(response.Body, maxDrainResponseBytes)); err == nil {
+			debuglog.Printf("Client error %d: %s", response.StatusCode, string(body))
+		}
+		return false
+	}
+
+	if response.StatusCode >= 500 {
+		debuglog.Printf("Server error %d - will retry", response.StatusCode)
```
lcian: > will retry

Chat, is this real? I believe we just fail to send and that's it.

internal/http/transport.go
lcian: Also I'm thinking this should live in `internal/ratelimit/category.go`, if it's possible with the imports

internal/http/transport.go
```diff
@@ -0,0 +1,505 @@
...
+		atomic.AddInt64(&t.sentCount, 1)
+	} else {
+		atomic.AddInt64(&t.errorCount, 1)
+	}
+}
+
+func (t *AsyncTransport) sendEnvelopeHTTP(envelope *protocol.Envelope) bool {
+	category := categoryFromEnvelope(envelope)
+	if t.isRateLimited(category) {
+		return false
```
lcian, line 476: This way we consider being rate limited an error, see above where we do
`		atomic.AddInt64(&t.errorCount, 1)`
if this fun returns false.
Is this what we want?

I guess this stuff will be replaced with client reports anyways so it's also okay to leave it as is for now.

internal/protocol/interfaces.go
```diff
@@ -0,0 +1,41 @@
...
+// TelemetryTransport represents the envelope-first transport interface.
+// This interface is designed for the telemetry buffer system and provides
+// non-blocking sends with backpressure signals.
+type TelemetryTransport interface {
+	// SendEnvelope sends an envelope to Sentry. Returns immediately with
+	// backpressure error if the queue is full.
+	SendEnvelope(envelope *Envelope) error
+
+	// SendEvent sends an event to Sentry. Returns immediately with
+	// backpressure error if the queue is full.
```
lcian: false, it cannot error

internal/protocol/envelope.go
```diff
@@ -0,0 +1,213 @@
...
+	// Dsn can be used for self-authenticated envelopes.
+	// This means that the envelope has all the information necessary to be sent to sentry.
+	// In this case the full DSN must be stored in this key.
+	Dsn string `json:"dsn,omitempty"`
+
+	// Sdk carries the same payload as the sdk interface in the event payload but can be carried for all events.
+	// This means that SDK information can be carried for minidumps, session data and other submissions.
+	Sdk *SdkInfo `json:"sdk,omitempty"`
+
+	// Trace contains trace context information for distributed tracing
```
lcian:
```suggestion
	// Trace contains the [Dynamic Sampling Context](https://develop.sentry.dev/sdk/telemetry/traces/dynamic-sampling-context/)
```

internal/protocol/envelope.go
```diff
@@ -0,0 +1,213 @@
...
+// Size returns the total size of the envelope when serialized.
+func (e *Envelope) Size() (int, error) {
+	data, err := e.Serialize()
+	if err != nil {
+		return 0, err
+	}
+	return len(data), nil
+}
+
+// MarshalJSON converts the EnvelopeHeader to JSON and ensures it's a single line.
```
lcian: > ensures it's a single line

does it?

lcian (commented):
Left some mostly minor comments, looks good overall. Haven't looked at the tests yet.

internal/http/transport.go
lcian: Hmm maybe I'm wrong on the first comment, because I think there can be only one top-level item.

giortzisg: Currently relay can only parse envelope items of the same type, as you also mentioned in the last comment. Also IIRC, the idea of different envelope item types was mentioned and got turned down, because it would have resulted in slower parsing for different items.

internal/http/transport.go
giortzisg: Not real :joy:

internal/http/transport.go
giortzisg: Well the error behavior is only used for logging purposes currently. I wanted to differentiate an actual error with a ratelimit that's why the impl returns false. But yeah client reports will cover this case.

lcian: Cool

internal/http/transport.go
lcian: Alright, thanks for confirming.

interfaces.go
```diff
@@ -485,6 +476,78 @@ func (e *Event) SetException(exception error, maxErrorDepth int) {
...
+	}
+
+	envelope.AddItem(mainItem)
+	for _, attachment := range e.Attachments {
+		attachmentItem := protocol.NewAttachmentItem(attachment.Filename, attachment.ContentType, attachment.Payload)
+		envelope.AddItem(attachmentItem)
+	}
+
+	return envelope, nil
+}
```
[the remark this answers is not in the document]
giortzisg: maybe you want to have a better look at that LLM.

internal/http/transport.go
```diff
@@ -0,0 +1,548 @@
...
+	}
```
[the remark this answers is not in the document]
giortzisg: https://github.com/getsentry/sentry-go/pull/1094#discussion_r2410781132

transport.go
```diff
@@ -743,3 +745,143 @@ func (noopTransport) FlushWithContext(context.Context) bool {
...
+// This is not yet exposed in the public API and is for internal experimentation.
+func NewInternalSyncTransport() Transport {
+	return &internalSyncTransportAdapter{}
+}
+
+// internalSyncTransportAdapter wraps the internal SyncTransport to implement
+// the root-level Transport interface, avoiding cyclic imports.
+type internalSyncTransportAdapter struct {
+	transport protocol.TelemetryTransport
+	dsn       *protocol.Dsn
```
lcian: We don't need this I think

internal/http/transport_test.go
```diff
@@ -0,0 +1,823 @@
...
+			QueueSize: 2,
+			Timeout:   defaultTimeout,
+			done:      make(chan struct{}),
+			limits:    make(ratelimit.Map),
+			dsn:       dsn,
+			transport: &http.Transport{},
+			client:    &http.Client{Timeout: defaultTimeout},
+		}
+		// manually set the queue size to simulate overflow
+		transport.queue = make(chan *protocol.Envelope, transport.QueueSize)
```
lcian, line 222: Needed? We already se the QueueSize above

internal/http/transport_test.go
```diff
@@ -0,0 +1,823 @@
...
+			tr := NewAsyncTransport(TransportOptions{
+				Dsn: "http://key@" + server.URL[7:] + "/123",
+			})
+			transport, ok := tr.(*AsyncTransport)
+			if !ok {
+				t.Fatalf("expected *AsyncTransport, got %T", tr)
+			}
+			defer transport.Close()
+
+			transport.SendEvent(tt.event)
```
lcian, line 293: What are we testing here? That it doesn't panic?

giortzisg: Yes, we already test that envelopes are sent on `TestAsyncTransport_SendEnvelope`, so the only thing that can go wrong here is the envelope conversion.

internal/http/transport_test.go
giortzisg: we need to still create the buffered channel.

internal/http/transport_test.go
lcian: Do we have a way to assert it succeeds? Because it doesn't seem we're asserting anything

internal/http/transport.go
```diff
@@ -0,0 +1,571 @@
...
+	default:
+		atomic.AddInt64(&t.droppedCount, 1)
+		return ErrTransportQueueFull
+	}
+}
+
+func (t *AsyncTransport) SendEvent(event protocol.EnvelopeConvertible) {
+	envelope, err := event.ToEnvelope(t.dsn)
+	if err != nil {
+		debuglog.Printf("Failed to convert to envelope: %v", err)
```
aldy505, line 381:
```suggestion
		debuglog.Printf("Failed to convert to envelope: %s", err.Error())
```

using `%v` is heavier performance-wise than `%s`

internal/http/transport.go
```diff
@@ -0,0 +1,571 @@
...
+		return
+	}
+
+	if envelope == nil {
+		debuglog.Printf("Error: event with empty envelope")
+		return
+	}
+
+	if err := t.SendEnvelope(envelope); err != nil {
+		debuglog.Printf("Error sending the envelope: %v", err)
```
aldy505, line 391:
```suggestion
		debuglog.Printf("Error sending the envelope: %s", err.Error())
```
same goes with this

internal/protocol/dsn.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// NewDsn creates a Dsn by parsing rawURL. Most users will never call this
+// function directly. It is provided for use in custom Transport
+// implementations.
+func NewDsn(rawURL string) (*Dsn, error) {
+	// Parse
+	parsedURL, err := url.Parse(rawURL)
+	if err != nil {
+		return nil, &DsnParseError{fmt.Sprintf("invalid url: %v", err)}
```
aldy505, line 61: Why don't we do something like this?
```suggestion
		return nil, &DsnParseError{fmt.Errorf("invalid url: %w", err)}
```
That way, we can trace what the actual inner error is.

This would modify `DsnParseError` to:
```go
type DsnParseError struct {
	Message error
}
```

internal/protocol/dsn.go
```diff
@@ -0,0 +1,236 @@
...
+	case "https":
+		scheme = SchemeHTTPS
+	default:
+		return nil, &DsnParseError{"invalid scheme"}
+	}
+
+	// PublicKey
+	publicKey := parsedURL.User.Username()
+	if publicKey == "" {
+		return nil, &DsnParseError{"empty username"}
```
aldy505, line 78: `username` is confusing, users only understand "public key" and "project id"
```suggestion
		return nil, &DsnParseError{"empty public key"}
```

internal/protocol/dsn.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// String formats Dsn struct into a valid string url.
+func (dsn Dsn) String() string {
+	var url string
+	url += fmt.Sprintf("%s://%s", dsn.scheme, dsn.publicKey)
+	if dsn.secretKey != "" {
+		url += fmt.Sprintf(":%s", dsn.secretKey)
+	}
+	url += fmt.Sprintf("@%s", dsn.host)
```
aldy505, line 139: Annoying request, but: What about self-hosted users that uses private IPv6 address on their DSN? 😀

lcian: Why is this problematic for IPv6?

internal/http/transport_test.go
giortzisg: You mean to actually test the event to envelope transformation? Or if the event was sent. We can check that the event was received as we also do on the envelope implementation, but I didn't do this, since we only do `event.ToEnvelope` and call the other function.

internal/protocol/dsn.go
giortzisg: I don't really want to break the signature for message, since it doesn't really give that much value.

internal/http/transport.go
giortzisg: for errors I think this doesn't really apply, both just call `err.Error()` IIRC.

internal/protocol/dsn.go
aldy505: For IPv6, you would need to enclose it within `[]` brackets. Or maybe just create a test case for this one.

giortzisg: Let's have a look at this on another PR, don't want to introduce another behavioral change here.

Issue #1104, opened 2025-10-06
Create a package for debug logger and move it internally


giortzisg:
Any packages that require the logger for debug statements, that are under internal might create cyclic imports with the current implementation. We should move the debug logger under it's own package and configure it in a way to be more easily accessible.

PR #1105, opened 2025-10-06
ref: create debuglog package

giortzisg:
### Description
Create a `debuglog` package to easily access the `sentry.DebugLogger` without depending on the root level. This change aims to remove any cyclic imports when internal packages are used on the top level, but also want to access the debug logger. 

#### Issues
* resolves: #1104 
* resolves: GO-88

### DIFF
--- a/internal/debuglog/log.go
+++ b/internal/debuglog/log.go
@@ -7,13 +7,9 @@ import (
 )
 
 var (
-	logger *log.Logger
+	logger = log.New(io.Discard, "[Sentry] ", log.LstdFlags)
 	mu     sync.RWMutex
 )
-
-func init() {
-	logger = log.New(io.Discard, "[Sentry] ", log.LstdFlags)
-}
 
 // SetLogger replaces the current debug logger with a new one.
 // This function is thread-safe and can be called concurrently.
--- a/internal/debuglog/log.go
+++ b/internal/debuglog/log.go
@@ -1,0 +1,69 @@
+package debuglog
+
+import (
+	"io"
+	"log"
+	"sync"
+)
+
+var (
+	logger *log.Logger
+	mu     sync.RWMutex
+)
+
+func init() {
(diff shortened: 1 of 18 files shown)


internal/debuglog/log.go
```diff
@@ -0,0 +1,69 @@
...
+// This function is thread-safe and can be called concurrently.
+func SetLogger(l *log.Logger) {
+	mu.Lock()
+	defer mu.Unlock()
+	logger = l
+}
+
+func SetOutput(w io.Writer) {
+	logger.SetOutput(w)
+}
```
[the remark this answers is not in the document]
giortzisg: The *log.Logger implementation is thread-safe in itself, so no need for mutex.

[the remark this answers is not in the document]
lcian: `h`: Comment from Cursor seems correct, if we're locking in `GetLogger` I don't see why we shouldn't be locking here.

internal/debuglog/log.go
```diff
@@ -0,0 +1,65 @@
+package debuglog
```
lcian, line 1: `l`: I notice we already have a `debug` package, so maybe we can put it there? But feel free to choose whichever you prefer.

lcian (approved):
LGTM in principle and I understand the problem this solves.
I think it's a bit unfortunate that we have to lock on every message, and I wonder if there's a different way to do it without locking.
Anyways, I would assume that customers don't enable `Debug: true` in production, so it shouldn't be a big deal in practice.

internal/debuglog/log.go
giortzisg: Wanted to keep the signature clean so that we can use the convenience methods like `debuglog.Print`, etc. Also makes more sense to me to be it's own package.

lcian: Ah yeah the point about the signature makes sense.

Issue #1095, opened 2025-09-26
Create envelope scheduler


giortzisg:
Create an envelope scheduler according to [https://develop.sentry.dev/sdk/miscellaneous/telemetry-buffer/](https://develop.sentry.dev/sdk/miscellaneous/telemetry-buffer/). <br>The scheduler should take buffers with events (logs, errors, transactions, etc) and prioritize how to convert them to envelopes and push them to queues based on a weighted round robin algorithm.

PR #1107, opened 2025-10-07
feat: add telemetry scheduler

giortzisg:
### Description

Add telemetry scheduler implementation.

#### Issues

* resolves: #1095
* resolves: GO-85


internal/protocol/envelope.go
```diff
@@ -211,3 +211,40 @@ func NewLogItem(itemCount int, payload []byte) *EnvelopeItem {
...
+		return nil, fmt.Errorf("cannot create envelope from empty items")
+	}
+
+	firstItem := items[0]
+
+	header := &EnvelopeHeader{
+		EventID: firstItem.GetEventID(),
+		SentAt:  time.Now(),
+		Trace:   firstItem.GetDynamicSamplingContext(),
+	}
```
giortzisg: Not sure if this works would work in the future for batching spans. For logs it's ok, since we get the trace id from each log item.

lcian: If we would send the DSC as span attributes then you would be able to batch spans in any way you want, not necessarily group them by trace id always.
@cleptric is this planned?

internal/protocol/envelope.go
```diff
@@ -211,3 +211,40 @@ func NewLogItem(itemCount int, payload []byte) *EnvelopeItem {
...
+// For log items, it will create a single batched log envelope item.
+func CreateEnvelopeFromItems(items []EnvelopeItemConvertible, dsn *Dsn) (*Envelope, error) {
+	if len(items) == 0 {
+		return nil, fmt.Errorf("cannot create envelope from empty items")
+	}
+
+	firstItem := items[0]
+
+	header := &EnvelopeHeader{
+		EventID: firstItem.GetEventID(),
```
lcian: I'm not sure how it happens but when we send logs this is empty string so we get 400 from Relay, please fix it.

internal/protocol/envelope.go
cleptric: No plans at the moment. The main reason we batch spans by trace id is to enforce a certain traffic pattern in the ingestion segment buffer when it comes to insertion performance.

interfaces.go
```diff
@@ -714,6 +731,56 @@ type Log struct {
...
+	logData, err := json.Marshal(l)
+	if err != nil {
+		return nil, err
+	}
+	return &protocol.EnvelopeItem{
+		Header: &protocol.EnvelopeItemHeader{
+			Type: protocol.EnvelopeItemTypeLog,
+		},
+		Payload: logData,
+	}, nil
```
lcian: When would you ever call this? If I understand correctly this should be unreachable in normal code.

lcian: I see it's used in tests, and the usage there seems questionable, it's not clear to me why this is needed.

internal/telemetry/scheduler.go
```diff
@@ -0,0 +1,278 @@
...
+	for _, buffer := range s.buffers {
+		if !buffer.IsEmpty() {
+			return true
+		}
+	}
+	return false
+}
+
+func (s *Scheduler) processNextBatch() {
+	s.mu.Lock()
```
lcian: Do we even need a mutex? Is it even possible for this function to be called concurrently by 2 different goroutines?

internal/protocol/log_batch.go
```diff
@@ -0,0 +1,78 @@
...
+
+// LogAttribute is the JSON representation for a single log attribute value.
+type LogAttribute struct {
+	Value any    `json:"value"`
+	Type  string `json:"type"`
+}
+
+// LogItem represents the serialized shape of a single log record inside a batched
+// log envelope item. Keep in sync with sentry.Log fields that are meant to be emitted.
+type LogItem struct {
```
lcian: I don't understand why this is needed at all, it's the same as the `Log` type but with `TraceID` and `Level` changed to `string`.
It seems to me it would be possible to use `Log` directly, we would just need to implement `MarshalText` on `Level`.
`TraceID` already implements `MarshalText`.
`Attributes` is the same as in the `Log` struct but with different names for the types it seems.

internal/telemetry/scheduler.go
```diff
@@ -0,0 +1,278 @@
...
+	default:
+		// if the buffers are properly configured, buffer.PollIfReady should return a single item for every category
+		// other than logs. We still iterate over the items just in case, because we don't want to send broken envelopes.
+		for _, it := range items {
+			s.sendItems(it)
+		}
+	}
+}
+
+func (s *Scheduler) sendItems(item protocol.EnvelopeItemConvertible) {
```
lcian:
```suggestion
func (s *Scheduler) sendItem(item protocol.EnvelopeItemConvertible) {
```

internal/telemetry/bucketed_buffer.go
```diff
@@ -0,0 +1,398 @@
...
+	}
+	b.traceIndex = make(map[string]int)
+	b.head = 0
+	b.tail = 0
+	b.totalItems = 0
+	b.bucketCount = 0
+	return res
+}
+
+func (b *BucketedBuffer[T]) Peek() (T, bool) {
```
lcian, line 297: where do we use this?

interfaces.go
giortzisg: nice catch, it was a remnant from the initial implementation, when I was converting single item logs to envelope items.

internal/telemetry/bucketed_buffer.go
giortzisg: On the scheduler we don't, might be useful at some point though, that's why I kept it.

interfaces.go
giortzisg: To add a log on the buffer the item needs to implement `ToEnvelopeItem`, but you are correct in the sense that this should be used when creating the log items payload. Changed the implemenetation

interfaces.go
```diff
@@ -731,34 +731,47 @@ type Log struct {
...
-	if err != nil {
-		return nil, err
+	type logJSON struct {
+		Timestamp  *float64                         `json:"timestamp,omitempty"`
+		TraceID    string                           `json:"trace_id,omitempty"`
+		Level      string                           `json:"level"`
+		Severity   int                              `json:"severity_number,omitempty"`
+		Body       string                           `json:"body,omitempty"`
+		Attributes map[string]protocol.LogAttribute `json:"attributes,omitempty"`
+	}
```
lcian, line 743: Hmm I see the reason why we need this type now, e.g. `Timestamp` is a `time.Time` but you want to serialize it as `float64`, etc.
In Rust (serde) you could do this using `serialize_with`: https://serde.rs/field-attrs.html#serialize_with 😎

internal/telemetry/scheduler.go
```diff
@@ -256,7 +242,7 @@ func (s *Scheduler) sendItems(item protocol.EnvelopeItemConvertible) {
 	envelope := protocol.NewEnvelope(header)
 	envItem, err := item.ToEnvelopeItem()
 	if err != nil {
-		debuglog.Printf("error converting item: %v", err)
+		debuglog.Printf("error converting item to envelope: %v", err)
```
lcian:
```suggestion
		debuglog.Printf("error converting item to envelope item: %v", err)
```
or 
```suggestion
		debuglog.Printf("error converting item: %v", err)
```
I guess

internal/telemetry/buffer_wrapper.go
```diff
@@ -0,0 +1,61 @@
...
+	"time"
+
+	"github.com/getsentry/sentry-go/internal/protocol"
+	"github.com/getsentry/sentry-go/internal/ratelimit"
+)
+
+// Buffer is the top-level buffer that wraps the scheduler and category buffers.
+type Buffer struct {
+	scheduler *Scheduler
+	storage   map[ratelimit.Category]Storage[protocol.EnvelopeItemConvertible]
```
sl0thentr0py: scheduler already has `buffers` inside, why the `storage` here as well?

client.go
```diff
@@ -391,6 +401,41 @@ func (client *Client) setupTransport() {
...
+		Dsn:           client.options.Dsn,
+		HTTPClient:    client.options.HTTPClient,
+		HTTPTransport: client.options.HTTPTransport,
+		HTTPProxy:     client.options.HTTPProxy,
+		HTTPSProxy:    client.options.HTTPSProxy,
+		CaCerts:       client.options.CaCerts,
+	})
+	client.Transport = &internalAsyncTransportAdapter{transport: transport}
+
+	storage := map[ratelimit.Category]telemetry.Storage[protocol.EnvelopeItemConvertible]{
```
sl0thentr0py, line 442: storage could be initialized directly inside the scheduler buffers, don't see why we have to expose it?

sl0thentr0py (approved):

Issue #1117, opened 2025-10-23
Slog integration only attaches to trace when using slog.***Context()


Gitznik:
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

0.35.3

### Go version

1.25.2

### Are you using Go Modules?

yes

### Steps to reproduce

```go
package logger

import (
	"context"
	"log/slog"

	sentryslog "github.com/getsentry/sentry-go/slog"
)

func NewSentryLogger(ctx context.Context) *Logger {
	handler := sentryslog.Option{
		EventLevel: []slog.Level{slog.LevelError},
		LogLevel:   []slog.Level{slog.LevelWarn, slog.LevelInfo},
		AddSource:  true,
	}.NewSentryHandler(ctx)
    slogger := slog.New(handler)
    slogger.Info("I should attach to the trace in `ctx`")
```

### Expected result

I expected from the docs that the info log would be attached to the trace in the context I provided to `NewSentryHandler`.

### Actual result

slogger.Info creates a new `context.Background()`, so my log is not attached to the original trace. To do so, I would have to call `slogger.InfoContext(ctx, ...)` on every log. If this is intended behavior, I found the docs extremely confusing.

giortzisg:
Actually dove a bit deeper, and for multiple transactions there is an issue when getting the corrent span, so nice catch. Verified that the behavior works now (with my fix), but if you can also provide a full gin example to also test on that I would appreciate it!

Gitznik:
Here you go. In this example, when I post to the endpoint both logs end up in sentry.
- `This should be on the trace, but is not` ends up in an empty trace
- `This is actually on the trace` is attached to the trace of the http request

Thanks for digging into this 🙏 

```mod
go 1.25.2

require (
	github.com/getsentry/sentry-go v0.36.1
	github.com/getsentry/sentry-go/gin v0.36.1
	github.com/getsentry/sentry-go/slog v0.36.1
	github.com/gin-gonic/gin v1.11.0
)
```

```go
package main

import (
	"context"
	"log/slog"
	"net/http"
	"os"
	"time"

	"github.com/getsentry/sentry-go"
	sentrygin "github.com/getsentry/sentry-go/gin"
	sentryslog "github.com/getsentry/sentry-go/slog"
	"github.com/gin-gonic/gin"
)

func main() {
	if err := sentry.Init(sentry.ClientOptions{
		Dsn:              "YOUR_DSN",
		EnableTracing:    true,
[...]

PR #1118, opened 2025-10-24
fix context propagation for logs

giortzisg:
### Description

This PR fixes two issues with how the context was passed on logs:
- check for falling back on the initialization context, if slog is passing `context.Background()`
- use `SpanFromContext` first and then default back to `GetHubFromContext`.

#### Issues
* resolves: #1117
* resolves: GO-92

### DIFF
--- a/log_context_test.go
+++ b/log_context_test.go
@@ -1,0 +1,76 @@
+package sentry
+
+import (
+	"context"
+	"testing"
+
+	"github.com/getsentry/sentry-go/internal/testutils"
+)
+
+func TestSentryLogger_ShouldLinkToCorrectSpan(t *testing.T) {
+	transport := &MockTransport{}
+	err := Init(ClientOptions{
+		Dsn:              testDsn,
+		EnableTracing:    true,
+		TracesSampleRate: 1.0,
+		EnableLogs:       true,
+		Transport:        transport,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer Flush(testutils.FlushTimeout())
+
+	hub := CurrentHub().Clone()
+	ctx1 := SetHubOnContext(context.Background(), hub)
+	span1 := StartTransaction(ctx1, "request-1")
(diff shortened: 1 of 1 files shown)


slog/sentryslog.go
```diff
@@ -238,6 +238,10 @@ func (h *logHandler) Enabled(_ context.Context, level slog.Level) bool {
 }
 
 func (h *logHandler) Handle(ctx context.Context, record slog.Record) error {
+	// when logging without context, slog passes `context.Background`. Check for span existence to not overwrite the root context.
+	if sentry.SpanFromContext(ctx) == nil {
```
lcian: Hacky. And, are we guaranteed to never have a span on `context.Background`?

giortzisg: StartSpan adds the span as a context key, so it doesn't matter where it originates from, it only matters if the key exists.

log.go
```diff
@@ -90,20 +90,27 @@ func (l *sentryLogger) log(ctx context.Context, level LogLevel, severity int, me
...
 	var span *Span
 	var user User
 
+	span = SpanFromContext(ctx)
+	if span == nil {
+		span = SpanFromContext(l.ctx)
+	}
+	hub := GetHubFromContext(ctx)
+	if hub == nil {
+		hub = CurrentHub()
```
lcian, line 107: Should it be
```suggestion
		hub = GetHubFromContext(l.ctx)
```
and then if this is also nil, fallback to global hub?

slog/sentryslog.go
lcian: So your intention with this is to distinguish a legitimate context from `context.Background`, right?

Are we guaranteed to have no span on `context.Background`?
I suspect there might be cases where you don't have a span on the context, even if it's not `context.Background` necessarily.
For example, if there's simply no ongoing trace at that moment.

Would it be better to check `sentry.HubFromContext(ctx)` instead?

PR #1138, opened 2025-11-17
feat: drop rate-limited events

giortzisg:
### Description

To avoid filling up the buffer queues when rate-limited, we should be dropping events that got selected for processing. This is a behavioral change, since both the new and old transports/scheduler were just skipping the event without dropping.


cleptric:
Should note somewhere that we need client reports for this.

PR #1148, opened 2025-11-27
fix: deep-copy user mutated attributes

giortzisg:
### Description

Since the addition of Telemetry Buffers moves the serialization to a background worker, user provided attributes that can be mutated should be deep copied, to avoid panics during serialization.


scope.go
```diff
@@ -338,7 +340,7 @@ func (scope *Scope) Clone() *Scope {
 		clone.contexts[key] = cloneContext(value)
 	}
 	for key, value := range scope.extra {
-		clone.extra[key] = value
```
lcian, line 343: Correct me if I'm wrong, but:
this was here already and was not causing problems, and it's a separate code path from the buffer/transport, so I don't think this should be changed at all.
I assume this was deliberately done this way because scope forking happens frequently and so you wouldn't want to do an expensive deep copy.

scope.go
```diff
@@ -483,14 +489,6 @@ func (scope *Scope) ApplyToEvent(event *Event, hint *EventHint, client *Client)
 }
 
 // cloneContext returns a new context with keys and values copied from the passed one.
-//
-// Note: a new Context (map) is returned, but the function does NOT do
-// a proper deep copy: if some context values are pointer types (e.g. maps),
-// they won't be properly copied.
 func cloneContext(c Context) Context {
-	res := make(Context, len(c))
```
lcian, line 493: same as above

client.go
```diff
@@ -756,7 +756,32 @@ func (client *Client) processEvent(event *Event, hint *EventHint, scope EventMod
...
+			}
+			eventToBuffer.Breadcrumbs = bc
+		}
+		if event.Modules != nil {
+			eventToBuffer.Modules = deepCopyMapStringString(event.Modules)
+		}
+		if event.User.Data != nil {
+			eventToBuffer.User.Data = deepCopyMapStringString(event.User.Data)
+		}
+
```
lcian: Nested in `Event` there are other things we might care about, for example a Span can have Data (and apparently tags and extra), same about Log attributes, etc. Do we care about those?

giortzisg:
> ...or we can serialize it up front, and then send them to telemetry buffer after serialization?

I was weighting both approaches, but i think that pre-serialization has more drawbacks:
- the scheduler needs the whole event, to get the trace context to build the envelope (we would need to pass metadata around, along with the serialized event)
- batching logic for logs/spans would also require metadata and more complicated
- performance wise, it's better to serialize on the background for:
	- request latency for the user is low
	- when rate limited we drop immediately, skipping serialization completely
 
 > I would assume doing deep copy would inflict in more RAM

 That is a valid concern but we only keep the copy around till we flush the event (short lifespan) and other SDKs already copy some event data.

giortzisg:
> Okay so what's blocking this PR? Just the reviews?

Validity mostly. 
- Want to be sure that all values get deep copied correctly (maybe we introduce a library for that)
- that the issue isn't caused in another resource that we can't deep copy (eg. On SetRequest)

I am also not 100% sure that this is better than preserialization so weighting that as well.

PR #1151, opened 2025-11-30
feat: Add metrics support

### LINKED ISSUE #1136: Sentry Metrics for Go
[https://develop.sentry.dev/sdk/telemetry/metrics/](https://develop.sentry.dev/sdk/telemetry/metrics/)

Add support for [Sentry Metrics ](https://docs.sentry.io/product/explore/metrics/getting-started/)to the Go SDK!

This will enable users to record [counters, gauges, and distributions](https://develop.sentry.dev/sdk/telemetry/metrics/#metric-types).

1. Define the [Metrics Protocol](https://develop.sentry.dev/sdk/telemetry/metrics/#metrics-protocol) and `trace_metric` envelope type in the SDK.
2. Add the Public API (SDK options and methods) as per the [docs](https://develop.sentry.dev/sdk/telemetry/metrics/#public-api).
3. Make sure the the SDK follow the documented [behavior](https://develop.sentry.dev/sdk/telemetry/metrics/#sdk-behavior).
4. Attach [default attributes](https://develop.sentry.dev/sdk/telemetry/metrics/#default-attributes) and optional [user attributes](https://develop.sentry.dev/sdk/telemetry/metrics/#user-attributes) to the SDK as per docs.
5. Follow requirements for rate limiting, buffering, batching & flushing and build in measures to prevent excessive memory usage.

For general feedback & questions about Sentry Metrics use this [GH discussion](https://github.com/getsentry/sentry/discussions/102275); for Go SDK specific feedback & questions use the comment functionality on this GH issue.

aldy505:
### Description
Implement metrics based on https://develop.sentry.dev/sdk/telemetry/metrics/ and https://develop.sentry.dev/sdk/data-model/envelope-items/#trace-metric
#### Issues
- Closes #1136 
- Closes GO-99

### Usage example                                                                                                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                               
```go
ctx := context.Background()
meter := sentry.NewMeter(ctx)
// Attaching permanent attributes on the meter
meter.SetAttributes(
	attribute.String("version", "1.0.0"),
)

meter.Distribution("file_upload_size", 3.14,
	sentry.WithUnit(sentry.UnitMegabyte), // Using standard unit constants
	sentry.WithAttributes(
		attribute.String("file.type", "image/png"),
		attribute.String("bucket.region", "us-west-2"),
		attribute.String("bucket.name", "user-uploads"),
	),
)
```

### Changelog Entry

- Add metric support
	- support for three metric methods (counter, gauge, distribution)
	- custom metric units
	- unexport batchlogger

### DIFF
--- a/_examples/metrics/main.go
+++ b/_examples/metrics/main.go
@@ -24,7 +24,7 @@ func main() {
 			switch metric.Type {
 			case sentry.MetricTypeCounter:
 				if v, ok := metric.Value.Int64(); ok && v > 5 {
-					return nil // drop low-value counters
+					return nil // drop high-value counters
 				}
 			case sentry.MetricTypeGauge:
 				if v, ok := metric.Value.Float64(); ok && v < 10.0 {
@@ -54,7 +54,7 @@ func main() {
 	)
 
 	// Count metrics to measure occurrences of an event.
-	meter.Count("sent_emails", 1,
+	meter.Count("sent_emails", 6,
 		sentry.WithAttributes(
 			attribute.String("email.provider", "sendgrid"),
 			attribute.Int("email.number_of_recipients", 3),


interfaces.go
```diff
@@ -587,6 +632,33 @@ func (e *Event) defaultMarshalJSON() ([]byte, error) {
...
+			*event
+			Items           []Metric        `json:"items,omitempty"`
+			Type            json.RawMessage `json:"type,omitempty"`
+			Timestamp       json.RawMessage `json:"timestamp,omitempty"`
+			StartTime       json.RawMessage `json:"start_timestamp,omitempty"`
+			Spans           json.RawMessage `json:"spans,omitempty"`
+			TransactionInfo json.RawMessage `json:"transaction_info,omitempty"`
+		}
+		return json.Marshal(metricEvent{event: (*event)(e), Items: e.Metrics})
+	}
```
aldy505, line 685: @giortzisg Nice! Better than my old implementation lol https://github.com/getsentry/sentry-go/compare/master...aldy505:sentry-go:feat/tracemetrics#diff-b7b79f1ac09ee75185e8a967df69488f9fe9597d89aaa80d50296d9c028ee7d9R577-R593

batch_processor.go
```diff
@@ -0,0 +1,127 @@
...
+
+type BatchProcessor[T any] struct {
+	sendBatch    func([]T)
+	itemCh       chan T
+	flushCh      chan chan struct{}
+	cancel       context.CancelFunc
+	wg           sync.WaitGroup
+	startOnce    sync.Once
+	shutdownOnce sync.Once
+}
```
aldy505: @giortzisg I told you so :joy:

interfaces.go
```diff
@@ -141,6 +149,36 @@ type LogEntry interface {
...
+}
+
+// Meter provides an interface for recording metrics.
+type Meter interface {
+	// SetAttributes allows attaching parameters to the meter using the attribute API.
+	SetAttributes(...attribute.Builder)
+	// Count records a count metric.
+	Count(name string, count int64, options MeterOptions)
+	// Gauge records a gauge metric.
+	Gauge(name string, value float64, options MeterOptions)
```
aldy505: @giortzisg `Gauge` can accept normal integers (non-floats), that's why I created the `Numbers` interface above.

giortzisg: I know, but the generic solution didn't make much sense. It didn't apply to Count and Dist and in most cases you would need float64 anyways for most calculations, so it is extra verbose for no reason. Type casting in some cases is fine, but I might introduce also a GaugeInt and some other convenience methods.

aldy505: The convenience methods would be nice!

aldy505: On another note, I was thinking it's because of JSON serialization for float64 and int differences.

giortzisg: > JSON serialization for float64 and int differences.

That still is a problem, but doesn't get solved by generics. Count should be an int and Gauge/Dist floats, so the problem is storing and converting without precision loss between count/gauge/dist metric events.

internal/protocol/envelope.go
```diff
@@ -211,3 +212,17 @@ func NewLogItem(itemCount int, payload []byte) *EnvelopeItem {
...
 }
+
+// NewTraceMetricItem creates a new envelope item for trace metrics.
+func NewTraceMetricItem(itemCount int, payload []byte) *EnvelopeItem {
+	length := len(payload)
+	return &EnvelopeItem{
+		Header: &EnvelopeItemHeader{
+			Type:        EnvelopeItemTypeTraceMetric,
+			Length:      &length,
+			ItemCount:   &itemCount,
```
Litarnus, line 226: Do these have to be pointers? Since they are integer types I think they can be value types

internal/protocol/metric_batch.go
```diff
@@ -0,0 +1,41 @@
...
+	wrapper := struct {
+		Items []json.RawMessage `json:"items"`
+	}{Items: items}
+
+	payload, err := json.Marshal(wrapper)
+	if err != nil {
+		return nil, err
+	}
+
+	return NewTraceMetricItem(len(items), payload), nil
```
Litarnus, line 35: Since we are pre-allocating based on `len(ms)` and we `continue` on error, it could be that `len(items)` is greater than the actual number of items in the slice. Not sure if it would cause problems in Relay

batch_processor.go
```diff
@@ -0,0 +1,132 @@
...
+					<-timer.C
+				}
+				timer.Reset(batchTimeout)
+			}
+		case <-timer.C:
+			if len(items) > 0 {
+				p.sendBatch(items)
+				items = nil
+			}
+			timer.Reset(batchTimeout)
```
Litarnus: Probably nit: I'm not sure this follows specs. The doc says ` telemetry buffer MUST start a timeout of 5 seconds when the first item is added` but from what I understand here, it will immediately start the timeout and reset it every `batchTimeout` unless the buffer is filled. So if I'm not mistaken, the timer can run for 4 seconds, then an item is added and after 1 second, the batch is flushed.

Given that I don't see a `Start` method on the timer, this might be the best approach for Go though

interfaces.go
```diff
@@ -782,7 +881,7 @@ type Attribute struct {
 	Type  AttrType `json:"type"`
 }
 
-// MarshalJSON is a custom implementation that skips SpanID and timestamp when empty.
+// MarshalJSON converts a Log to JSON that skips SpanID and timestamp when empty.
```
Litarnus: nit: it might be worth noting that empty refers to the zero span id

metrics.go
```diff
@@ -0,0 +1,261 @@
...
+			"sentry.sdk.name":       client.sdkIdentifier,
+			"sentry.sdk.version":    client.sdkVersion,
+		}
+
+		defaultAttrs := make(map[string]Attribute)
+		for k, v := range defaults {
+			if v != "" {
+				defaultAttrs[k] = Attribute{Value: v, Type: AttributeString}
+			}
+		}
```
Litarnus, line 78: I think we can pull this out into a separate function since it's being used here too: https://github.com/getsentry/sentry-go/pull/1151/changes#diff-973afd7a3fa6644d1e62a8846c6ee7dde87077ff5c0ee193624f98154eb77687R72

metrics.go
```diff
@@ -0,0 +1,261 @@
...
+		if user.ID != "" {
+			attrs["user.id"] = Attribute{Value: user.ID, Type: AttributeString}
+		}
+		if user.Name != "" {
+			attrs["user.name"] = Attribute{Value: user.Name, Type: AttributeString}
+		}
+		if user.Email != "" {
+			attrs["user.email"] = Attribute{Value: user.Email, Type: AttributeString}
+		}
+	}
```
Litarnus: This also looks like a duplication of https://github.com/getsentry/sentry-go/pull/1151/changes#diff-973afd7a3fa6644d1e62a8846c6ee7dde87077ff5c0ee193624f98154eb77687R120

metrics.go
```diff
@@ -0,0 +1,261 @@
...
+				debuglog.Printf("Dropping event: metric buffer full or category missing")
+			}
+		} else if m.client.batchMeter != nil {
+			m.client.batchMeter.Send(metric)
+		}
+	}
+
+	if m.client.options.Debug {
+		debuglog.Printf("Metric %s [%s]: %v %s", metricType, name, value.AsInterface(), unit)
+	}
```
Litarnus: Should we move this into the `if metric != nil` block? If the callback returns nil this gets logged and it might be confusing to see this show up

scope.go
```diff
@@ -494,3 +495,74 @@ func cloneContext(c Context) Context {
...
+
+	scope := hub.Scope()
+	if scope == nil {
+		return User{}
+	}
+
+	scope.mu.RLock()
+	defer scope.mu.RUnlock()
+
+	return scope.user
```
Litarnus: Do we need this lock? The lock gets immediately released after the user is returned so if something is racing and waits on the lock somewhere else, it acquires it immediately afterwards allowing mutations

internal/protocol/envelope.go
giortzisg: they are pointers so that omitempty skips serializing them.

internal/protocol/metric_batch.go
giortzisg: we are preallocation, but the len of the slice actually tracks the amount of items inside. The `len(ms)` defines the capacity. So if any object fails to serialize we still keep track of the correct amount of items.

batch_processor.go
giortzisg: nice observation, we can refactor this so that we actually stick to the batchTimeout as expected.

scope.go
giortzisg: yeah, we need to lock to get the user, so that we don't get partially mutated user.

batch_processor.go
```diff
@@ -0,0 +1,127 @@
...
+		select {
+		case item := <-p.itemCh:
+			if len(items) == 0 {
+				timer.Reset(batchTimeout)
+			}
+			items = append(items, item)
+			if len(items) >= batchSize {
+				p.sendBatch(items)
+				items = nil
+			}
```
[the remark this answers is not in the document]
giortzisg: we don't support older go versions. the min is go1.23

metrics.go
giortzisg: for now I think it's fine to have it duplicated only on two spots. The helper would need the client, and at some point we might need to expand the default attributes for either metrics/logs and they might not match.

_examples/metrics/main.go
```diff
@@ -0,0 +1,141 @@
...
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/getsentry/sentry-go"
+	"github.com/getsentry/sentry-go/attribute"
+	sentryhttp "github.com/getsentry/sentry-go/http"
+)
+
+var meter sentry.Meter
```
aldy505, line 14: Nooo don't do this. It looks a lot like otel.

interfaces.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/getsentry/sentry-go/internal/ratelimit"
 )
 
+const errorType = ""
```
aldy505, line 17: What is this for? Providing comment would be great.

interfaces_test.go
```diff
@@ -785,6 +785,21 @@ func TestEvent_ToEnvelope(t *testing.T) {
...
+				EventID: "12345678901234567890123456789012",
+				Type:    "trace_metric",
+				Metrics: []Metric{
+					{
+						Name:      "test.metric",
+						Value:     Int64MetricValue(42),
+						Timestamp: time.Date(2023, 1, 1, 12, 0, 0, 0, time.UTC),
+						Type:      MetricTypeCounter,
+					},
+				},
```
aldy505, line 800: We should also test for invalid and nil MetricValue, as well as invalid metric type (other than count, dist, gauge)

internal/protocol/envelope.go
aldy505: Alternatively we can use github.com/guregu/null

internal/testutils/maps.go
```diff
@@ -0,0 +1,15 @@
+package testutils
+
+// MergeMaps merges multiple maps into a single map.
+// If there are duplicate keys, the value from the last map takes precedence.
+//
+// CC BY-SA 4.0 Oliver (https://stackoverflow.com/a/74750675/3153224)
+func MergeMaps[M ~map[K]V, K comparable, V any](src ...M) M {
```
aldy505, line 7: We can safely remove this if this is not used.

metrics.go
```diff
@@ -0,0 +1,261 @@
...
+const (
+	UnitNanosecond  = "nanosecond"
+	UnitMicrosecond = "microsecond"
+	UnitMillisecond = "millisecond"
+	UnitSecond      = "second"
+	UnitMinute      = "minute"
+	UnitHour        = "hour"
+	UnitDay         = "day"
+	UnitWeek        = "week"
+)
```
aldy505, line 24: I would prefer to have this as a type of it's own, better for static typing.

interfaces.go
giortzisg: the errorType if an empty string if you have a look at how the sdk currently works. Just added the type so that it makes more sense.

metrics.go
giortzisg: We want to receive custom units as string, but also allow the users to use the supported ones like that, so can't really have a custom type for it.

aldy505: So the freedom of type/unit is intentional?

PR #1167, opened 2026-01-09
ci(release): Switch from action-prepare-release to Craft

BYK:
## Summary

This PR migrates from the deprecated `action-prepare-release` to the new Craft GitHub Actions.

## Changes

- Migrated `.github/workflows/release.yml` to Craft reusable workflow

## Documentation

See https://getsentry.github.io/craft/github-actions/ for more information.

### DIFF
--- a/.github/workflows/changelog-preview.yml
+++ b/.github/workflows/changelog-preview.yml
@@ -7,6 +7,7 @@ on:
     - reopened
     - edited
     - labeled
+    - unlabeled
 permissions:
   contents: write
   pull-requests: write


giortzisg (approved):
Hey @BYK thanks for the contribution, that's a nice change. While skimming through the craft docs, the comment from the bug bot seems to be accurate, so let's change that and then we can merge this.

BYK:
@giortzisg nope as we now support `auto` versioning and that kicks in when the version is empty. We just may need to set

```yaml
versioning:
  policy: auto
```

in `.craft.yml` file

PR #1183, opened 2026-01-24
feat!: add support for Echo v5

Scorfly:
**BREAKING CHANGE:** echo integration now requires github.com/labstack/echo/v5 and Go 1.25. Echo v4 is no longer supported.

## Dependency and module

- Bump github.com/labstack/echo from v4.10.1 to v5.0.0.
- Set go directive to 1.25.0 in echo/go.mod.
- Drop Echo v4-only indirects (gommon, go-colorable, go-isatty, bytebufferpool, fasttemplate, x/crypto, x/net) and update remaining (x/sys, x/text). go.sum updated accordingly.

## Echo v5 API adjustments

- Use `*echo.Context` instead of `echo.Context` in handler and public API (GetHubFromContext, SetHubOnContext, GetSpanFromContext) to match v5’s pointer-based Context.
- Update all handler signatures in code and docs from `func(c echo.Context) error` to `func(c *echo.Context) error`.
- In v5, `Context.Response()` returns `http.ResponseWrite`r, not `*Response`, so `ctx.Response().Status` is no longer available.

## HTTP status and Response handling

- Replace direct `ctx.Response().Statu`s with `echo.UnwrapResponse(ctx.Response())` to obtain `*echo.Response` and use `resp.Status` when `UnwrapResponse` succeeds and `resp.Status != 0`.
- When `UnwrapResponse` fails (e.g. middleware replaces the response with a writer that does not unwrap to `*echo.Response`), leave status at its zero value (0) instead of defaulting to 200.
- For handler-returned errors, use `echo.HTTPStatusCoder` instead of `*echo.HTTPError` so that both `*HTTPError` and unexported `*httpError` (ErrNotFound, ErrMethodNotAllowed, etc.) are handled and the correct status is used for the transaction.

## Tests

- In TestIntegration, skip route registration when Handler is nil so the “404 / no route” case does not call `router.GET("", nil)`. Echo v5’s router rejects `Handler == nil` and panics with “adding route without handler function”.
- Add `TestUnwrapResponseError`: when the response is wrapped by a writer that does not implement `Unwrap()` (e.g. `&struct{ http.ResponseWriter }{}`), UnwrapResponse returns an error; the middleware must not panic and must record `http.response.status_code` as 0 in the transaction.

## Documentation and examples

- `README.md` and `example_test.go`: switch imports and examples from echo/v4 to echo/v5 and from `echo.Context` to `*echo.Context`.


giortzisg:
We already had some discussions internally about upgrading go versions for the integrations separately to the main sdk, so we will support v5.

echo/sentryecho.go
```diff
@@ -93,10 +93,13 @@ func (h *handler) handle(next echo.HandlerFunc) echo.HandlerFunc {
 		transaction.SetData("http.request.method", r.Method)
 
 		defer func() {
-			status := ctx.Response().Status
+			var status int
+			if resp, err := echo.UnwrapResponse(ctx.Response()); err == nil && resp.Status != 0 {
```
giortzisg, line 98: We should skip setting `http.response.status_code` if Status = 0. Also let's add a debug statement that we couldn't get the response status code.

MartinodF:
What are the plans for cutting a release with this? We're eagerly awaiting a new version to be published and I wanted to avoid pinning a specific commit on the master branch :) Thanks!

giortzisg:
I want to also include one more thing for the release, but will happen this or the next week.

giortzisg:
@MartinodF just released https://github.com/getsentry/sentry-go/releases/tag/v0.45.0

Issue #1127, opened 2025-11-03
Add client reports


giortzisg:
Implement client reports based on [https://develop.sentry.dev/sdk/telemetry/client-reports/](https://develop.sentry.dev/sdk/telemetry/client-reports/)

PR #1192, opened 2026-02-10
feat: add support for client reports

giortzisg:
### Description
add support for [client reports](https://develop.sentry.dev/sdk/telemetry/client-reports/). 

### Issues
* resolves: #1127 
* resolves: GO-94

<details>

</details>

<details>
<summary>Reminders</summary>

</details>

### DIFF
--- a/client.go
+++ b/client.go
@@ -851,7 +851,7 @@ func (client *Client) processEvent(event *Event, hint *EventHint, scope EventMod
 	// (errors, messages) are sampled here. Does not apply to check-ins.
 	if event.Type != transactionType && event.Type != checkInType && !sample(client.options.SampleRate) {
 		debuglog.Println("Event dropped due to SampleRate hit.")
-		client.reportRecorder.RecordOne(report.ReasonSampleRate, ratelimit.CategoryError)
+		client.reportRecorder.RecordOne(report.ReasonSampleRate, event.toCategory())
 		return nil
 	}
 


report/aggregator.go
```diff
@@ -0,0 +1,160 @@
...
+
+	a.mu.Lock()
+	counter, exists := a.outcomes[key]
+	if !exists {
+		counter = &atomic.Int64{}
+		a.outcomes[key] = counter
+	}
+	a.mu.Unlock()
+
+	counter.Add(quantity)
```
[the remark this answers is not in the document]
giortzisg: The counter is atomic

transport.go
```diff
@@ -758,6 +916,8 @@ func (a *internalAsyncTransportAdapter) Configure(options ClientOptions) {
 		HTTPProxy:     options.HTTPProxy,
 		HTTPSProxy:    options.HTTPSProxy,
 		CaCerts:       options.CaCerts,
+		Recorder:      a.recorder,
+		Provider:      a.provider,
 	}
 
 	a.transport = httpinternal.NewAsyncTransport(transportOptions)
```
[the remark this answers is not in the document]
giortzisg: the internal/http/transport does the check.

tracing.go
```diff
@@ -429,6 +431,17 @@ func (s *Span) doFinish() {
...
+		c := hub.Client()
+		if c != nil {
+			if !s.IsTransaction() {
+				// we count the sampled spans from the transaction root. it is guaranteed that the whole transaction
+				// would be sampled
+				return
+			}
+			children := s.recorder.children()
+			c.reportRecorder.RecordOne(report.ReasonSampleRate, ratelimit.CategoryTransaction)
+			c.reportRecorder.Record(report.ReasonSampleRate, ratelimit.CategorySpan, int64(len(children)+1))
```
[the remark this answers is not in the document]
giortzisg: Sampling a transaction would drop all spans. The number should be correct.

internal/util/util.go
```diff
@@ -81,3 +82,38 @@ func EnvelopeIdentifier(envelope *protocol.Envelope) string {
...
+	response, err := client.Do(request)
+	if err != nil {
+		return nil, err
+	}
+
+	success := HandleHTTPResponse(response, identifier)
+	limits := ratelimit.FromResponse(response)
+
+	// Drain body up to a limit and close it, allowing the transport to reuse TCP connections.
+	_, _ = io.CopyN(io.Discard, response.Body, MaxDrainResponseBytes)
```
sl0thentr0py, line 111: what is this lol

```
 is the drain in @internal/util/util.go shady
  ⎿  Read internal/util/util.go (120 lines)

⏺ Not shady — it's a standard Go pattern. Here's why it exists:

  The problem: Go's http.Client reuses TCP connections (called "keep-alive") for performance. But it can only reuse a connection if the previous response body was fully read and closed. If you just close without reading, Go
   throws away the connection and opens a new one for the next request — slower.

  What the code does:

  _, _ = io.CopyN(io.Discard, response.Body, MaxDrainResponseBytes)  // read up to 16KB, throw it away
  response.Body.Close()                                                // then close

  It reads up to 16KB of leftover response data into the void, then closes the body. This lets Go recycle the TCP connection.

  The 16KB cap (16 << 10) is a safety net — if the server sends back a huge response (e.g. an HTML error page), it won't waste time/memory reading all of it. It just gives up connection reuse for that one response.

  The _, _ = part intentionally discards both the byte count and any error. That's fine here because:
  - The data itself is unwanted (already logged what was needed in HandleHTTPResponse)
  - If the drain fails, the worst outcome is a non-reused connection — not a correctness issue

  This is a well-known Go idiom. You'll see it in the Go standard library docs and most HTTP client code.
```

client.go
```diff
@@ -383,10 +388,18 @@ func NewClient(options ClientOptions) (*Client, error) {
...
+		options:        options,
+		dsn:            dsn,
+		sdkIdentifier:  sdkIdentifier,
+		sdkVersion:     SDKVersion,
+		reportRecorder: report.NoopRecorder(),
+		reportProvider: report.NoopProvider(),
+	}
+
+	if !options.DisableClientReports {
+		a := report.NewAggregator()
```
sl0thentr0py, line 400: i understand that aggregator has the same functions as both recorder and provider, but then why make those interfaces in the first place? this is very confusing for a non-go person

internal/util/util.go
giortzisg: The first part is that for go to reuse the tcp connection we need to empty and close the io connection, so we just throw all remaining bytes.
The second part with MaxDrainResponseBytes already existed in the SDK as a pattern and I am not really convinced about it's use(since we don't really expect relay to return a malicious response), but it's what the llm suggests, just a safety limit for draining that many response bytes.

client.go
giortzisg: it's mostly separation of concerns and not exposing the take reports functionality to a layer that doesn't need it.

Issue #1189, opened 2026-02-09
Extend the attribute API to support slices


giortzisg:
Add support for adding homogenous arrays/slices to the `attribute` package according to [https://develop.sentry.dev/sdk/telemetry/attributes/#complex-types](<https://develop.sentry.dev/sdk/telemetry/attributes/#complex-types>)

PR #1203, opened 2026-02-20
feat: add support for homogenous arrays

giortzisg:
### Description
Add support for scalar slices on the attribute API. We also need to modify the log_entry to extend it to support the new methods as well.

### Issues
* resolves: #1189
* resolves: GO-118

<details>

</details>

<details>
<summary>Reminders</summary>

</details>

### DIFF
--- a/attribute/value.go
+++ b/attribute/value.go
@@ -53,6 +53,10 @@ const (
 	FLOAT64SLICE
 	// STRINGSLICE is a slice of strings Type Value.
 	STRINGSLICE
+	// UINT64 is a 64-bit unsigned integral Type Value.
+	//
+	// This type is intentionally not exposed through the Builder API.
+	UINT64
 )
 
 // BoolValue creates a BOOL Value.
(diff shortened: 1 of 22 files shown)


attribute/value.go
```diff
@@ -107,24 +157,60 @@ func (v Value) AsBool() bool {
 	return rawToBool(v.numeric)
 }
 
+// AsBoolSlice returns the []bool value. Make sure that the Value's type is
+// BOOLSLICE.
+func (v Value) AsBoolSlice() []bool {
```
Litarnus, line 162: nit: I was thinking if it makes sense to return an error instead of `nil` for those methods instead. But it might be too noisy for the benefit it would bring

giortzisg: We don't return an error on the other `AsX` methods, so I don't think it makes sense to add it only here. Also use case wise:
- if we want just the value `AsInterface` is the one to use.
- and if we want the type for modifying the value then, we have to use `v.Type()` anyways.

PR #1252, opened 2026-04-08
fix(fiber): Use UserContext for transaction to enable OTel trace linking

giortzisg:
## Summary

- The fiber middleware was using `ctx.Context()` (raw fasthttp context) instead of `ctx.UserContext()` when starting the transaction. The raw fasthttp context does not propagate `context.WithValue` wrappers, which broke OTel trace linking since the OTel span context is stored via `context.WithValue`.
- The transaction context was not being set back on `UserContext` before calling `ctx.Next()`, so downstream handlers and panic recovery did not have access to the active span.
- The converted `*http.Request` now carries the transaction context via `r.WithContext(transaction.Context())` so that `scope.request.Context()` has trace info available for event processing (e.g., trace resolution in the OTel integration).

## Motivation

This fix is needed for the OTel integration to properly link errors, logs, and metrics to the active trace in fiber handlers. Without it, the OTel span context is lost at the point where the Sentry transaction is created, breaking the trace linkage.

## Changes

- `fiber/sentryfiber.go`: Use `ctx.UserContext()` instead of `ctx.Context()` when passing context to `sentry.StartTransaction`
- `fiber/sentryfiber.go`: Call `ctx.SetUserContext(transaction.Context())` before `ctx.Next()` so downstream handlers see the transaction span
- `fiber/sentryfiber.go`: Set `r = r.WithContext(transaction.Context())` before `scope.SetRequest(r)` so the scope request carries trace context

## Test plan

- [x] `go build ./...` passes in the fiber module
- [x] `go test -race ./...` passes in the fiber module
- [x] Verify OTel trace linking works end-to-end with a fiber app instrumented with both Sentry and OTel

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/fiber/sentryfiber.go
+++ b/fiber/sentryfiber.go
@@ -80,11 +80,17 @@ func (h *handler) handle(ctx *fiber.Ctx) error {
 		sentry.WithSpanOrigin(sentry.SpanOriginFiber),
 	}
 
+	savedCtx := ctx.UserContext()
+	requestCtx, cancel := context.WithCancel(savedCtx)
+	defer cancel()
+	defer func() { ctx.SetUserContext(savedCtx) }()
+
 	transaction := sentry.StartTransaction(
-		sentry.SetHubOnContext(ctx.UserContext(), hub),
+		sentry.SetHubOnContext(requestCtx, hub),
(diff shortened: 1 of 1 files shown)


Litarnus (approved):
LGTM

PR #1276, opened 2026-04-22
fix(lint): Resolve workspace submodule lint issues

giortzisg:
Update the workspace lint target to run across go.work modules and fix the issues that surfaced once submodules were included.

Clean up tests and integration helpers so the expanded golangci-lint run passes across the root module and submodules.

#skip-changelog

Co-Authored-By: GPT-5.4 <noreply@example.com>

### DIFF
--- a/zap/converter.go
+++ b/zap/converter.go
@@ -97,6 +97,16 @@ type uint64LogEntry interface {
 	Uint64(key string, value uint64) sentry.LogEntry
 }
 
+func int64BitsToUint64(value int64) uint64 {
+	//nolint:gosec // zap stores unsigned integers and float bit patterns in Field.Integer.
+	return uint64(value)
+}
+
+func int64BitsToUint32(value int64) uint32 {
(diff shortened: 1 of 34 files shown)


zap/converter.go
```diff
@@ -97,6 +97,16 @@ type uint64LogEntry interface {
...
 
+func int64BitsToUint64(value int64) uint64 {
+	//nolint:gosec // zap stores unsigned integers and float bit patterns in Field.Integer.
+	return uint64(value)
+}
+
+func int64BitsToUint32(value int64) uint32 {
+	//nolint:gosec // zap stores float32 bit patterns in the lower 32 bits of Field.Integer.
+	return uint32(value)
+}
```
Litarnus: 🚲 : Maybe it would be better to just call to `toUint64`/`toUint32` to immediately see what the result is. The first half of the function name is in the expected type anyway

Litarnus (approved):
LGTM

PR #1281, opened 2026-04-24
feat(sql): add base sql/driver wrappers

### LINKED ISSUE #1277: feat(sql): implement sentrysql sub-module with passthrough driver wrappers
Adds the [`github.com/getsentry/sentry-go/sql`](<http://github.com/getsentry/sentry-go/sql>) sub-module with the full database/sql/driver wrapper tree

giortzisg:
### Description
This PR adds the base passthrough wrappers for `database/sql/driver` and the public API . Instrumentation will be added in a follow-up PR.

The core design is for `sentryConn` to always expose context-aware interfaces. Then each method delegates to the underlying driver implementation. For legacy drivers the wrapper falls back to the non-context methods. This way we can correctly instrument the legacy drivers as well with the correct context.

#skip-changelog 

### Issues
* resolves: #1277 
* resolves: GO-137

<details>

</details>

<details>
<summary>Reminders</summary>

</details>


sql/conn.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	values, err := namedValuesToValues(args)
+	if err != nil {
+		return nil, err
+	}
+	select {
+	default:
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	}
```
aldy505, line 67: Honestly this is useless, it just guards whenever the `cancel` is called before we call the `ex.Exec()`. If it's already passed, then it won't be canceled at all.

sql/stmt.go
```diff
@@ -0,0 +1,85 @@
...
+
+func newStmt(s driver.Stmt, cfg *config, query string) driver.Stmt {
+	return &sentryStmt{stmt: s, cfg: cfg, query: query}
+}
+
+// Close implements driver.Stmt.
+func (s *sentryStmt) Close() error { return s.stmt.Close() }
+
+// NumInput implements driver.Stmt.
+func (s *sentryStmt) NumInput() int { return s.stmt.NumInput() }
```
sl0thentr0py, line 24: doesn't wrapping the struct do all this automatically? why do we need all these explicit forwards?

sql/sentrysql.go
```diff
@@ -0,0 +1,105 @@
...
+		return nil
+	}
+	sql.Register(registeredName, wrapped)
+	registered[registeredName] = struct{}{}
+	return nil
+}
+
+// getDriver opens and closes a driver to retrieve the driver implementation we need to wrap.
+// We won't use the connection opened here, so we close it to avoid leaks.
+func getDriver(name, dataSourceName string) (driver.Driver, error) {
```
sl0thentr0py, line 95: technically we aren't allowed to do this because it's a network side effect. Can you explain why we need to open a connection and why there's no other way to achieve what we want here?

sql/stmt.go
giortzisg: Yes embedding does this automatically. However with embedding you get all methods exposed by the underlying `driver.Stmt` or `driver.Driver` and we can't really have that here, because we don't want to expose the old Deprecated `Query` and `Exec` methods.

Instrumentation flows on QueryContext and calls `Query` only if the driver is a legacy one, so we shouldn't implement `Query` on our wrapped stmt or driver.

sql/sentrysql.go
giortzisg: The driver we want to get is registered with `sql.Register` and there's no other way to get it back (to my knowledge). Also from the `database/sql` docs: [Open may just validate its arguments without creating a connection to the database. ](https://pkg.go.dev/database/sql#Open).

Also for more reference, here's how [otelsql](https://github.com/XSAM/otelsql/blob/9af8b52bf5f2ccb79af8bc1557530b57a8a705d6/sql.go#L87) Opens the connection as well.

sql/conn.go
giortzisg: Not completely useless, but I get your point

sl0thentr0py (approved):

PR #1282, opened 2026-04-25
feat: implement valkey and rueidis sentry instrumentation

### LINKED ISSUE #1268: Valkey SDK cache/db support
### Problem Statement

Like in case of [GO-95](https://linear.app/getsentry/issue/GO-95/databasesql-integration) sentry should have a sub package implementation for valkey/redis kind of databases, which should support cache and db behavior for insights modules. (Basically on the Backend tab we should be able to see and use these metrics not just in Traces)

### Solution Brainstorm

Honestly I have a working POC for valkey already.<br>Luckily valkey exposes a hook: [https://github.com/valkey-io/valkey-go/tree/main/valkeyhook](<https://github.com/valkey-io/valkey-go/tree/main/valkeyhook>) for instrumentation purposes.

Also since valkey-go is a fork of [rueidis](<https://github.com/redis/rueidis>) we can use the same hook implementation with that client too.

Next we have to make the traces compliant with [https://develop.sentry.dev/sdk/telemetry/traces/modules/caches/](<https://develop.sentry.dev/sdk/telemetry/traces/modules/caches/>) and [https://develop.sentry.dev/sdk/telemetry/traces/modules/queries/](<https://develop.sentry.dev/sdk/telemetry/traces/modules/queries/>) and [https://develop.sentry.dev/sdk/telemetry/traces/sql-transactions/](<https://develop.sentry.dev/sdk/telemetry/traces/sql-transactions/>) (for the multi command)

Following a config pattern like this: [https://github.com/getsentry/sentry-go/blob/master/http/sentryhttp.go](<https://github.com/getsentry/sentry-go/blob/master/http/sentryhttp.go>)

I've ended up in my POC with the following public interface:

```go
	Dsn:              os.Getenv("SENTRY_DSN"),
	TracesSampleRate: 1.0,
	EnableTracing:    true,
	AttachStacktrace: true,
})
if err != nil {
	log.Fatalf("sentry.Init: %s", err)
}

// Create a valkey client
dbClient, err := valkey.NewClient(valkey.ClientOption{
	InitAddress: []string{"localhost:6379"},
})

if err != nil {
	log.Fatalf("valkey.NewClient (db): %s", err)
}

// The sentryvalkey.Type can be either DB or Cache, so you can switch it up per use case.
// Here we wrap valkey client with a custom hook that implements the `valkeyhook.Hook` interface
dbClient = valkeyhook.WithHook(dbClient, sentryvalkey.New(sentryvalkey.Options{
	Type: sentryvalkey.TypeDB,
}))
defer dbClient.Close()
```

Also I'm more than happy to finalize and make this implementation production ready and contribute/maintain this, since we are already using this.

It expose traces like this, which are perfectly capable to be parsed by the caching layer:

Image

Image

(Currently the DB layer is kinda buggy, it cannot show query descriptions, but it can parse QPS and every other variable.)

Shikachuu:
Implemented Redis and Valkey (collectively redis like) database and cache layer instrumentation.

Redis implementation is using the upstream https://github.com/redis/rueidis library, it's a faster alternative to redis-go and almost 1/1 compatible with valkey-go.

Valkey implementation using the upstream https://github.com/valkey-io/valkey-go library.

All implementations are built around the upstream library's hook system to avoid collision with otel, while still providing the leanest possible downstream interface.

All common logic has been extracted to an `internal/redis` folder to help us maintain feature parity between implementation.

PS: This ended up being a much larger PR that I've initially anticipated, lmk if we should split this into 2 or 3 PRs.

### Issues

* resolves #1268
* resolves GO-136

<details>

</details>

<details>
<summary>Reminders</summary>

</details>


Shikachuu:
Also here are some sentry screenshots how the traces looks like in practice, I've managed to fix database query recognition for the db insight module.

As well as followed the practices demonstrated in other upstream Sentry SDK implementations such as laravel and nextjs, so we have the cache insight module function correctly as well, you can try them by running the examples.

image

image

We the default metric naming and labels uses the cache one, however we can change it via changing the `Type` field in `valkey.Options` eg: `sentryvalkey.Options{Type: sentryvalkey.TypeDB}`

valkey/sentryvalkey.go
```diff
@@ -0,0 +1,241 @@
...
+	stream := client.DoStream(span.Context(), cmd)
+
+	if err := stream.Error(); err != nil {
+		span.Status = sentry.SpanStatusInternalError
+	} else {
+		span.Status = sentry.SpanStatusOK
+	}
+	span.Finish()
+
+	return stream
```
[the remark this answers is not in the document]
Shikachuu: `valkeyhook.Hook` interface explicitly requires these interfaces (hence the interface implementation guard `var _` hack), the hook must return the stream object. Once returned, the caller consumes it - we have no callback or wrapper point to detect when streaming completes or errors.

Edit: this is true btw for both implementations!

valkey/sentryvalkey.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	return redis.Address{}
+}
+
+func respSize(result valkey.ValkeyResult) int {
+	data, err := result.AsBytes()
+	if err != nil {
+		return 0
+	}
+	return len(data)
```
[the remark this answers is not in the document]
Shikachuu: For commands like `MGET` or `INCR` there is no meaningful size to report imho this isn't a bug, yet a known limitation, absent data is better than wrong data, since this ain't going to fail, but return 0.

I guess this is why this was categorized as `Severity: LOW`

internal/redis/scrub.go
```diff
@@ -0,0 +1,244 @@
...
+	"SREM":        {fixed: []argRole{roleKey}, repeating: []argRole{roleValue}},
+	"SISMEMBER":   {fixed: []argRole{roleKey, roleValue}},
+	"SMISMEMBER":  {fixed: []argRole{roleKey}, repeating: []argRole{roleValue}},
+	"SMEMBERS":    {fixed: []argRole{roleKey}},
+	"SCARD":       {fixed: []argRole{roleKey}},
+	"SRANDMEMBER": {fixed: []argRole{roleKey}, tailRole: roleField},
+	"SPOP":        {fixed: []argRole{roleKey}, tailRole: roleField},
+
+	// Sorted set commands
+	"ZADD":          {fixed: []argRole{roleKey}, tailRole: roleField},
```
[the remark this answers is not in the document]
Shikachuu: `ZADD` flags best to handle as regular params and scrub them, instead of handling them by `ZADD myset NX ? ?` we would have `ZADD myset ? ? ?` since it would be too hard to distinguish form regular args since it's variable sized (`NX`, `INCR` etc.) and the args are also variable sized. (see the new test cases)

Otherwise fixed with 3 new test cases created to verify this behavior!

rueidis/sentryrueidis.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	result := client.DoCache(span.Context(), cmd, ttl)
+
+	redis.FinishSpan(span, h.typ, isReadOnly, result.Error(), rueidis.IsRedisNil(result.Error()), respSize(result))
+	return result
+}
+
+//nolint:revive // ctx position is dictated by rueidishook.Hook interface.
+func (h *sentryHook) DoMulti(client rueidis.Client, ctx context.Context, multi ...rueidis.Completed) []rueidis.RedisResult {
```
giortzisg, line 76: is there a way to wrap what redis actually executes here? Creating a single span here seems misleading and duration would probably be inaccurate.

Also we don't really have a pipeline span convention, so I feel like this is not exactly what we want.

internal/redis/scrub.go
```diff
@@ -0,0 +1,244 @@
...
+
+	for i, arg := range args {
+		if roleForPosition(i, fixedLen, pattern) == roleKey {
+			keys = append(keys, arg)
+		}
+	}
+
+	if len(keys) == 0 && len(cmds) > 1 {
+		return []string{cmds[1]}
+	}
```
giortzisg: the bugbot comments looks correct here. the behavior should be to skip instrumenting instead of falling back to the first argument since we might leak PII here. In general we should completely scrub AUTH like [python SDK](https://github.com/getsentry/sentry-python/blob/a4c16d7bd2f6206ee2b6881fd39e4571135b6421/sentry_sdk/integrations/redis/consts.py#L14).

We should add a testcase for that and verify that we don't leak it.

internal/redis/span.go
```diff
@@ -0,0 +1,201 @@
...
+		if isNilErr {
+			span.SetData("cache.hit", false)
+		} else if err == nil {
+			span.SetData("cache.hit", true)
+			if itemSize > 0 {
+				span.SetData("cache.item_size", itemSize)
+			}
+		}
+	}
+}
```
giortzisg: we should set `cache.write` here also.

internal/redis/scrub.go
```diff
@@ -0,0 +1,244 @@
...
+	args := cmds[1:]
+	fixedLen := len(pattern.fixed)
+
+	for i, arg := range args {
+		b.WriteByte(' ')
+
+		role := roleForPosition(i, fixedLen, pattern)
+		switch role {
+		case roleKey, roleField:
+			b.WriteString(arg)
```
giortzisg: There's a case that fields can actually contain PII. I went ahead to compare with python and they preserve only the key and filter later args unless send_default_pii=true.

rueidis/sentryrueidis.go
```diff
@@ -0,0 +1,232 @@
...
+func (h *sentryHook) Do(client rueidis.Client, ctx context.Context, cmd rueidis.Completed) rueidis.RedisResult {
+	cmds := cmd.Commands()
+	addr := extractAddr(client)
+
+	span := redis.StartSpan(ctx, h.typ, redis.DBSystemRedis, cmds, cmd.IsReadOnly(), addr)
+	defer span.Finish()
+
+	result := client.Do(span.Context(), cmd)
+
+	redis.FinishSpan(span, h.typ, cmd.IsReadOnly(), result.Error(), rueidis.IsRedisNil(result.Error()), respSize(result))
```
giortzisg, line 51: Does `rueidis.IsRedisNil` cover every case for a cache miss?

rueidis/sentryrueidis.go
```diff
@@ -0,0 +1,232 @@
+package rueidis
```
giortzisg, line 1: we should add module level tests for both valkey and rueidis.

go.work
```diff
@@ -13,7 +13,9 @@ use (
 	./negroni
 	./otel
 	./otel/otlp
+	./rueidis
```
giortzisg, line 16: Not related to this file, but we should add two entries to `.craft.yml` for release targets.

internal/redis/span.go
```diff
@@ -0,0 +1,201 @@
...
+		if isPipeline {
+			return "cache.pipeline"
+		}
+		if IsDeleteCommand(cmds) {
+			return "cache.remove"
+		}
+		if isReadOnly {
+			return "cache.get"
+		}
+		return "cache.put"
```
giortzisg: it seems too broad to default to only those two here.

rueidis/sentryrueidis.go
```diff
@@ -0,0 +1,232 @@
...
+	span := redis.StartSpan(ctx, h.typ, redis.DBSystemRedis, cmds, cmd.IsReadOnly(), addr)
+
+	stream := client.DoStream(span.Context(), cmd)
+
+	if err := stream.Error(); err != nil {
+		span.Status = sentry.SpanStatusInternalError
+	} else {
+		span.Status = sentry.SpanStatusOK
+	}
+	span.Finish()
```
giortzisg: Saw the reply to bugbot for not being able to properly determine the actual finish time for streams. IMO it's misleading to instrument stream spans with an incorrect duration, so it's better to skip them for now. We can leave this as a noop with a comment.

giortzisg (commented):
Hey @Shikachuu,

Thanks for taking the time. Had an initial first pass through the code and raised a few things. I want to go a bit deeper on commandPatterns on the next pass to properly validate each command, but leaving an initial review to keep things moving.

rueidis/sentryrueidis.go
Shikachuu: ahh so you mean every single call should be it's fully own span in a sense of there should be 0 difference between a `DoMulti` w 5 commands and calling `Do` 5 times?

internal/redis/scrub.go
Shikachuu: Done ✅

internal/redis/scrub.go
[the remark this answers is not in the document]
Shikachuu: Done ✅

Shikachuu: Done ✅

internal/redis/span.go
Shikachuu: Done ✅

internal/redis/span.go
Shikachuu: fair point! there were multiple logical issues w my impl. if we compare it to python (it was way less strict), let me summarize, what has been done in: https://github.com/getsentry/sentry-go/pull/1282/changes/64d532f9275956fc1768c8d0b7f3a249ef64b3c8

1. i've refactored the whole instrumentation and operation logic to match the python implementation! first of all `spanOp` is now a fully exhaustive switch
2. besides that commands that has nothing to do w caching like `ECHO` or `PING` etc. aren't getting any spans in `TypeCache` mode to 1:1 match the Python implementation
3. tests were also refactored to match this new mentality
4. `cache.flush` was completely missing, so it has been implemented too

rueidis/sentryrueidis.go
Shikachuu: `RESP3` protocol treats all classic cache miss scenarios like missing key etc. `nil` responses, every other error would be either be a non-cache related command's logical error or network error afaik.

i might be wrong here 🤔

rueidis/sentryrueidis.go
Shikachuu: Done ✅

rueidis/sentryrueidis.go
Shikachuu: yeah right, was thinking the same but wasn't really sure, removed!

rueidis/sentryrueidis.go
giortzisg: Yes that was the idea, I feel it's easier to understand and trace through the logic, because each `Do` is a separate command. Conceptually it makes sense to me to have a multi span that spawns 5 more spans for each `Do`. Might be wrong though so feel free to correct me here.

giortzisg: Actually, since this is a single network call, so a single span is the correct approach. So we can keep it as is.

internal/redis/scrub.go
```diff
@@ -0,0 +1,259 @@
...
+func ExtractKeys(cmds []string) []string {
+	if len(cmds) < 2 {
+		return nil
+	}
+
+	cmdName := strings.ToUpper(cmds[0])
+	pattern, known := commandPatterns[cmdName]
+	if !known {
+		// Unknown commands: best-effort, return first arg as key.
+		return []string{cmds[1]}
```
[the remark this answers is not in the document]
giortzisg: This sounds legit and I agree that for unknown spans it's better to skip instrumenting @Shikachuu

internal/redis/scrub.go
```diff
@@ -0,0 +1,259 @@
...
+	"SCAN":      {fixed: []argRole{roleField}, tailRole: roleField},
+	"DUMP":      {fixed: []argRole{roleKey}},
+	"RESTORE":   {fixed: []argRole{roleKey, roleField, roleValue}, tailRole: roleField},
+
+	// Pub/Sub
+	"SUBSCRIBE":    {repeating: []argRole{roleKey}},
+	"UNSUBSCRIBE":  {repeating: []argRole{roleKey}},
+	"PSUBSCRIBE":   {repeating: []argRole{roleKey}},
+	"PUNSUBSCRIBE": {repeating: []argRole{roleKey}},
+	"PUBLISH":      {fixed: []argRole{roleKey, roleValue}},
```
giortzisg, line 120: these should be under roleField right?

rueidis/sentryrueidis.go
```diff
@@ -0,0 +1,205 @@
...
+		cmds := cmd.Commands()
+		childSpan := redis.StartChildSpan(parentSpan, h.typ, redis.DBSystemRedis, cmds, cmd.IsReadOnly(), addr)
+		err := results[i].Error()
+		isNil := rueidis.IsRedisNil(err)
+		if err != nil && !isNil {
+			hasError = true
+		}
+		redis.FinishSpan(childSpan, h.typ, cmd.IsReadOnly(), err, isNil, respSize(results[i]))
+		redis.FinishIfNotNil(childSpan)
+	}
```
giortzisg, line 100: Currently in this part of the code, child spans would start after the redis client already finished, so timing would be incorrect here. Also as mentioned in https://github.com/getsentry/sentry-go/pull/1282#discussion_r3273327210, I think it's better to have a single multi span, since it better illustrates what actually happened in the code. I need to get back to you for what span data we should set, because it might be nice to have something like `cache.commands: ["SET", "GET", "DEL"]`.

internal/redis/scrub.go
```diff
@@ -0,0 +1,259 @@
...
+	b.WriteString(cmds[0])
+
+	args := cmds[1:]
+	fixedLen := len(pattern.fixed)
+
+	for i, arg := range args {
+		b.WriteByte(' ')
+
+		role := roleForPosition(i, fixedLen, pattern)
+		if role == roleKey || (sendDefaultPII && role == roleField) {
```
giortzisg, line 174: This misses scrubbing keys like `session:<sess-id>`, where the key also holds PII.

internal/redis/scrub.go
```diff
@@ -0,0 +1,259 @@
...
+
+// commandPatterns maps uppercase command names to their argument structure.
+// Unknown commands fall back to scrubbing all arguments.
+var commandPatterns = map[string]commandPattern{
+	// String commands
+	"GET":         {fixed: []argRole{roleKey}},
+	"SET":         {fixed: []argRole{roleKey, roleValue}, tailRole: roleField},
+	"SETNX":       {fixed: []argRole{roleKey, roleValue}},
+	"SETEX":       {fixed: []argRole{roleKey, roleField, roleValue}},
+	"PSETEX":      {fixed: []argRole{roleKey, roleField, roleValue}},
```
giortzisg, line 29: I feel that this approach is not really extendable and quite error prone. Brainstorming a bit for a potential improvement: 

```go
var commandPatterns = map[string]commandSpec{
    "GET":    read(firstKey), 
    "MGET":   read(allKeys),
    "SET":    write(firstKey),
	"AUTH": noop(), // skip command
}

type pattern struct {
	op cacheOp
	// roleKeys would be the current command key pattern with firstKey, allKeys, noKeys, etc
	roleKeys func(args []string) []string
}

func read(keys func([]string) []string) commandSpec {
    return pattern{op: cacheGet, keys: keys}
}

 func write(keys func([]string) []string) commandSpec {
    return pattern{op: cachePut, keys: keys}
 }

// and then scrubbing becomes extendable
// scrub gets called with 
// scrub(cmds, commandPatterns[cmd[0]], PII)
func scrub(cmds []string, roles []argRole, sendPII bool) string {
  for each cmd {
     switch {
     case role == roleKey:
        b.WriteString(scrubKey(arg, sendPII))
     case role == roleField && sendPII:
        b.WriteString(arg)
     default:
        b.WriteByte('?')
     }
  }

  return b.String()
}
```

One more thing here, we would also need to implement a `ScrubKey` for scrubbing keys like `session:<sess-id>` to `session:?`. Mentioned https://github.com/getsentry/sentry-go/pull/1282#discussion_r3273571543

giortzisg (commented):
Left some more comments around scrubbing

PR #1334, opened 2026-06-24
fix: preserve literal '%' in non-format log messages

sueun-dev:
The structured logger formats every log body with `fmt.Sprintf(message, args...)` in `(*sentryLogger).log`, including the non-format methods. `Emit` builds its message with `fmt.Sprint(args...)` and calls `log` with no remaining args, and `Write` goes through `Emit`, so a literal `%` in those messages is treated as a format verb.

```go
l.Info().Emit("disk usage at 95% capacity")
// Body: "disk usage at 95%!c(MISSING)apacity"
```

This also affects messages routed through the logging integrations, which forward the user's message via `Emit(...)` with no args (e.g. `slog`, `logrus`, `zap`).

Only run `fmt.Sprintf` when there are args; otherwise use the message as-is. The same resolved body is used for the debug logger output.

Added `Test_sentryLogger_EmitPreservesLiteralPercent` covering `Emit`, `Write`, and a message with format-like verbs. It fails before the change and passes after.

Tested: `go test -race .`, `go vet .`, `gofmt -s`, and `golangci-lint run ./` all clean.

### DIFF
--- a/log_test.go
+++ b/log_test.go
@@ -432,6 +432,61 @@ func Test_sentryLogger_Write(t *testing.T) {
 	}
 }
 
+func Test_sentryLogger_EmitPreservesLiteralPercent(t *testing.T) {
+	// Emit and Write are the non-format logging methods, so a literal '%' in
+	// the message must be preserved verbatim instead of being interpreted as a
+	// fmt verb.
+	tests := []struct {
+		name     string
+		logFunc  func(ctx context.Context, l Logger)
+		wantBody string
+	}{
+		{
+			name: "Emit with single percent",
+			logFunc: func(ctx context.Context, l Logger) {
+				l.Info().WithCtx(ctx).Emit("disk usage at 95% capacity")
+			},
+			wantBody: "disk usage at 95% capacity",
+		},
+		{
+			name: "Emit with format-like verbs",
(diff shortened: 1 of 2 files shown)


log_test.go
```diff
@@ -432,6 +432,61 @@ func Test_sentryLogger_Write(t *testing.T) {
...
+			},
+			wantBody: "disk usage at 95% capacity",
+		},
+		{
+			name: "Emit with format-like verbs",
+			logFunc: func(ctx context.Context, l Logger) {
+				l.Warn().WithCtx(ctx).Emit("got %s and %d, 100% done")
+			},
+			wantBody: "got %s and %d, 100% done",
+		},
```
giortzisg, line 457: This makes more sense here, since users should escape `%` in Emitf cases. Also the first case already guarantees that `%` literals are not skipped.
```suggestion
		{
			name: "Emit with format-like verbs",
			logFunc: func(ctx context.Context, l Logger) {
				l.Warn().WithCtx(ctx).Emitf("got %s and %d, 100%% done", "string", 10)
			},
			wantBody: "got %s and %d, 100% done",
		},
```

giortzisg:
Closing in favor of #1358

PR #1335, opened 2026-06-24
feat: PushScope shorthand now returns the new scope reference

DoctorJohn:
### Description

The [documentation](https://pkg.go.dev/github.com/getsentry/sentry-go#PushScope) says `PushScope` is a shorthand for `CurrentHub().PushScope`. However, so far, `PushScope` returned nothing while `CurrentHub().PushScope` returned the pushed `Scope`.

This PR updates `sentry.PushScope` to forward the return value from `CurrentHub().PushScope`, so that the shorthand behaves the same as the written-out variant. This aligns with the implementation of other shorthands, such as `FlushWithContext` and `LastEventID`.

### DIFF
--- a/sentry.go
+++ b/sentry.go
@@ -101,9 +101,9 @@ func ConfigureScope(f func(scope *Scope)) {
 }
 
 // PushScope is a shorthand for CurrentHub().PushScope.
-func PushScope() {
+func PushScope() *Scope {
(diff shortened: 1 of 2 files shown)


giortzisg (approved):

giortzisg:
Thanks for the contribution !

PR #1336, opened 2026-06-25
fix(scope): deep-copy event processors when cloning a scope

sueun-dev:
### Description

`Scope.Clone()` deep-copies every slice field (breadcrumbs, attachments, fingerprint) except `eventProcessors`, which shares the original's backing array:

```go
clone.eventProcessors = scope.eventProcessors
```

This was changed to a shared slice in #349, on the reasoning that "`Scope.AddEventProcessor` always creates a new slice, so we can share the backing array." That only holds when the slice has no spare capacity. `AddEventProcessor` appends, and `append` reuses the backing array while `len < cap`, so two clones of a scope that already has spare capacity write the same slot:

```
parent: 3 processors -> len 3, cap 4 (spare)
clone1 := parent.Clone(); clone2 := parent.Clone()  // both share the backing array
clone1.AddEventProcessor(x)  // writes backing[3]
clone2.AddEventProcessor(y)  // overwrites backing[3]; clone1 now runs y
```

Hubs/Scopes are cloned per request, so two goroutines cloning the same scope and adding a processor also race on that slot. `go test -race` on `master`:

```
WARNING: DATA RACE
Write at 0x... by goroutine 9:   scope.go:365   (append in AddEventProcessor)
Previous write at 0x... by goroutine 10: scope.go:365
```

Fix: copy the slice like the other fields (the `nil` guard keeps an unset processor list `nil` rather than an empty slice). `TestScopeCloneEventProcessorsIsolation` fails before this change and passes after. One assertion in `TestScopeBasicInheritance` compared the two processor slices with `assertEqual`; that only passed because `reflect.DeepEqual` short-circuits when two slices share a backing pointer (function values are otherwise never equal), so it now compares the count.

Tested with `go test -race ./` and `golangci-lint run ./`.

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -346,7 +346,10 @@ func (scope *Scope) Clone() *Scope {
 	clone.level = scope.level
 	clone.request = scope.request
 	clone.requestBody = scope.requestBody
-	clone.eventProcessors = scope.eventProcessors
+	if scope.eventProcessors != nil {
+		clone.eventProcessors = make([]EventProcessor, len(scope.eventProcessors))
+		copy(clone.eventProcessors, scope.eventProcessors)
+	}
 	clone.propagationContext = scope.propagationContext
 	clone.span = scope.span
 	return clone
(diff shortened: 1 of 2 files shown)


giortzisg:
Closing this in favor of https://github.com/getsentry/sentry-go/pull/1337.

PR #1373, opened 2026-07-23
feat!: add scope merging mechanism

### LINKED ISSUE #1369: Add scope merging for global and isolation scopes
Create the scope merging mechanism for global and isolation scopes.

giortzisg:
### Description
This adds the scope merging mechanism by allocating a `captureState` object and copying all needed event attributes in place. This removes taking multiple scope clones on every capture.

The PR also includes breaking changes around `EventModifier`. The `EventModifier` was an agnostic modifier that was passed on `ApplyToEvent`. This duplicated the EventProcessor behavior and also complicated the scope merging mechanism, since we need to have a concrete `*Scope` type to merge the scopes. Furthermore, there is no documented behavior or signature for any `EventModifier` and this seems to be just the first archaic way to modify events before `EventProcessor` and `BeforeSend` hooks where a thing. Thus decided to remove this and have a concrete `*Scope` passed on `ApplyToEvent`. This behavior aligns with other sdks that pass a concrete scope. 

This affects the public API with all the `CaptureX` methods now requiring a concrete scope to be passed, but further scope changes in the stack also change the `CaptureX` APIs to work with `context.Context`, so this change seems acceptable. Functionality isn't really removed but just shifted to `BeforeSend` hooks as intended. 

### Issues
* resolves: #1369
* resolves: GO-155

<details>

</details>

<details>
<summary>Reminders</summary>

</details>


internal/sentrytest/fixture.go
szokeasaurusrex, line 1: [question] I am a bit confused by these changes, could you explain why these are needed and how they relate to the rest of the PR?

client.go
```diff
@@ -661,18 +666,16 @@ func (client *Client) CaptureCheckIn(checkIn *CheckIn, monitorConfig *MonitorCon
 // The event must already be assembled. Typically, code would instead use
 // the utility methods like CaptureException. The return value is the
 // event ID. In case Sentry is disabled or event was dropped, the return value will be nil.
-func (client *Client) CaptureEvent(event *Event, hint *EventHint, scope EventModifier) *EventID {
-	if !client.IsEnabled() {
-		return nil
-	}
-	return client.processEvent(event, hint, scope)
+func (client *Client) CaptureEvent(event *Event, hint *EventHint, scope *Scope) *EventID {
+	return client.processEvent(event, scope, captureOptions{hint: hint})
```
szokeasaurusrex: Does the comment from above still hold?

> In case Sentry is disabled or event was dropped, the return value will be nil.

client.go
szokeasaurusrex, line 1: [question] I am a bit confused by all the client enabled checks being removed. Why are these no longer needed? Do we check the client is enabled elsewhere?

szokeasaurusrex (commented):
Admittedly only skimmed this, as the PR is quite huge, but I left a few questions

client.go
giortzisg: This is probably me missing some merge conflicts, a client.IsEnabled() on captureEvent should suffice for all other `CaptureX` functions. Let me rebase this properly.

internal/util/map.go
```diff
@@ -2,6 +2,23 @@ package util
 
 import "sync"
 
+// FillMap copies keys from src that are absent in dst. It allocates dst when
+// needed and never aliases src.
```
szokeasaurusrex: m: Please document the precedence between `dst` and `src` if both contain the same key with different corresponding values.

scope.go
```diff
@@ -536,30 +615,11 @@ func cloneContext(c Context) Context {
...
-	}
-
-	for k, v := range scope.attributes {
-		attrs[k] = v
-	}
+type signalCaptureContext struct {
+	scope             *Scope
+	ctx               context.Context
+	fallback          context.Context
+	defaultAttributes map[string]attribute.Value
```
szokeasaurusrex: l: Please document what this is, including documentation for the individual fields, because it is a bit unclear why we need this.

szokeasaurusrex (approved):

PR #1377, opened 2026-07-23
feat(otel/otlp): add WithProxy option to span exporter

### LINKED ISSUE #1376: otel/otlp: Add WithOTLPOptions escape hatch to pass arbitrary otlptracehttp.Option values
### Problem Statement

**Problem**

The `github.com/getsentry/sentry-go/otel/otlp` package only exposes `WithCompression`, `WithTLSClientConfig`, `WithTimeout`, and `WithRetry`. There's no way to configure other OTLP client behavior, like routing through an HTTP proxy via a custom `http.Client`.

### Solution Brainstorm

**Proposal**

Add a generic option to forward arbitrary `otlptracehttp.Option` values:

```go
// WithOTLPOptions appends additional otlptracehttp.Option values to the
// underlying OTLP HTTP client, for options not directly exposed by this
// package (e.g. custom http.Client / proxy). Applied after the DSN-derived
// options, so they take precedence over the defaults.
func WithOTLPOptions(opts ...otlptracehttp.Option) Option {
    return func(c *config) {
        c.otlpOptions = append(c.otlpOptions, opts...)
    }
}
```

The plumbing already exists: `config.otlpOptions` is a `[]otlptracehttp.Option` and `buildOTLPOptions` already appends them after the DSN options (`span_exporter.go:100`).

**Use case**

Routing Sentry OTLP traffic through an HTTP proxy — typically done by passing a custom `http.Client` with a `http.ProxyURL(...)` transport.

pierrre:
Closes https://github.com/getsentry/sentry-go/issues/1376

### Changelog Entry
- add `WithProxy`option for OTLP. This allows setting an `otlptracehttp.HTTPTransportProxyFunc` for the span exporter


giortzisg (approved):
Thanks for the contribution @pierrre

PR #1388, opened 2026-08-11
fix: guard the buffered request body against concurrent access

knQzx:
`Scope.SetRequest` replaces the request body with a `TeeReader` that fills a `LimitedBuffer`, so the buffer is written from whichever goroutine reads the body. the same buffer is read in `Scope.ApplyToEvent` when an event is captured

nothing synchronizes the two sides: the reader holds the scope lock, the `TeeReader` takes none. `Scope.Clone` also copies the pointer to the buffer, so cloning the hub for a goroutine, which is the documented way to use the SDK concurrently, shares the buffer instead of isolating it

it is reachable without opting into body collection: `SetRequest` installs the tee unconditionally, and `ApplyToEvent` evaluates `requestBody.Overflow()` before it checks `CollectHTTPBody`, so the flag is read while the body is still streaming. `go test -race` reports the write in `LimitedBuffer.Write` against the read in `Overflow`

this serializes access inside `LimitedBuffer`. the embedded `bytes.Buffer` is unexported so the promoted methods cannot bypass the lock, and `Bytes` returns a copy so the result stays stable while the body keeps growing. the buffer is capped at 10 KiB, so the copy is bounded

### DIFF
--- a/internal/httputils/body.go
+++ b/internal/httputils/body.go
@@ -3,6 +3,7 @@ package httputils
 import (
 	"bytes"
 	"io"
+	"sync"
 )
 
 // MaxBodyBytes is the maximum number of bytes the SDK captures from an HTTP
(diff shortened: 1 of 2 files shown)


giortzisg:
I would consider the racing window here quite unlikely and the tests don't really demonstrate a realistic scenario here. The limited buffer was introduced to eliminate wasteful copies of request bodies and the PR here introduces a copy, where at this stage we might as well just copy the request body eagerly on the integrations themselves. 

Closing this as it seems just hypothetical.

PR #1400, opened 2026-08-29
feat: Runtime metrics integration

aldy505:
### Description

This is a proof of concept that I talked really briefly on Slack. Might or might not be merged. NodeJS has [this integration](https://docs.sentry.io/platforms/javascript/guides/node/configuration/integrations/noderuntimemetrics/), and I thought, this is easy and really possible in Go. Let's just have it!

To use or test this feature/integration, just copy and paste the code into your codebase, and fix all the import bugs (stuff like `SetHubOnContext` should be `sentry.SetHubOnContext`, etc). Then use it, like so:

```go
// optional, but recommended. otherwise how would you handle graceful shutdown?
ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
defer cancel()

if err := sentry.Init(sentry.ClientOptions{}); err != nil {
  panic("yada yada yada")
}
defer sentry.FlushWithContext(ctx)
go sentry.StartRuntimeMetrics(sentry.RuntimeMetricsConfig{Context: ctx})

// the rest of your program
```

### Issues

<details>

</details>

<details>
<summary>Reminders</summary>

</details>

### DIFF
--- a/runtime_metrics.go
+++ b/runtime_metrics.go
@@ -78,7 +78,7 @@ func StartRuntimeMetrics(config RuntimeMetricsConfig) {
 
 	// XXX(aldy505): Do we guard the `interval` when it's below or over a certain threshold?
 	// Javascript SDK defaults to 30 seconds.
-	if config.Interval == 0 {
+	if config.Interval <= 0 {
 		config.Interval = time.Second * 30
 	}
 
--- a/runtime_metrics.go
+++ b/runtime_metrics.go
@@ -2,6 +2,7 @@ package sentry
 
 import (
 	"context"
+	"runtime"
 	runtime_metrics "runtime/metrics"
 	"sync"
 	"time"
@@ -61,6 +62,52 @@ var runtimeMetricsSamples = []runtime_metrics.Sample{
 // > scratch, which is always safe, though may be inefficient.
 var onceRuntimeMetrics = sync.Once{}
 
+// A simple marker to guarantee that the runtime metrics integration is only
+// started once.
+var runtimeMetricsRunning = false
+
+type cpuUtilizationTracker struct {
+	lastCPUSeconds float64
+	lastSampleTime time.Time
+}
+
+var maxProcs = float64(runtime.GOMAXPROCS(-1))
+
+func (t *cpuUtilizationTracker) GetCPUUtilization(currentCPUSeconds float64) float64 {
+	now := time.Now()
+
+	// First sample — can't calculate yet
+	if t.lastSampleTime.IsZero() {
+		t.lastCPUSeconds = currentCPUSeconds
+		t.lastSampleTime = now
+		return 0.0
+	}
+
+	// Calculate deltas
+	cpuDelta := currentCPUSeconds - t.lastCPUSeconds
+	wallClockDelta := now.Sub(t.lastSampleTime).Seconds()
+
+	// Normalize by GOMAXPROCS to get utilization percentage
+	// cpuDelta represents CPU-seconds consumed across all GOMAXPROCS goroutines
+	// wallClockDelta is real time that passed
+	// Divide by GOMAXPROCS to account for parallel CPUs
+	utilization := cpuDelta / (wallClockDelta * maxProcs)
+
+	// Clamp to [0.0, 1.0] (shouldn't exceed unless there's jitter)
+	if utilization > 1.0 {
+		utilization = 1.0
+	}
+	if utilization < 0.0 {
+		utilization = 0.0
+	}
+
+	// Update state for next sample
+	t.lastCPUSeconds = currentCPUSeconds
+	t.lastSampleTime = now
+
+	return utilization
+}
+
 // StartRuntimeMetrics starts the runtime metrics integration.
 // This should be called using `go sentry.StartRuntimeMetrics(config)`.
 // When invoked multiple times, the integration is only started once.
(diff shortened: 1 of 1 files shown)


runtime_metrics.go
```diff
@@ -0,0 +1,255 @@
...
+// You may leave this empty to use the default values. For proper graceful
+// shutdown, you should provide a context that is canceled when the program
+// exits.
+type RuntimeMetricsConfig struct {
+	// Disabled disables the runtime metrics integration.
+	Disabled bool
+	// Interval is the interval at which the runtime metrics are collected.
+	// Default is 30 seconds. You don't want to set this too low, as it will
+	// trigger a lot of "stop-the-world" activity.
+	Interval time.Duration
```
giortzisg: I don't think adding Interval makes sense for now.

giortzisg, line 24: This is misleading since the `runtime/metrics` library doesn't actually stop the world. Also the only thing this changes is collection frequency, and I don't think the the overhead is that high in general. Let's still keep the default at 30s, but I don't think that setting it to even 5s would be a problem.

runtime_metrics.go
```diff
@@ -0,0 +1,255 @@
...
+// From the Go docs:
+//
+// > It is safe to execute multiple Read calls concurrently, but their arguments
+// > must share no underlying memory. When in doubt, create a new []Sample from
+// > scratch, which is always safe, though may be inefficient.
+var onceRuntimeMetrics = sync.Once{}
+
+// A simple marker to guarantee that the runtime metrics integration is only
+// started once.
+var runtimeMetricsRunning = false
```
giortzisg, line 81: I agree that we should have only a single collector running at once, but an `atomic.Bool` is better here to allow restarts.

runtime_metrics.go
```diff
@@ -0,0 +1,255 @@
...
+	// Default is 30 seconds. You don't want to set this too low, as it will
+	// trigger a lot of "stop-the-world" activity.
+	Interval time.Duration
+	// Context is the context that is used to stop the runtime metrics
+	// integration. If you don't provide a context, the SDK will create
+	// an empty background context.
+	Context context.Context
+	// CollectGCMetrics enables the collection of GC metrics.
+	// Default is false (disabled).
+	CollectGCMetrics bool
```
giortzisg, line 31: I would name this something along the lines of `CollectOptionalMetrics`. We might add more than gc metrics in the future.

runtime_metrics.go
```diff
@@ -0,0 +1,255 @@
...
+			runtimeMetricKeyMap{"go.memory.gc.cycles", "cycles", MetricTypeCounter},
+			runtimeMetricKeyMap{"go.memory.gc.pause", UnitSecond, MetricTypeDistribution},
+		)
+
+		runtimeMetricsSamples = append(
+			runtimeMetricsSamples,
+			runtime_metrics.Sample{Name: "/gc/cycles/total:gc-cycles"},
+			runtime_metrics.Sample{Name: "/sched/pauses/total/gc:seconds"},
+		)
+	}
```
giortzisg, line 161: If we allow collector restarts that I recommend above, this becomes harder to manage and seem kinda hacky. I'd prefer we explicitly handle both, or at least clear the memory first rather than append.

runtime_metrics.go
```diff
@@ -0,0 +1,255 @@
...
+	// Javascript SDK defaults to 30 seconds.
+	if config.Interval <= 0 {
+		config.Interval = time.Second * 30
+	}
+
+	hub := CurrentHub()
+
+	if config.Context == nil {
+		config.Context = SetHubOnContext(context.Background(), hub)
+	}
```
giortzisg, line 173: We need to set the hub for both contexts

runtime_metrics.go
```diff
@@ -0,0 +1,275 @@
...
+// The following metrics are collected by the runtime metrics integration.
+// This should be synced with the `runtimeMetricsSamples` variable.
+var runtimeMetricsKeys = []runtimeMetricKeyMap{
+	{"go.cpu.utilization", UnitRatio, MetricTypeGauge},
+	{"go.memory.used.total", UnitByte, MetricTypeGauge},
+	{"go.memory.used.heap.objects", UnitByte, MetricTypeGauge},
+	{"go.memory.used.heap.free", UnitByte, MetricTypeGauge},
+	{"go.memory.used.heap.unused", UnitByte, MetricTypeGauge},
+	{"go.memory.used.heap.stacks", UnitByte, MetricTypeGauge},
+	{"go.memory.used.other", UnitByte, MetricTypeGauge},
```
giortzisg, line 56: These should be `go.memory.used` and have attributes that define the `detailed_type`.

runtime_metrics.go
```diff
@@ -0,0 +1,275 @@
...
+	if utilization < 0.0 {
+		utilization = 0.0
+	}
+
+	// Update state for next sample
+	t.lastCPUSeconds = currentCPUSeconds
+	t.lastSampleTime = nowTime
+
+	return utilization
+}
```
giortzisg, line 137: We can just get this from `/cpu/classes/total:cpu-seconds` and `/cpu/classes/idle:cpu-seconds`

runtime_metrics.go
```diff
@@ -0,0 +1,275 @@
...
+					}
+				case runtime_metrics.KindFloat64Histogram:
+					hist := sample.Value.Float64Histogram()
+					if hist == nil || len(hist.Buckets) < 2 || len(hist.Counts) != len(hist.Buckets)-1 {
+						// Invalid histogram
+						continue
+					}
+
+					for j := 0; j < len(hist.Counts); j++ {
+						count := hist.Counts[j]
```
giortzisg, line 243: Both bugbot comments are correct here, and I don't think it's that trivial to properly support the histograms here.

runtime_metrics.go
```diff
@@ -0,0 +1,275 @@
...
+				case runtime_metrics.KindFloat64:
+					value := sample.Value.Float64()
+					if sample.Name == "/cpu/classes/total:cpu-seconds" {
+						value = cpuTracker.GetCPUUtilization(value)
+					}
+					switch runtimeMetricsKeys[i].Type {
+					case MetricTypeGauge:
+						meter.Gauge(runtimeMetricsKeys[i].Key, value, WithUnit(runtimeMetricsKeys[i].Unit))
+					case MetricTypeCounter:
+						meter.Count(runtimeMetricsKeys[i].Key, int64(value), WithUnit(runtimeMetricsKeys[i].Unit))
```
giortzisg, line 218: Counters need deltas.

PR #1417, opened 2026-09-04
feat: add context providers to logging integrations

giortzisg:
### Issues

<details>

</details>

<details>
<summary>Reminders</summary>

</details>

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


logrus/logrusentry.go
```diff
@@ -56,30 +58,39 @@ type Hook interface {
...
+		return
+	}
+	h.SetContextProvider(func() context.Context {
+		hub := provider()
+		if hub == nil {
+			return nil
+		}
+		return sentry.SetHubOnContext(context.Background(), hub)
+	})
 }
```
szokeasaurusrex, line 88: l: Why is this still needed if we are getting rid of hubs? 

Is it just to satisfy the `Hook` interface, and if yes, why not just break that interface in this PR?

logrus/logrusentry.go
```diff
@@ -218,11 +218,20 @@ func (h *logHook) Levels() []logrus.Level {
...
 
 func (h *logHook) Flush(timeout time.Duration) bool {
-	return h.hubProvider().Client().Flush(timeout)
+	return sentry.ClientFromContext(h.providerContext()).Flush(timeout)
 }
 
 func (h *logHook) FlushWithContext(ctx context.Context) bool {
-	return h.hubProvider().Client().FlushWithContext(ctx)
+	return sentry.ClientFromContext(h.providerContext()).FlushWithContext(ctx)
+}
```
szokeasaurusrex, line 226: [question] in what case would someone want to use this API?

It seems confusing to have two different contexts. What is the difference between the two contexts and how they are used?

logrus/logrusentry_test.go
```diff
@@ -109,7 +110,46 @@ func TestLogHookSetHubProviderUsesProviderForLogs(t *testing.T) {
...
+	ctx = sentry.ContextWithClient(ctx, sentry.NewNoopClient())
+
+	hook := NewLogHookFromClient([]logrus.Level{logrus.InfoLevel}, defaultClient)
+	hook.SetContextProvider(func() context.Context { return ctx })
+
+	err := hook.Fire(&logrus.Entry{Level: logrus.InfoLevel, Message: "suppressed"})
+	assert.NoError(t, err)
+	assert.True(t, defaultClient.Flush(testutils.FlushTimeout()))
+	assert.Empty(t, defaultTransport.Events())
+}
```
szokeasaurusrex, line 150: m: I am skeptical we need this test. 

At least as far as I can tell, there is no notion of a "global default" Sentry client in Go, at least not after these changes.

So, it is therefore a bit strange to test that a Logrus hook set up with a no-op client would then record events on a totally unrelated test client; even if this client is the default client for tests, it is not a true global default.

It is possible I am missing something, and in that case, I would appreciate if you can let me know where my reasoning is wrong. If I am right, my recommendation would be either to completely remove this test, or to instead narrow it a bit just to check that the logrus hook can be initialized with a no-op client successfully, without then asserting another unrelated client's behavior.

zerolog/sentryzerolog.go
```diff
@@ -243,12 +251,17 @@ func (w *Writer) WriteLevel(level zerolog.Level, p []byte) (int, error) {
...
+	if ok := sentry.ClientFromContext(w.ctx).Flush(w.flushTimeout); !ok {
 		return ErrFlushTimeout
 	}
 	return nil
 }
 
+func contextWithClient(client *sentry.Client) context.Context {
+	ctx := sentry.ContextWithScope(context.Background(), sentry.NewScope())
+	return sentry.ContextWithClient(ctx, client)
+}
```
szokeasaurusrex, line 263: l: I would rename this function to indicate more clearly the difference versus `sentry.ContextWithClient` (i.e. that a new scope is also created).

```suggestion
func contextWithClientAndNewScope(client *sentry.Client) context.Context {
	ctx := sentry.ContextWithScope(context.Background(), sentry.NewScope())
	return sentry.ContextWithClient(ctx, client)
}
```

Although as far as I can tell, the function is only used in one place, so you could also consider removing it.

zerolog/sentryzerolog.go
```diff
@@ -135,16 +135,24 @@ func NewWithHub(hub *sentry.Hub, opts Options) (*Writer, error) {
...
 }
 
+// NewWithHub creates a writer using an existing sentry Hub and options.
+func NewWithHub(hub *sentry.Hub, opts Options) (*Writer, error) {
+	if hub == nil {
+		return nil, errors.New("hub cannot be nil")
+	}
+	return NewWithContext(sentry.SetHubOnContext(context.Background(), hub), opts)
+}
+
```
szokeasaurusrex, line 152: m: This appears to be unused; I am somewhat surprised that there is not a linter to catch this (or perhaps I am mistaken)

```suggestion
```

zerolog/sentryzerolog_test.go
```diff
@@ -47,16 +48,15 @@ func TestFailedClientCreation(t *testing.T) {
...
 
-	_, err := NewWithHub(hub, Options{
+	_, err := NewWithContext(ctx, Options{
 		Levels: []zerolog.Level{zerolog.ErrorLevel},
 	})
 	require.Nil(t, err)
 
-	_, err = NewWithHub(nil, Options{})
+	_, err = NewWithContext(nil, Options{}) // nolint: staticcheck // nil validation
 	require.NotNil(t, err)
```
szokeasaurusrex, line 60: l: I am guessing the linter would catch such a case (that's why we suppress it), and if so, why do we need to test it here?

szokeasaurusrex (commented):
Mostly lgtm, but I noticed some small things likely worth addressing

PR #1419, opened 2026-09-04
feat: add capture feedback API

### LINKED ISSUE #842: Add function to capture user feedback
Other Sentry SDKs have a function called `captureUserFeedback` that takes an `event_id`, and user input such as name, email and comments. develop docs on it are here: https://develop.sentry.dev/sdk/features/#user-feedback

A new feedback format was added recently to power the Web Widget (see: https://docs.sentry.io/product/user-feedback/).
In JS the function name became `captureFeedback`. Final docs are not yet on _develop_ but tracked here:

* https://github.com/getsentry/develop/issues/1206

al1re3a:
### Description

Adds first-class support for the new Sentry User Feedback format.

- exposes `CaptureFeedback` through the package, Hub, and Client APIs
- adds a `Feedback` model with message, contact, URL, source, and associated-event fields
- sends feedback using the dedicated envelope item and rate-limit category
- preserves scope tags, contexts, attachments, and event processors
- covers both legacy and telemetry transport paths

### Issues

- resolves: #842

### Testing

- `go test -count=1 -timeout 300s -race work`
- `go vet work`
- `golangci-lint v2.11.4 run ./...` (0 issues)

### DIFF
--- a/client.go
+++ b/client.go
@@ -614,6 +614,15 @@ func (client *Client) CaptureException(exception error, hint *EventHint, scope E
 	return client.CaptureEvent(event, hint, scope)
 }
 
+// CaptureFeedback captures user-provided feedback.
+func (client *Client) CaptureFeedback(feedback *Feedback, hint *EventHint, scope EventModifier) *EventID {
+	event := client.EventFromFeedback(feedback)
+	if event == nil {
+		return nil
+	}
+	return client.CaptureEvent(event, hint, scope)
+}
+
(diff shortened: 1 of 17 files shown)


giortzisg:
The implementation is missing registering a telemetry processor buffer with a `ratelimit.CategoryFeedback`. We should also add an integration test that uses the default processor similar to the ones already existing under `client_test.go`

PR #1425, opened 2026-09-10
ref(transport): move HTTP delivery into the root package

giortzisg:
### Description
This moves the `internal/http/transport` to the root package.

### Issues

<details>

</details>

<details>
<summary>Reminders</summary>

</details>

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -1,4 +1,4 @@
-package http
+package sentry
 
 import (
 	"bytes"
@@ -21,11 +21,11 @@ import (
 )
 
 const (
-	apiVersion = 7
+	httpAPIVersion = 7
 
-	defaultTimeout           = time.Second * 30
-	defaultQueueSize         = 1000
-	defaultClientReportsTick = time.Second * 30
+	defaultHTTPTimeout           = time.Second * 30
+	defaultHTTPQueueSize         = 1000
+	defaultHTTPClientReportsTick = time.Second * 30
 )
 
 var (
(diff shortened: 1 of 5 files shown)


sentry.go
```diff
@@ -10,9 +10,9 @@ import (
 // The version of the SDK.
 const SDKVersion = "0.49.0"
 
-// apiVersion is the minimum version of the Sentry API compatible with the
+// legacyAPIVersion is the minimum version of the Sentry API compatible with the
 // sentry-go SDK.
-const apiVersion = "7"
+const legacyAPIVersion = "7"
```
szokeasaurusrex: l: Why is this change being made here? Seems possibly out of scope

transport_legacy.go
szokeasaurusrex, line 1: l: Also wondering why the renames here are needed. It looks like the items renamed here are private items and given the filename already includes "legacy," it seems a bit redundant to also put "legacy" in the variable names.

transport.go
szokeasaurusrex, line 167: m: As I understand it, `telemetry.Transport` is a private interface. Returning such an interface from this public function may be awkward.

Is this intentional?

szokeasaurusrex, line 319: m: Same question asked of `NewSyncTransport` also applies here

transport_legacy.go
giortzisg: Moving the internal transport to root actually produces naming conflicts with things that we had to duplicate for internal use. The `legacy_transport` file would be deleted in the last step so currently just renaming the conflicting variables with a legacy prefix.

sentry.go
giortzisg: Yeah good point for this one, since it's under `sentry.go`

transport.go
giortzisg: This is currently done for the sake of not exposing way too many changes on this PR. We have a conflict on the two Transport interfaces currently and renaming the legacy one would mean needing to change all places where it's accessed. But good point, the intention here is for this to become public on the next step.

